Issue #61, opened 2023-06-16
Receiving server is rejecting at MAIL FROM and complaining about lack of EHLO when sending over TLS


tommairs:
### What Operating System are you seeing this problem on?

NAME="Ubuntu"
VERSION="20.04.6 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.6 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal

### What Hardware is this system running?

Standard D8lds v5 (8 vcpus, 16 GiB memory)

### KumoMTA version

kumod 2023.06.15-4cf2e65c

### Did you try the latest release to see if the issue is better (or worse!) than your current version?

Yes, and I updated the version box above to show the version that I tried

### Describe the bug

When using TLS, the receiving server *may* reject the message complaining about a bad MAIL FROM.
When you turn odd TLS, the message flows properly.
It is possible that the EHLO is not being issued again after the STARTTLS.
 

### To Reproduce

See above ^^

### Configuration

kumo.on('get_egress_path_config', function(domain, source_name, site_name)
  return kumo.make_egress_path {
    enable_tls = 'Opportunistic',
  }
end)

PR #62, opened 2023-06-16
Resend EHLO after STARTTLS

bendyer:
Addresses https://github.com/KumoCorp/kumomta/issues/61

RFC 3207 says clients MUST forget the EHLO capabilities obtained prior to STARTTLS, and SHOULD re-send an EHLO after TLS negotiation completes (https://datatracker.ietf.org/doc/html/rfc3207#section-4.2). Some SMTP servers (incorrectly?) seem to _require_ an EHLO after STARTTLS so this ensures we send one.

Renamed `caps` to `pretls_caps` as a reminder to use the post-TLS caps returned at line 220 for anything else we might need later on.


wez:
Thanks! Yeah, this looks good!

PR #70, opened 2023-07-06
Add method remove_all_named_header in lua

postmastery:
To support removal of any header the remove_all_named_header is exported in Lua. This can be used to remove bogus headers added by internal applications, for example Return-Path.

### DIFF
--- a/crates/message/src/message.rs
+++ b/crates/message/src/message.rs
@@ -665,7 +665,7 @@ impl Message {
         })
     }
 
-    pub fn remove_all_named_header(&self, name: &str) -> anyhow::Result<()> {
+    pub fn remove_all_named_headers(&self, name: &str) -> anyhow::Result<()> {
         self.retain_headers(|hdr| !hdr.get_key_ref().eq_ignore_ascii_case(name))
     }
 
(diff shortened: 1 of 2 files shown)


crates/message/src/message.rs
```diff
@@ -827,6 +827,12 @@ impl UserData for Message {
                     .map_err(any_err)?)
             },
         );
+        methods.add_method(
+            "remove_all_named_header",
```
wez: Let's pluralize this; I can see that the underlying rust method isn't pluralized and that seems like a typo that I made.  If you feel like pluralizing the rust code while you're looking at this, that would be appreciated!

```suggestion
            "remove_all_named_headers",
```

wez (commented):
Thanks!

You can resolve the code formatting issue by running `make fmt`.  You may need to explicitly add the nightly rust toolchain via rustup for that to succeed; we use stable rust when compiling, but our formatting options currently require features from nightly rustfmt.

Please also add some documentation; you can copy https://github.com/KumoCorp/kumomta/blob/main/docs/reference/message/remove_x_headers.md to `remove_all_named_headers.md` in that same directory and use that as a starting point.

Please place `{{since('dev')}}`  on line 2 or 3 (under the initial header); that's a doc macro that will act as a placeholder until we tag the next release.

To build and test the docs locally:

```console
$ ./docs/build.sh serve
```

should install the required components for building the docs, and then run a webserver to host them; it will print out the URL you can use with your browser.

wez:
Thanks!

PR #80, opened 2023-08-29
add mx_list_ip_map to smtp protocole for tls connection

cai-n:
add a hashmap of ip, domains to have the canonical name for tls connection and avoid the error: rust invalid peer certificate: NotValidForName when enable_tls is not "Disabled" or "OpportunisticInsecure"

### DIFF
--- a/assets/policy-extras/rollup.lua
+++ b/assets/policy-extras/rollup.lua
@@ -33,6 +33,7 @@ local function compute_ip_rollup_mx_list(domain, routing_domain)
   end
 
   local addrs = {}
+  local addrs_map = {}
   -- The host names are pre-sorted by lookup_mx
   for _, host in ipairs(hosts) do
     local host_addrs = kumo.dns.lookup_addr(host)
@@ -41,19 +42,23 @@ local function compute_ip_rollup_mx_list(domain, routing_domain)
     -- and hash consistently with the resulting ready queue name
     table.sort(host_addrs)
     for _, a in ipairs(host_addrs) do
-      table.insert(addrs, string.format('[%s]', a))
+      local ip = string.format('[%s]', a)
+      table.insert(addrs, ip)
+      addrs_map[ip] = host
     end
   end
 
-  return addrs
+  return addrs, addrs_map
 end
 
 function mod.apply_ip_rollup_to_queue_config(domain, routing_domain, params)
-  local mx_list = compute_ip_rollup_mx_list(domain, routing_domain)
+  local mx_list, mx_list_ip_map =
+    compute_ip_rollup_mx_list(domain, routing_domain)
   if mx_list then
     params.protocol = {
       smtp = {
         mx_list = mx_list,
+        mx_list_ip_map = mx_list_ip_map,
       },
     }
   end
--- a/assets/policy-extras/rollup.lua
+++ b/assets/policy-extras/rollup.lua
@@ -41,7 +41,7 @@ local function compute_ip_rollup_mx_list(domain, routing_domain)
     -- and hash consistently with the resulting ready queue name
     table.sort(host_addrs)
     for _, a in ipairs(host_addrs) do
-      table.insert(addrs, { ['name'] = host, ['addr'] = a })
+      table.insert(addrs, { name = host, addr = a })
     end
   end
 
--- a/assets/policy-extras/rollup.lua
+++ b/assets/policy-extras/rollup.lua
@@ -33,7 +33,6 @@ local function compute_ip_rollup_mx_list(domain, routing_domain)
   end
 
   local addrs = {}
-  local addrs_map = {}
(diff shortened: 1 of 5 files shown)


crates/kumod/src/smtp_dispatcher.rs
```diff
@@ -55,11 +58,14 @@ impl SmtpDispatcher {
...
-                        .with_context(|| format!("resolving mx_list entry {a}"))?,
-                );
+                let addrs = &mut resolve_a_or_aaaa(a)
+                    .await
+                    .with_context(|| format!("resolving mx_list entry {a}"))?;
+                match proto_config.mx_list_ip_map.get(a) {
+                    Some(name) => addrs[0].name = name.to_string(),
+                    None => (),
+                }
+                addresses.append(addrs);
```
wez: Thanks for this!

How about this as an alternative implementation approach:

Add a new type:

```rust
#[derive(Deserialize, Serialize, Debug, Clone, Default)]
#[serde(untagged)]
pub enum MxListEntry {
   // A name that needs to be resolved to its A or AAAA record in DNS,
   // or an IP domain literal enclosed in square brackets like `[10.0.0.1]`
   Name(String),
   // A pre-resolved name and IP address
   Resolved(ResolvedAddress)
}
```

The `untagged` attribute will allow serde to pattern match either a string or a lua table to this type,
provided the lua table matches the shape of the `ResolvedAddress` type defined in `kumo-log-types`.

Change the `mx_list` field in `SmtpProtocol` to use this type instead of just `String`:

```rust
    #[serde(default)]
    pub mx_list: Vec<MxListEntry>,   
```

Change this section of code to resolve if needed and append to the list:

```suggestion
                match a {
                    MxListEntry::Name(a) => {
                            addresses.append(&mut resolve_a_or_aaaa(a)
                               .await
                               .with_context(|| format!("resolving mx_list entry {a}"))?));
                    }
                    MxListEntry::Resolved(addr) => {
                            addresses.append(addr);
                     }
                }
```

Now, `mx_list` can be specified like:

```lua
   mx_list = { 'smart.host.local', {name="mx.example.com", addr='10.0.0.1'} },
```

and we won't need to add the additional `mx_list_ip_map`.

cai-n: this is way better, I'll push a new commit with these changes

assets/policy-extras/rollup.lua
```diff
@@ -41,7 +41,7 @@ local function compute_ip_rollup_mx_list(domain, routing_domain)
     -- and hash consistently with the resulting ready queue name
     table.sort(host_addrs)
     for _, a in ipairs(host_addrs) do
-      table.insert(addrs, string.format('[%s]', a))
+      table.insert(addrs, { ['name'] = host, ['addr'] = a })
```
wez: I know these are equivalent, but I favor less punctuation as it is easier to read:

```suggestion
      table.insert(addrs, { name = host, addr = a })
```

crates/kumod/src/smtp_dispatcher.rs
```diff
@@ -17,7 +17,17 @@ use std::net::SocketAddr;
...
+    pub mx_list: Vec<MxListEntry>,
+}
+
+#[derive(Deserialize, Serialize, Debug, Clone)]
+#[serde(untagged)]
+pub enum MxListEntry {
+    // A name that needs to be resolved to its A or AAAA record in DNS,
+    // or an IP domain literal enclosed in square brackets like `[10.0.0.1]`
+    Name(String),
+    // A pre-resolved name and IP address
```
wez: Sorry, I should have used the `///` version of this when I proposed this initially!

```suggestion
    /// A name that needs to be resolved to its A or AAAA record in DNS,
    /// or an IP domain literal enclosed in square brackets like `[10.0.0.1]`
    Name(String),
    /// A pre-resolved name and IP address
```

wez (commented):
Looks good! I have a couple of very minor suggestions that I can apply and merge, or I can wait for you to apply and then merge; let me know which you prefer!

docs/userguide/policy/routing.md
```diff
@@ -54,7 +54,7 @@ several options.
             return kumo.make_queue_config {
                 protocol = {
                     smtp = {
-                        mx_list = { 'smart.host.local', '[10.0.0.1]' },
+                        mx_list = { 'smart.host.local', {name="mx.example.com", addr='10.0.0.1'} }
```
wez: Ah, the CI is unhappy about this because it hasn't been lua code formatted.  You can run `./docs/build.sh` to take care of that in the same way that it does, but really I suspect that it just wants you to use single quotes instead of double quotes around the strings

cai-n:
It should be alright now, hopefully

wez:
Thank you!

PR #91, opened 2023-11-02
add minijinja_contrib dependency

cai-n:
add minijinja_contrib dependency to support datetime filters.

### DIFF
--- a/crates/kumod/src/logging.rs
+++ b/crates/kumod/src/logging.rs
@@ -10,6 +10,7 @@ pub use kumo_log_types::*;
 use kumo_server_runtime::rt_spawn_non_blocking;
 use message::{EnvelopeAddress, Message};
 use minijinja::{Environment, Template};
+use minijinja_contrib::add_to_environment;
 use mlua::{Lua, Value as LuaValue};
 use once_cell::sync::{Lazy, OnceCell};
 use rfc5321::{EnhancedStatusCode, Response, TlsInformation};
@@ -82,6 +83,9 @@ pub struct LogRecordParams {
     /// minijinja template
     #[serde(default)]
     pub template: Option<String>,
+
+    #[serde(default)]
+    pub template_datetime_format: Option<String>,
 
     /// Written to the start of each newly created log file segment
     #[serde(default)]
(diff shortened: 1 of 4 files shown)


crates/kumod/src/logging.rs
```diff
@@ -197,6 +201,10 @@ impl Logger {
 
         for (kind, per_rec) in &params.per_record {
             if let Some(template_source) = &per_rec.template {
+                if let Some(datetime_format) = &per_rec.template_datetime_format {
+                    template_engine.add_global("DATETIME_FORMAT", datetime_format.clone());
+                }
```
wez: I'm not so sure about this part; here's what's on my mind:

* If we add this, it suggests that we also should add the other global variables mentioned in the docs; `DATE_FORMAT`, `TIME_FORMAT` and `TIMEZONE`
* Setting up those optional globals would need to happen in each place where we configure minijinja; there are two locations in this file, but we also use it for the HTTP injection API, and any other place where we add templating in the future
* It feels like a lot of repetition for an optional feature.

I wonder if we can simply skip this and leave it to jinja syntax.  My understanding is that you should be able to set global variables at the top of your template block, and then use the new filter like this:

```jinja
{% set DATETIME_FORMAT = "iso" %}
created={{ created | datetimeformat }}
when={{ timestamp | datetimeformat }}
```

alternatively, although slightly more verbose, this doesn't seem too bad:

```jinja
created={{ created | datetimeformat(format="iso") }}
when={{ timestamp | datetimeformat(format="iso") }}
```

what do you think about this? Can you share more information about what motivated you to do it this way?

wez:
Please also enable these new functions for the injection API over here: https://github.com/KumoCorp/kumomta/blob/ffc15f2f74e358c68701e7a947a8f95ca8033e70/crates/kumod/src/http_server/inject_v1.rs#L237
Thanks!

crates/kumod/src/logging.rs
cai-n: Ah yeah, haven't though about setting the global in jinja directly, that totally works.
`template = '{%- set DATETIME_FORMAT = "[year]-[month]-[day] [hour]:[minute]:[second][offset_hour sign:mandatory][offset_minute]" -%}{{ type[:1] | lower }},{{ timestamp | datetimeformat }}'`
This gives me exactly what I'll looking for

wez:
Thanks!

PR #94, opened 2023-11-13
only add meta to log if not null

cai-n:
To keep it the same as headers, only add meta to logs if they are not null.


wez:
Thanks!

PR #101, opened 2023-11-30
queue helper: move kumo.on get_queue_config to own function

cai-n:
Add setup_get_queue_config function to make use of the helper and still be able to set get_queue_config handlers in init.lua for rollup.

### DIFF
--- a/assets/policy-extras/queue.lua
+++ b/assets/policy-extras/queue.lua
@@ -57,6 +57,7 @@ retry_interval = '17 mins'
 ---
 local queue_module = require 'policy-extras.queue'
 local queue_helper = queue_module:setup({'/opt/kumomta/etc/queue.toml'})
+queue_helper:setup_get_queue_config()
 
 kumo.on('smtp_server_message_received', function(msg)
   queue_helper:apply(msg)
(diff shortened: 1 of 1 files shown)


cai-n:
Yeah, I use apply_ip_rollup_to_queue_config in the queue config event handler in init.lua that's why I wanted to setup a way to skip the event handling in the helper.
I'll change it to your alternate setup method.

assets/policy-extras/queue.lua
```diff
@@ -176,31 +176,44 @@ local function resolve_config(data, domain, tenant, campaign)
...
+    file_names = options.file_names,
   }
 
+  function helper:resolve_config(domain, tenant, campaign)
+    local data = cached_load_data(options.file_names)
+    local params = resolve_config(data, domain, tenant, campaign)
+    return params
+  end
+
+  if options.skip_queue_config_hook == false then
```
wez: nit: this "reads aloud" better:

```suggestion
  if not options.skip_queue_config_hook then
```

wez (commented):
OK, makes sense!
One small suggestion and I think that this can be merged!

PR #135, opened 2024-02-07
reuse connection for lua deliveries

cai-n:
Don't create clients on every webhook call and reuse the existing one.

### DIFF
--- a/crates/kumod/src/lua_deliver.rs
+++ b/crates/kumod/src/lua_deliver.rs
@@ -59,7 +59,6 @@ impl QueueDispatcher for LuaQueueDispatcher {
 
     async fn attempt_connection(&mut self, dispatcher: &mut Dispatcher) -> anyhow::Result<()> {
         if self.connection.is_some() {
-            dispatcher.delivered_this_connection = 0;
             return Ok(());
         }
         let connection_wrapper = dispatcher.metrics.wrap_connection(());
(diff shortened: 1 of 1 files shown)


crates/kumod/src/lua_deliver.rs
```diff
@@ -58,6 +58,10 @@ impl QueueDispatcher for LuaQueueDispatcher {
     }
 
     async fn attempt_connection(&mut self, dispatcher: &mut Dispatcher) -> anyhow::Result<()> {
+        if self.connection.is_some() {
```
wez, line 61: ah, good catch!

crates/kumod/src/lua_deliver.rs
```diff
@@ -58,6 +58,10 @@ impl QueueDispatcher for LuaQueueDispatcher {
...
+            dispatcher.delivered_this_connection = 0;
```
wez: This line would appear to disable max_deliveries_per_connection; is that intentional?

cai-n: yeah, I thought it would be better to just keep using the same connection and let the maintainer_task close it if there's no messages anymore.

wez: I'd like to keep some kind of limit here to defend against pathological situations if the system is ever under extreme stress.  You should be able to configure this differently via the egress path configuration if you'd like to use more than the default of 1024 per connection.
Can we remove this line from the PR?

cai-n: for sure, I'll push an update to remove the line

wez:
Thanks!

PR #157, opened 2024-03-21
add serializer for response content field

cai-n:
Added serde to change \r and \n to spaces in response content because If the content is multiline, the regex used for automation needs to use \\n to get the line break and dsn_rewrite only works on the first line.

### DIFF
--- a/crates/rfc5321/src/client_types.rs
+++ b/crates/rfc5321/src/client_types.rs
@@ -145,6 +145,7 @@ impl SmtpClientTimeouts {
 pub struct Response {
     pub code: u16,
     pub enhanced_code: Option<EnhancedStatusCode>,
+    #[serde(serialize_with = "as_single_line")]
     pub content: String,
     pub command: Option<String>,
 }
@@ -157,13 +158,7 @@ impl Response {
             line.push_str(&format!("{}.{}.{} ", enh.class, enh.subject, enh.detail));
         }
 
-        for c in self.content.chars() {
-            match c {
-                '\r' => line.push_str("\\r"),
-                '\n' => line.push_str("\\n"),
-                c => line.push(c),
-            }
-        }
+        line.push_str(&remove_line_break(&self.content));
 
         line
     }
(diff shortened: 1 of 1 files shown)


wez:
Hi! Can you share more information about how this affected you? The transformation that is performed in `main` (eg: prior to your PR) is to convert the CRLF into literal `\r\n` (eg: with the backslash and the letters, rather than the control characters). The docs at https://docs.kumomta.com/reference/events/smtp_client_rewrite_delivery_status/ also reflect this (although, perhaps not as clearly as they could!).

cai-n:
The log I was trying the automation rules on was: 
`550 4.7.28 Gmail has detected an unusual rate of unsolicited mail originating\nfrom your DKIM domain [fakemessage.com 36]. To protect our users from spam,\nmail sent from your domain has been temporarily rate limited. For more\ninformation, go to https://support.google.com/mail/?p=UnsolicitedRateLimitError\nto review our Bulk Email Senders Guidelines.\ne26-20020a5d595a000000b0033ce06c303esi4306826wri.186 - gsmtp`
I used this regex for tsa:
`Gmail has detected an unusual rate of unsolicited mail originating from your DKIM domain` and it only worked if I added the '\n' between 'originating' and 'from' which is expected. 
But when I tried it on dsn_rewrite, I couldn't get it to work, I've tried with '\n'  and '\\n' but maybe I'm not doing it right.  
I thought it would be easier if the line breaks became spaces in case the string used in the regex just happened to be split in 2 when it's received.

wez:
One reason I'm reticent to rewrite the line endings to spaces is that it is conceptually a lossy transformation, and I think it might also be confusing in some situations. For example, I recently saw someone share a log with this as the content:

```
452 4.2.2 The recipient's inbox is out of storage space. Please direct the\n4.2.2 recipient to\n4.2.2 https://support.google.com/mail/?p=OverQuotaTemp 
```

In the transformation to a single line, the SMTP code was removed from the secondary lines, but in this particular instance, those continued lines were prefixed with the enhanced status code (4.2.2), which, AFAICT, is not strictly conforming to the RFCs, but the language in them is vague on this point, and is not consistent with the log line that you shared above.

For this particular line, if we wrote the newline to space then it would look weird to have `direct the 4.2.2 recipient` in the payload.

Now, I suppose we could strip both `452 4.2.2 ` and `452 ` from the start of lines in this sort of situation to resolve that part of the ambiguity.

If we were to change this, we'd need to consider whether such a change might impact other already deployed configurations for other users before we ship it out, so that we can work to minimize any potential disruption to them.

Separately: it should be possible to use a regex like `\\r` to match this log line.
We're you using the example code from https://docs.kumomta.com/reference/events/smtp_client_rewrite_delivery_status/ to match and process this? I think we'd like to try to replicate what you were seeing to make sure that we've got something that works for this case independently from potentially modifying how we represent responses as a single line.

cai-n:
I did use the code from the example and managed to get dsn_rewrite working in the end but I had to use `\\\\n`, haven't figured out why yet, but for tsa, only `\\n` was needed, it might have something to do with the format, tsa was toml and dsn_rewrite json. 
If it's too bothersome, I'll rework the regexes to make it work for us and you can close this PR, it's alright. If not I'll rework the serialization to also strip the enhanced_code from the line to make the payload clearer

wez:
Ah, I see. Since the response replaces the LF character with the literal `\n` sequence, the regex must include a literal `\`, but to represent that correctly to the regex engine, it needs one `\\` to indicate a backslash in the JSON, but that itself needs to be quoted with a backslash so that the regex engine knows that it is a literal backslash rather than a backslash escape.

I've pushed f54922b9f0f210456834bbf49cba618ad918d3b6 to strip the enhanced status code that I mentioned previously.

I will think a bit about removing the newlines!

wez:
I've merged this, thank you!

PR #169, opened 2024-04-18
Add docs build on GH Actions

wez:


wez:
retired the idea of using arc runners for this

PR #173, opened 2024-04-20
ci: testing sccache

wez:


wez:
Vedict: even though we spend a total of 4 minutes downloading and uploading from minio using meltwater/drone-cache, it is still faster than using sccache.  The reason is that significant portions of our build are not cacheable by sccache.

PR #178, opened 2024-04-30
Remove ready_count metrics on site reap

cai-n:
Add READY_COUNT_GAUGE to the list of metrics to remove on site reap


wez:
Thanks!

PR #179, opened 2024-05-02
add prefix parameter for prometheus metrics

cai-n:
Add prefix parameter in prometheus metrics url

### DIFF
--- a/crates/kumo-server-common/src/http_server/mod.rs
+++ b/crates/kumo-server-common/src/http_server/mod.rs
@@ -9,7 +9,7 @@ use axum_server::tls_rustls::RustlsConfig;
 use cidr_map::{AnyIpCidr, CidrSet};
 use data_loader::KeySource;
 use kumo_server_runtime::spawn;
-use serde::{Deserialize};
+use serde::Deserialize;
 use std::net::{IpAddr, SocketAddr, TcpListener};
 use std::str::FromStr;
 use std::sync::Arc;
--- a/crates/kumo-server-common/src/http_server/mod.rs
+++ b/crates/kumo-server-common/src/http_server/mod.rs
@@ -1,6 +1,6 @@
 use crate::diagnostic_logging::set_diagnostic_log_filter;
 use anyhow::Context;
-use axum::extract::{DefaultBodyLimit, Json};
+use axum::extract::{DefaultBodyLimit, Json, Query};
 use axum::http::StatusCode;
 use axum::response::{IntoResponse, Response};
 use axum::routing::{get, post};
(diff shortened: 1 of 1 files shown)


crates/kumo-server-common/src/http_server/mod.rs
```diff
@@ -229,9 +230,37 @@ where
     }
 }
 
-async fn report_metrics(_: TrustedIpRequired) -> Result<String, AppError> {
-    let report = prometheus::TextEncoder::new()
-        .encode_to_string(&prometheus::default_registry().gather())?;
+#[derive(Deserialize)]
+struct PrometheusMetrics {
```
wez: Let's make it clear that these are parameters by adjusting the name

```suggestion
struct PrometheusMetricsParams {
```

crates/kumo-server-common/src/http_server/mod.rs
```diff
@@ -229,9 +230,37 @@ where
...
+    #[serde(default, deserialize_with = "empty_string_as_none")]
```
wez: I think handling the empty string case at the serde layer is more complex than is needed; I'd suggest not doing it here, and perhaps not worrying about it at all

```suggestion
    #[serde(default)]
```

crates/kumo-server-common/src/http_server/mod.rs
```diff
@@ -229,9 +230,37 @@ where
...
+    T: FromStr,
+    T::Err: fmt::Display,
+{
+    let opt = Option::<String>::deserialize(de)?;
+    match opt.as_deref() {
+        None | Some("") => Ok(None),
+        Some(s) => FromStr::from_str(s).map_err(de::Error::custom).map(Some),
+    }
+}
+
```
wez:
```suggestion

```

crates/kumo-server-common/src/http_server/mod.rs
```diff
@@ -229,9 +230,37 @@ where
...
+async fn report_metrics(
+    _: TrustedIpRequired,
+    Query(params): Query<PrometheusMetrics>,
```
wez:
```suggestion
    Query(params): Query<PrometheusMetricsParams>,
```

crates/kumo-server-common/src/http_server/mod.rs
```diff
@@ -229,9 +230,37 @@ where
...
+) -> Result<String, AppError> {
+    let mut metrics = prometheus::default_registry().gather();
+    if let Some(prefix) = params.prefix {
```
wez, line 243: if you wanted to make the empty string case do no additional work, then I think you could do something like:

```rust
if let Some(prefix) = params.prefix.and_then(|s| if s.is_empty() { None } else { Some(s) }) {
```

however, I don't think it is worth the extra code (which makes this harder to read and reason about) because you can simply not set the prefix parameter at all to achieve that result, and having an empty string prefix doesn't break anything, it is just slightly more expensive.

cai-n:
Hi, you're right, the serde was a bit over kill. I've committed changes following your comments

wez:
Thanks!

PR #190, opened 2024-05-29
Switch to debian bookwork for runtime docker image 

dschaaff:
This ensures the same os base as cargo chef, which in theory should help minimize breaking changes. It also fixes a conflict with uid/gid 1000 that was present in the ubuntu base due to the ubuntu user/group being uid/gid 1000.


wez:
Thanks!

PR #194, opened 2024-06-03
ready_queue: move maintainer notify logic in dispatcher drop

cai-n:
The maintainer wasn't notified if max_deliveries_per_connection was hit

### DIFF
--- a/crates/kumod/src/ready_queue.rs
+++ b/crates/kumod/src/ready_queue.rs
@@ -719,8 +719,6 @@ impl Drop for Dispatcher {
         // Ensure that we re-queue any message that we had popped
         if let Some(msg) = self.msg.take() {
             let activity = self.activity.clone();
-            let name = self.name.to_string();
-            let notify_dispatcher = self.notify_dispatcher.clone();
             READYQ_RUNTIME
                 .spawn_non_blocking("Dispatcher::drop".to_string(), move || {
                     Ok(async move {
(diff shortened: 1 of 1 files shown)


crates/kumod/src/ready_queue.rs
```diff
@@ -729,18 +727,26 @@ impl Drop for Dispatcher {
...
-                                if let Some(q) = ready_queue {
-                                    q.notify_maintainer.notify_one();
-                                }
-                                notify_dispatcher.notify_one();
                             }
                         }
                     })
                 })
                 .ok();
+        } else {
```
wez: Ah, I see what you're saying. I think that the shape of the logic in here could be tweaked a bit to look roughly like this:

```rust
fn drop(&mut self) {
   let msg = self.msg.take();
   READYQ_RUNTIME
                .spawn_non_blocking("Dispatcher::drop".to_string(), move || {
             Ok(async move {   
                 if let Some(msg) = msg {
                       if activity.is_shutting_down() {
                            Queue::save_if_needed_and_log(&msg).await;
                       } else {
                            if let Err(err) = Dispatcher::requeue_message(msg, false, None).await {
                                tracing::error!("error requeuing message: {err:#}");
                             }
                       }
                 } else {
                        tokio::time::sleep(Duration::from_secs(1)).await;
                        let ready_queue = ReadyQueueManager::get_by_name(&name);
                        if let Some(q) = ready_queue {
                            q.notify_maintainer.notify_one();
                        }
                        notify_dispatcher.notify_one();
                  }
          }
}

cai-n: I pushed an update, it looks better than my old copy paste :)

wez:
Thank you!

PR #217, opened 2024-06-18
use nanoseconds in redis lease durations

cai-n:
Use as_sec_f64 for more details, because the extend method fails if the expires value is the same. Also set redis-cell response to corresponding attributes.

### DIFF
--- a/crates/throttle/src/lib.rs
+++ b/crates/throttle/src/lib.rs
@@ -201,11 +201,11 @@ async fn redis_throttle(
         throttled: result[0] != 0,
         limit: result[1] as u64,
         remaining: result[2] as u64,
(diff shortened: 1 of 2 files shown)


crates/throttle/src/lib.rs
```diff
@@ -201,11 +201,11 @@ async fn redis_throttle(
         throttled: result[0] != 0,
         limit: result[1] as u64,
         remaining: result[2] as u64,
-        reset_after: Duration::from_secs(result[3].max(0) as u64),
-        retry_after: match result[4] {
+        retry_after: match result[3] {
```
wez, line 204: oof, that retry/reset mixup :-(
Thanks!

wez (approved):

PR #244, opened 2024-07-24
Patch bug in log_hooks.lua

aryeh:
I believe connection:close was inserted in the wrong place and should be inside the constructor function with the connection object.

Using version 2024.07.24-84bc2f05 causes crash when calling log_hooks:new_json


wez:
Yeah, good catch, thanks!

PR #248, opened 2024-08-02
swap label of scheduled_by_tenant_campaign

cai-n:
Use the same order for the labels and add metric removal at the same time as the removal of the schedule_queue metric.

### DIFF
--- a/crates/kumod/src/queue.rs
+++ b/crates/kumod/src/queue.rs
@@ -105,8 +105,8 @@ impl ScheduledMetrics {
         };
         let by_tenant_campaign = match &components.campaign {
             Some(campaign) => Some(TENANT_CAMPAIGN_GAUGE.get_metric_with_label_values(&[
-                campaign,
                 components.tenant.as_ref().map(|s| s.as_ref()).unwrap_or(""),
+                campaign,
             ])?),
             None => None,
         };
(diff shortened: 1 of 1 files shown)


crates/kumod/src/queue.rs
```diff
@@ -142,6 +142,29 @@ impl ScheduledMetrics {
...
+
+        DELAY_GAUGE.remove_label_values(&[name]).ok();
+
+        match &components.campaign {
+            Some(campaign) => Some(TENANT_CAMPAIGN_GAUGE.remove_label_values(&[
+                components.tenant.as_ref().map(|s| s.as_ref()).unwrap_or(""),
+                campaign,
+            ])),
+            None => None,
+        };
```
wez: I deliberately omitted this because it would remove too much: this particular counter aggregates across multiple scheduled queues, so it shouldn't be removed by any one scheduled queue being aged out.

I think that we don't strictly need to remove metrics for this one, because there should be a reasonably well bounded combination of tenants and campaigns; the cardinality is much smaller than the potential maximum number of email domains, for example.

If we were to age it out, I think that there would need to be some kind of periodic background task that checked for all combinations of tenant+campaign to decide when it would be appropriate to remove any given instance of this counter.

```suggestion
```

crates/kumod/src/queue.rs
```diff
@@ -105,8 +105,8 @@ impl ScheduledMetrics {
         };
         let by_tenant_campaign = match &components.campaign {
             Some(campaign) => Some(TENANT_CAMPAIGN_GAUGE.get_metric_with_label_values(&[
-                campaign,
                 components.tenant.as_ref().map(|s| s.as_ref()).unwrap_or(""),
+                campaign,
```
wez, line 109: oof, good catch!

crates/kumod/src/queue.rs
```diff
@@ -142,6 +142,29 @@ impl ScheduledMetrics {
         self.by_tenant.as_ref().map(|m| m.sub(amount));
         self.by_tenant_campaign.as_ref().map(|m| m.sub(amount));
     }
+
+    pub fn remove_metrics_for_queue(name: &str) {
+        let components = QueueNameComponents::parse(name);
```
wez: Can remove this bit because it was only used to facilitate the case that I think we should not be doing

```suggestion
```

crates/kumod/src/queue.rs
cai-n: ah yeah, my bad... I forgot that metrics of routing domains were also included.

wez:
Thank you!

PR #280, opened 2024-09-16
Fix compilation on macOS

djc:
Fixes some minor issues that prevented `cargo check` from working on macOS, and a newish warning.

I assume keeping it compiling on macOS is a non-goal (so checking it in CI is not worth it), but these changes seem small enough that they're probably still worth it? (Would be nice for me at least.)

### DIFF
--- a/crates/config/src/lib.rs
+++ b/crates/config/src/lib.rs
@@ -166,7 +166,7 @@ pub async fn load_config() -> anyhow::Result<LuaConfig> {
         };
 
         let _timer = latency_timer("context-creation");
-        func.call_async(()).await?;
+        func.call_async::<_, ()>(()).await?;
     }
     LUA_COUNT.increment(1.);
 
--- a/crates/kumo-server-common/src/disk_space.rs
+++ b/crates/kumo-server-common/src/disk_space.rs
@@ -141,8 +141,8 @@ impl MonitoredPath {
         let info = nix::sys::statvfs::statvfs(&self.path)
             .with_context(|| format!("statvfs({}) failed", self.path.display()))?;
 
-        let blocks_avail = info.blocks_available();
-        let blocks_total = info.blocks();
+        let blocks_avail = info.blocks_available() as u64;
+        let blocks_total = info.blocks() as u64;
 
         let space_avail_percent =
             ((blocks_avail as f64 / blocks_total as f64) * 100.0).floor() as u8;
@@ -156,7 +156,7 @@ impl MonitoredPath {
             .unwrap()
             .set(space_avail_percent as i64);
 
-        let inodes_avail = info.files_available();
+        let inodes_avail = info.files_available() as u64;
         let inodes_total = info.files();
         let inodes_avail_percent =
             ((inodes_avail as f64 / inodes_total as f64) * 100.0).floor() as u8;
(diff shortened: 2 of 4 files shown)


crates/proxy-server/Cargo.toml
```diff
@@ -13,4 +13,6 @@ libc = "0.2"
 log = "0.4"
 socksv5 = {version="0.3", default-features=false, features=["tokio"]}
 tokio = {workspace=true, features=["full", "tracing"]}
+
+[target.'cfg(target_os = "linux")'.dependencies]
```
djc, line 17:
```rust
error[E0425]: cannot find function `splice` in crate `libc`
  --> /Users/djc/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-splice-0.2.1/src/io.rs:40:15
   |
40 |         libc::splice(
   |               ^^^^^^ not found in `libc`
```

Perhaps these should be fixed in tokio-splice/libc, but since this crate already only calls the function on Linux, this seems actually correct.

crates/proxy-server/src/proxy_handler.rs
```diff
@@ -13,6 +13,7 @@ pub async fn handle_proxy_client(
     mut stream: TcpStream,
     peer_address: SocketAddr,
     timeout_duration: std::time::Duration,
+    #[cfg_attr(not(target_os = "linux"), allow(unused_variables))]
```
djc, line 16: I figured it would be better not to have the public interface change depending on platforms, with the downside that `no_splice = true` will be silently ignored on non-Linux platforms.

crates/kumo-server-common/src/disk_space.rs
```diff
@@ -141,8 +141,8 @@ impl MonitoredPath {
         let info = nix::sys::statvfs::statvfs(&self.path)
             .with_context(|| format!("statvfs({}) failed", self.path.display()))?;
 
-        let blocks_avail = info.blocks_available();
-        let blocks_total = info.blocks();
+        let blocks_avail = info.blocks_available() as u64;
```
djc, line 144: `fsblkcnt_t` and `fsfilcnt_t` are defined as `u32` (instead of `u64`) on macOS. Widening here seems like easiest fix, and I used `as` since that's already what is being used here. Alternatively, relying on `impl From<u32> for u64` would be slightly safer.

crates/config/src/lib.rs
```diff
@@ -166,7 +166,7 @@ pub async fn load_config() -> anyhow::Result<LuaConfig> {
         };
 
         let _timer = latency_timer("context-creation");
-        func.call_async(()).await?;
+        func.call_async::<_, ()>(()).await?;
```
djc, line 169:
```rust
warning: this function depends on never type fallback being `()`
   --> crates/config/src/lib.rs:119:1
    |
119 | pub async fn load_config() -> anyhow::Result<LuaConfig> {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #123748 <https://github.com/rust-lang/rust/issues/123748>
    = help: specify the types explicitly
note: in edition 2024, the requirement `!: FromLua<'_>` will fail
   --> crates/config/src/lib.rs:169:14
    |
169 |         func.call_async(()).await?;
    |              ^^^^^^^^^^
    = note: `#[warn(dependency_on_unit_never_type_fallback)]` on by default
```

djc:
Not sure why the `setup-cache` step is failing in the CI pipelines.

wez:
Thank you!

PR #282, opened 2024-09-18
throttle: add support for Redis installs without redis-cell module

### LINKED ISSUE #275: Introduce alternate throttle implementation that doesn't use redis-cell

djc:
Uses a Lua script for GCRA from https://github.com/Losant/redis-gcra/blob/master/lib/gcra.lua to perform the necessary logic without support for the redis-cell module that provides `CL.THROTTLE`.

Fixes #275.

### DIFF
--- a/crates/throttle/src/lib.rs
+++ b/crates/throttle/src/lib.rs
@@ -2,21 +2,21 @@
 //! The implementation uses an in-memory store, but can be adjusted in the future
 //! to support using a redis-cell equipped redis server to share the throttles
 //! among multiple machines.
-#[cfg(feature = "impl")]
+#[cfg(feature = "redis")]
 use mod_redis::{RedisConnection, RedisError};
-#[cfg(feature = "impl")]
+#[cfg(feature = "redis")]
 use once_cell::sync::OnceCell;
 use serde::{Deserialize, Serialize};
 use std::convert::TryFrom;
 use std::time::Duration;
 use thiserror::Error;
 
-#[cfg(feature = "impl")]
+#[cfg(feature = "redis")]
 pub mod limit;
-#[cfg(feature = "impl")]
+#[cfg(feature = "redis")]
 mod throttle;
 
-#[cfg(feature = "impl")]
+#[cfg(feature = "redis")]
 static REDIS: OnceCell<RedisConnection> = OnceCell::new();
 
 #[derive(Error, Debug)]
@@ -25,7 +25,7 @@ pub enum Error {
     Generic(String),
     #[error("{0}")]
     AnyHow(#[from] anyhow::Error),
-    #[cfg(feature = "impl")]
+    #[cfg(feature = "redis")]
     #[error("{0}")]
     Redis(#[from] RedisError),
     #[error("TooManyLeases, try again in {0:?}")]
@@ -44,7 +44,7 @@ pub struct ThrottleSpec {
     pub force_local: bool,
 }
 
-#[cfg(feature = "impl")]
+#[cfg(feature = "redis")]
 impl ThrottleSpec {
     pub async fn throttle<S: AsRef<str>>(&self, key: S) -> Result<ThrottleResult, Error> {
         self.throttle_quantity(key, 1).await
@@ -182,7 +182,7 @@ pub struct ThrottleResult {
     pub retry_after: Option<Duration>,
 }
 
-#[cfg(feature = "impl")]
+#[cfg(feature = "redis")]
 pub fn use_redis(conn: RedisConnection) -> Result<(), Error> {
     REDIS
         .set(conn)
--- a/crates/throttle/src/lib.rs
+++ b/crates/throttle/src/lib.rs
@@ -4,11 +4,10 @@
 //! among multiple machines.
 #[cfg(feature = "redis")]
 use mod_redis::{RedisConnection, RedisError};
-#[cfg(feature = "redis")]
-use once_cell::sync::OnceCell;
 use serde::{Deserialize, Serialize};
 use std::convert::TryFrom;
 use std::time::Duration;
+use std::sync::OnceLock;
 use thiserror::Error;
 
 #[cfg(feature = "redis")]
@@ -17,7 +16,7 @@ pub mod limit;
 mod throttle;
 
 #[cfg(feature = "redis")]
-static REDIS: OnceCell<RedisConnection> = OnceCell::new();
+static REDIS: OnceLock<RedisConnection> = OnceLock::new();
 
 #[derive(Error, Debug)]
 pub enum Error {
--- a/crates/throttle/src/lib.rs
+++ b/crates/throttle/src/lib.rs
@@ -3,23 +3,19 @@
 //! to support using a redis-cell equipped redis server to share the throttles
 //! among multiple machines.
 #[cfg(feature = "impl")]
-use mod_redis::{Cmd, FromRedisValue, RedisConnection, RedisError};
+use mod_redis::{RedisConnection, RedisError};
 #[cfg(feature = "impl")]
 use once_cell::sync::OnceCell;
-#[cfg(feature = "impl")]
-use redis_cell_impl::{time, MemoryStore, Rate, RateLimiter, RateQuota};
 use serde::{Deserialize, Serialize};
 use std::convert::TryFrom;
-#[cfg(feature = "impl")]
-use std::sync::Mutex;
 use std::time::Duration;
 use thiserror::Error;
 
 #[cfg(feature = "impl")]
 pub mod limit;
-
 #[cfg(feature = "impl")]
-static MEMORY: OnceCell<Mutex<MemoryStore>> = OnceCell::new();
+mod throttle;
+
 #[cfg(feature = "impl")]
 static REDIS: OnceCell<RedisConnection> = OnceCell::new();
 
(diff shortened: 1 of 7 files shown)


crates/throttle/src/throttle.rs
```diff
@@ -1,12 +1,74 @@
...
+  -- cost of 0 with remaining 0 is still limited
+  throttled = 1
+  remaining = 0
+  reset_after = math.ceil(tat - now)
+  retry_in = 0 -- retry_in is meaningless when quantity is 0
+else
+  throttled = 0
+  reset_after = math.ceil(new_tat - now)
+  retry_after = 0
+  redis.call("SET", key, new_tat, "PX", reset_after)
```
djc, line 63: This calls `SET` based on the data from the earlier `GET`, which I guess should really be some kind of compare-and-swap operation? It's not obvious to me yet how to solve that -- whether there is some support for CAS stuff in the Lua bindings, or if we should just retry a few times if there have been intermediate updates?

crates/throttle/src/throttle.rs
```diff
@@ -89,6 +151,38 @@ async fn redis_throttle(
...
+        .arg(period.as_secs())
+        .arg(max_burst)
+        .arg(quantity.unwrap_or(1));
+
+    let result =  conn.invoke_script(script).await.context("error invoking redis GCRA script")?;
+    let result = <(u64, u64, u64, u64, String, String) as FromRedisValue>::from_redis_value(&result)?;
+
+    Ok(ThrottleResult {
+        throttled: result.0 == 1,
+        limit: result.1, // TODO: this value is probably wrong
```
djc: Need to fix this, wasn't yet obvious what to put in here but I can root around some more.

crates/throttle/src/throttle.rs
```diff
@@ -89,6 +151,41 @@ async fn redis_throttle(
     })
 }
 
+async fn gcra_throttle(
```
djc: To do: configuration to wire this up.

djc (commented):
This hasn't seen any testing, just pushing up my current state for possible feedback.

crates/throttle/src/lib.rs
```diff
@@ -3,23 +3,20 @@
...
-use std::sync::Mutex;
 use std::time::Duration;
 use thiserror::Error;
 
 #[cfg(feature = "impl")]
 pub mod limit;
-
 #[cfg(feature = "impl")]
-static MEMORY: OnceCell<Mutex<MemoryStore>> = OnceCell::new();
+mod throttle;
```
djc, line 15: Not sure if the `impl` feature name makes a lot of sense to me, it seems like the main goal is to isolate code that relies on Redis, especially maybe for test purposes? At least this commit seems to help bring the number of guards down a bit.

crates/throttle/src/throttle.rs
wez: my understanding is that the script invocation is performed under a mutex that is held by the redis server to manage atomicity

djc: Ah, yes:

> Redis guarantees the script's atomic execution. While executing the script, all server activities are blocked during its entire runtime. These semantics mean that all of the script's effects either have yet to happen or had already happened.

That should take care of business, then!

crates/throttle/src/lib.rs
wez: the main goal was to allow the kumo-api-types to crate to import the `ThrottleSpec` type without pulling in redis itself

djc: Renamed the `impl` feature to `redis`.

crates/throttle/src/throttle.rs
```diff
@@ -0,0 +1,295 @@
...
+    period: Duration,
+    max_burst: u64,
+    quantity: Option<u64>,
+    force_local: bool,
+) -> Result<ThrottleResult, Error> {
+    if force_local {
+        local_throttle(key, limit, period, max_burst, quantity)
+    } else if let Some(redis) = REDIS.get().cloned() {
+        match redis.cl_throttle_command {
+            true => gcra_throttle(redis.connection, key, limit, period, max_burst, quantity).await,
```
wez: is this inverted? cl_throttle_command is true when cl.throttle is available

crates/throttle/src/throttle.rs
```diff
@@ -0,0 +1,295 @@
...
+        reset_after: Duration::from_secs(reset_after.max(0) as u64),
+        retry_after: if retry_after == -1 {
+            None
+        } else {
+            Some(Duration::from_secs(retry_after.max(0) as u64))
+        },
+    })
+}
+
+async fn redis_throttle(
```
wez: naming wise, both this and gcra_throttle are redis throttles, and both use gcra under the covers. Perhaps this should be renamed to redis_cl_throttle and gcra to something like redis_gcra_script_throttle to make the difference more obvious?

crates/throttle/src/lib.rs
wez: I'm not totally sold on renaming this, because now the feature name redis implies that it enables only redis related functions, but the current state of this PR also ties the in-memory implementation to the redis feature.

This is somewhat of an abstract point, because it doesn't really change anything today.

It would be nice though for the traffic-gen crate to be able to pull in the in-memory throttle impl without also pulling in the redis deps.

crates/throttle/src/lib.rs
```diff
@@ -2,34 +2,45 @@
...
-#[cfg(feature = "impl")]
+#[cfg(feature = "redis")]
 pub mod limit;
+#[cfg(feature = "redis")]
+mod throttle;
+
+#[derive(Clone)]
+struct RedisContext {
+    connection: RedisConnection,
+    cl_throttle_command: bool,
```
wez: nit: I think either `has_cl_throttle_command` or `cl_throttle_command_available` is slightly clearer. The former is less to type! `has_redis_cell` would also be acceptable and shorter.

djc:
@wez the new test now passes which leads me to think this is probably done? Let me know if you want to see more test coverage.

crates/throttle/src/throttle.rs
```diff
@@ -0,0 +1,366 @@
...
+    #[tokio::test]
+    async fn basic_throttle_60_000_burst_30k() {
+        // Note that the 5% tolerance here is the same as the basic_throttle_60_000
+        // test case because the variance is due to timing issues with very small
+        // time periods produced by the overally limit, rather than the burst.
+        test_big_limits(60_000, Some(30_000), 0.05, &*MEMORY).await;
+    }
+
+    #[tokio::test]
+    async fn redis_throttle_1_000() {
```
wez: can we have one version of this that tries to use redis cell, that skips if redis cell is unavailable, and another that explicitly uses the script form, even if redis cell is available?

crates/throttle/src/throttle.rs
```diff
@@ -0,0 +1,366 @@
...
+    let result =
+        <(u64, u64, u64, u64, String, String) as FromRedisValue>::from_redis_value(&result)?;
+
+    Ok(ThrottleResult {
+        throttled: result.0 == 1,
+        limit: max_burst + 1,
+        remaining: result.1,
+        reset_after: Duration::from_secs(result.2),
+        retry_after: match result.3 {
+            0 => None,
```
wez: the `redis_cell_throttle` implementation of this above handles the case where the number is negative, and also has a different representation for 0.

Which is right? TBH, I think neither is quite right!

The logic that consumes the throttles assumes that if `retry_after.is_some()`, then we were throttled, and will break out of the throttle checking loop otherwise.  It's probably best to formally encode that here for the various implementations and ensure that this is `None` if we are not throttled, and otherwise set this to the returned interval, clamping negative values to 0 so that we can't panic or return overflown/wrapped unsigned durations for this case.

I confess to not knowing whether it is possible to get a 0 second result here if we are throttled. That might be an awkward edge case that produces some busier looping in the consumer, so I'm open to considering clamping to a minimum of 1 second in the throttled case.

What do you think?

crates/throttle/src/throttle.rs
djc: Done.

crates/throttle/src/throttle.rs
djc: Hmm, I noticed that `ThrottleResult::retry_after` is actually the only field from that type that gets used outside the crate. Should we just get rid of everything else?

Will dig more into the details of this tomorrow, my brain is a little too fried to go deep on this now.

wez: preserving the other fields seems useful to me, if for example, we wanted to have those automatically translate into the conventional set of http throttling headers in the response to an http request.

djc: For the script throttle, `retry_after` has `math.ceil()` applied to it, so I don't think it can return 0 when throttling happens (in that case diff must be non-zero because remaining is not 0). I think that also means it's always more than 0 in the throttled case. I've made the code a bit more similar to the `redis_cell_throttle()`.

Not sure what else to do here? Feel free to make changes to the PR if you just want to get it merged.

crates/throttle/src/lib.rs
```diff
@@ -215,15 +215,15 @@ impl TryFrom<&str> for ThrottleSpec {
 #[derive(Debug, Eq, PartialEq, Serialize)]
 pub struct ThrottleResult {
     /// true if the action was limited
-    pub throttled: bool,
+    pub(crate) throttled: bool,
```
wez: I'd like to keep these in the public interface for future use

djc:
> Looks like there's an issue with running `make test`

You mean this, right?

Screenshot 2024-09-21 at 08 56 01

This confuses since I have a commit that should address this ("mod-redis: derive Debug for RedisConnection").

wez:
looks like some kind of weird issue with the CI; pulling and building it myself passes the tests just fine, so let's merge this!
Thank you!

PR #288, opened 2024-09-23
Replace lazy-static and once_cell with std types

djc:


wez:
Looks good, thanks!
I didn't think that this should go in before the earlier redis PR, so this now has a merge conflict. Please fix that up and we can merge it!

PR #292, opened 2024-09-25
Add support for SPF validation

djc:


wez:
just fyi: the CI appeared to get stunned by a locking issue around its default sqlite database. I switched it to postgres, which reset its contents, so the check status shown here in github will be wonky until the next push to this pr.

djc:
Rebased.

PR #295, opened 2024-09-26
Update shaping.toml

Solmea:
* shaping rules for gmail.com
* shaping rules for outlook.com
* shaping rules for yahoo.com
* shaping rules for web.de
* shaping rules for orange.fr
* shaping rules for qq.com
* shaping rules for 163.com

Tested on a production server running  2024.06.10-84e84b89

### DIFF
--- a/assets/community/shaping.toml
+++ b/assets/community/shaping.toml
@@ -27,14 +27,6 @@
 # kumo.on('get_egress_path_config', shaper.get_egress_path_config)
 
 #### BEGIN COMMUNITY CONTRIBUTED SHAPING RULES
-# There are some policies in the standard shaping.toml, depending on your install 
-# you might not have all of them which are in the lastest versions
-
-# Fallback to plain text delivery if TLS is not supported
-[["default".automation]]
-regex = "KumoMTA internal: failed to connect to any candidate hosts: All failures are related to OpportunisticInsecure STARTTLS. "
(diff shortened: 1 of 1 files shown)


MHillyer:
We expect the default shaping to be used in combination with the community one, so there is no need to reproduce the default automation, can you remove it?

Solmea:
I think I removed all defaults . Also removed the outlook.com default as that is also in the policy-extra shaping toml

assets/community/shaping.toml
```diff
@@ -26,4 +26,264 @@
...
+[["outlook.com".automation]]
+regex = "Temporary server error. Please try again later"
+trigger = {Threshold="5/hr"}
+action = "Suspend"
+duration = "1 hour"
+
+#===== orange.fr
+["orange.fr"]
+max_deliveries_per_connection = 75
+connection_limit = 1
```
Solvik:
```suggestion
connection_limit = 2
```

The postmaster page says: `Sending limits: Orange currently allows up to two simultaneous connections per sending IP address, and up to 100 recipients per message`

Solmea: Thanks, good call. I did not check the postmastery page again.. But usually in my experience orange was very strict on sending too much/fast.

MHillyer (approved):

PR #303, opened 2024-10-15
A trait-based resolver API

djc:
@wez what do you think of this direction? Might make sense to review the last commit first, the rest is sort of smaller/obvious stuff to get to this. As next steps, I'd add an `impl` feature for the dns-resolver crate so that most dependencies don't need to pull in libunbound, then I can fold the crate-specific `Lookup` traits into the main `Resolver` trait.

### DIFF
--- a/crates/dns-resolver/src/lib.rs
+++ b/crates/dns-resolver/src/lib.rs
@@ -1,4 +1,3 @@
-use crate::resolver::Resolver;
 use arc_swap::ArcSwap;
 use hickory_resolver::error::ResolveResult;
 pub use hickory_resolver::proto::rr::rdata::tlsa::TLSA;
@@ -13,7 +12,8 @@ use std::net::{IpAddr, Ipv6Addr};
 use std::sync::{Arc, LazyLock, Mutex as StdMutex};
 use std::time::Instant;
 
-pub mod resolver;
+mod resolver;
+pub use resolver::Resolver;
 
 static RESOLVER: LazyLock<ArcSwap<Resolver>> =
     LazyLock::new(|| ArcSwap::from_pointee(default_resolver()));
--- a/crates/dns-resolver/src/lib.rs
+++ b/crates/dns-resolver/src/lib.rs
@@ -14,6 +14,8 @@ use std::time::Instant;
 
 mod resolver;
 pub use resolver::Resolver;
+#[cfg(feature = "default-unbound")]
+pub use resolver::UnboundResolver;
 
 static RESOLVER: LazyLock<ArcSwap<Resolver>> =
     LazyLock::new(|| ArcSwap::from_pointee(default_resolver()));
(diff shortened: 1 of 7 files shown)


crates/dns-resolver/src/lib.rs
```diff
@@ -17,8 +17,8 @@ mod resolver;
 pub use resolver::UnboundResolver;
 pub use resolver::{HickoryResolver, Resolver};
 
-static RESOLVER: LazyLock<ArcSwap<Resolver>> =
-    LazyLock::new(|| ArcSwap::from_pointee(default_resolver()));
+static RESOLVER: LazyLock<ArcSwap<Box<dyn Resolver>>> =
```
djc, line 24: The `ArcSwap<Box<_>>` setup is a little weird. This is necessary because `ArcSwap` requires a `Sized` type. I got this trick from [their docs](https://docs.rs/arc-swap/latest/arc_swap/docs/limitations/index.html#sized-types):

> This currently works only for Sized types. Unsized types have „fat pointers“, which are twice as large as the normal ones. The [AtomicPtr](https://doc.rust-lang.org/nightly/core/sync/atomic/struct.AtomicPtr.html) doesn’t support them. One could use something like AtomicU128 for them. The catch is this doesn’t exist and the difference would make it really hard to implement the debt storage/stripped down hazard pointers.
> 
> A workaround is to use double indirection:

In the context of doing a DNS lookup, I don't think the extra pointer chasing will be a big deal?

djc:
(Also feeding experience back into upstream: https://github.com/hickory-dns/hickory-dns/pull/2509.)

crates/dns-resolver/src/lib.rs
wez: ah, makes sense; please add a comment here in the code about that, because it looks redundant and it was the first thing my brain wanted to comment on!

I agree, the double indirection should be tiny in cost given that we should only go through this on a cache miss and will need to go out to the network.

crates/dns-resolver/src/resolver.rs
```diff
@@ -47,79 +48,121 @@ impl Answer {
     }
 }
 
-pub enum Resolver {
-    Tokio(TokioAsyncResolver),
-    #[cfg(feature = "unbound")]
-    Unbound(AsyncContext),
+#[async_trait]
+pub trait Resolver: Send + Sync + 'static {
```
wez, line 52: I think that the shape of this trait looks right. I was thinking that this might end up being extracted into a `dns-resolver-trait` crate that is the minimal set of deps just for this, and that has some support for mocking/providing hard coded zone information as we discussed earlier.  Is that the ultimate plan?

djc: As mentioned in the OP, figured I might do an `impl` feature for the dns-resolver crate instead -- but happy to add a separate crate to your collection if you prefer.

crates/dns-resolver/src/lib.rs
djc: Will add a comment.

crates/dns-resolver/src/resolver.rs
wez: I think a separate crate will lend itself to potentially allowing others to reuse the dkim and spf crates independently from the rest of the stuff in this repo. That's not a hard requirement from my perspective, but a nice to have. For the sake of iteration speed, it's fine to keep it gated under a feature in the existing crate, because it can always be extracted later if we need to.

djc: Might even be enough to have the `unbound` feature? At least the trait still depends on hickory_proto types so that will need to be a dependency even for the separate trait (but I guess some of the other things wouldn't need to be).

djc: Makes sense -- either way, I think this is now in a good state if you want to merge it.

wez:
Thanks!

PR #304, opened 2024-10-16
Discard crate-specific DNS resolver traits

djc:


crates/dkim/src/errors.rs
```diff
@@ -60,6 +61,8 @@ pub enum DKIMError {
     MailParsingError(#[from] mailparsing::MailParsingError),
     #[error("Canonical CRLF line endings are required for correct signing and verification")]
     CanonicalLineEndingsRequired,
+    #[error("{0}")]
```
wez: a good candidate for transparent?  Alternatively, should we prefix this to indicate that it was a dns specific error as additional context, like we do with the other variants?

```suggestion
    #[error(transparent)]
```

crates/dns-resolver/src/resolver.rs
```diff
@@ -48,14 +110,243 @@ impl Answer {
...
+            .map_err(|err| DnsError::InvalidName(format!("invalid name {name}: {err}")))?;
         self.resolve(name, RecordType::TXT).await
     }
 
-    async fn resolve(&self, name: Name, rrtype: RecordType) -> anyhow::Result<Answer>;
+    async fn resolve(&self, name: Name, rrtype: RecordType) -> Result<Answer, DnsError>;
+}
+
+#[derive(Default)]
+pub struct TestResolver {
```
wez, line 145: thinking ahead for the higher level integration test use case, where it is desirable to shim in the zone/txt records from the kumod configuration, one way to achieve that is to have a version of `TestResolver` that includes an `Option<Box<dyn Resolver>>` that is consulted as a fallback when no records are found.

I don't thinks that changes what you are doing here in this PR, I'm just "thinking out loud"

wez (commented):
I think this looks good, but I lost internet for a a couple of hours part way through and went on a side quest to troubleshoot why my backup connection stopped working, and I think I was in the middle of checking something to make a suggestion just before that happened, but have forgotten what it was!
I'll come back and re-review later!

crates/dkim/src/errors.rs
djc: `DnsError` already includes "DNS" markers:

```rust
#[derive(Clone, Debug, Error, PartialEq)]
pub enum DnsError {
    #[error("invalid DNS name: {0}")]
    InvalidName(String),
    #[error("DNS: {0}")]
    ResolveFailed(String),
}
```

So making this `transparent` makes sense to me, will take care of that now.

crates/dns-resolver/src/resolver.rs
djc: I don't yet understand the integration test setup well enough to speculate on this, but obviously there's a bunch of stuff we can do inside this. So far I think the deduplication of all the test resolvers and the resolver adapater [sic] have worked out pretty well...

wez:
Thanks!

PR #305, opened 2024-10-16
switch smtp client from webpki-roots to rustls-platform-verifier

wez:
We had some feedback that it would be nice to be able to influence the set of ca-certs allowed by the client, which isn't possible when using the hard-coded list in webpki-roots.

This commit is a fairly dumb application of the example shown in the readme for rustls-platform-verifier.

I don't know if there is more nuance or something slightly different we should be doing with out setup here, so it would be great to have @djc review and suggest!


djc (approved):
LGTM. Just be aware that if you're deploying into a Linux/BSD environment that doesn't have any certificates installed (like the `ca-certificates` package on Debian), this will trust nothing. Not sure if that will be an issue?

wez:
re: ca-certificates, I think that sounds reasonable. We do suggest installing that package in the docs, and we also include it in our docker image. We could go and add it as a dep in our .deb and and the equivalent for rpm systems, but I'm hoping to avoid a hard dependency.

One thing that I'd like to do here is expand the `TlsOptions` struct in this crate to accept an optional list of files and/or directories and apply that via https://docs.rs/rustls-platform-verifier/latest/rustls_platform_verifier/struct.Verifier.html#method.new_with_extra_roots which I think will mitigate any surprises/pain caused by ca-certificates either not being there or somehow being undesirable as a hard dep.

That also needs to be applicable to the openssl impl in here also, so it can't simply require a list of `TrustAnchor` in the crate api.

djc:
> One thing that I'd like to do here is expand the TlsOptions struct in this crate to accept an optional list of files and/or directories and apply that via https://docs.rs/rustls-platform-verifier/latest/rustls_platform_verifier/struct.Verifier.html#method.new_with_extra_roots which I think will mitigate any surprises/pain caused by ca-certificates either not being there or somehow being undesirable as a hard dep.

That makes sense. Note that in 0.3, this API is Linux/BSD only (so would not compile for me on macOS), but 0.4 is in the process of being released.

djc:
Thinking about this more, assuming the binary size isn't a big deal it might make sense to just pass the webpki-roots to `new_with_extra_roots()`, which means you could largely avoid the need to install OS packages.

Also, I noticed that you're building a `ClientConfig` from scratch for every new client connection? The process for grabbing certificates from disk on Linux (alikes) can be a little expensive, so might make sense to hold on to the `ClientConfig` between config reloads? (There is a potential issue where you'd want to reload the certificates from disk every once in a while, so there's trade-offs...)

wez:
Ah, the reload is a complicating factor. I think I'll update the cache in the current state of the PR to use the same lruttl that we use for the dns resolver crate and set a max of something like 15 minutes, which should amortize the cost of loading the certs nicely while keeping the time window to pick up revocations (if applicable) and other updates to a reasonable small time frame.

crates/rfc5321/src/tls.rs
```diff
@@ -1,11 +1,72 @@
...
+        }
+    }
+}
+
+static RUSTLS_CACHE: LazyLock<Mutex<LruCacheWithTtl<RustlsCacheKey, Arc<ClientConfig>>>> =
+    LazyLock::new(|| Mutex::new(LruCacheWithTtl::new(32)));
+
+impl RustlsCacheKey {
+    fn get(&self) -> Option<Arc<ClientConfig>> {
+        RUSTLS_CACHE.lock().unwrap().get(self)
```
djc: Maybe consider using the parking_lot primitives? In addition to being faster they don't have the poisoning issue so you don't have to unwrap all over the place.

crates/rfc5321/src/tls.rs
```diff
@@ -1,11 +1,72 @@
...
+
+    fn set(self, value: Arc<ClientConfig>) {
+        RUSTLS_CACHE.lock().unwrap().insert(
+            self,
+            value,
+            // We allow the state to be cached for up to 15 minutes at
+            // a time so that we have an opportunity to reload the
+            // system certificates within a reasonable time frame
+            // as/when they are updated by the system.
+            std::time::Instant::now() + std::time::Duration::from_secs(15 * 60),
```
djc: Personally I prefer to import `Instant` and `Duration` directly unless there is other code in the crate using different types named `Instant`/`Duration`. (I'd also import `PartialEq`, `Eq` and `Hash`.)

crates/rfc5321/src/tls.rs
```diff
@@ -19,38 +80,55 @@ pub struct TlsOptions {
...
-            cipher_suites: cipher_suites.to_vec(),
-            ..provider::default_provider()
-        });
-
-        let config = ClientConfig::builder_with_provider(provider.clone())
-            .with_protocol_versions(tokio_rustls::rustls::DEFAULT_VERSIONS)
-            .expect("inconsistent cipher-suite/versions selected");
-
-        let config = if self.insecure {
+        let config = if let Some(config) = key.get() {
```
djc: Nit: I'd do `return TlsConnector::from(config);` here so that you can dedent the entirety of the `else` branch.

crates/rfc5321/src/tls.rs
```diff
@@ -19,38 +80,55 @@ pub struct TlsOptions {
...
+
+            let config = ClientConfig::builder_with_provider(provider.clone())
+                .with_protocol_versions(tokio_rustls::rustls::DEFAULT_VERSIONS)
+                .expect("inconsistent cipher-suite/versions selected");
+
+            let config = if self.insecure {
+                config
+                    .dangerous()
+                    .with_custom_certificate_verifier(Arc::new(
+                        danger::NoCertificateVerification::new(provider.clone()),
```
djc: Looks like you don't need the `clone()` here?

crates/rfc5321/src/tls.rs
```diff
@@ -19,38 +80,55 @@ pub struct TlsOptions {
...
+                &self.rustls_cipher_suites
+            };
+
+            let provider = Arc::new(CryptoProvider {
+                cipher_suites: cipher_suites.to_vec(),
+                ..provider::default_provider()
+            });
+
+            let config = ClientConfig::builder_with_provider(provider.clone())
+                .with_protocol_versions(tokio_rustls::rustls::DEFAULT_VERSIONS)
```
djc: Could instead do `with_safe_default_protocol_versions()`.

crates/rfc5321/src/tls.rs
```diff
@@ -19,38 +80,55 @@ pub struct TlsOptions {
...
+                .expect("inconsistent cipher-suite/versions selected");
+
+            let config = if self.insecure {
```
djc: Nit: consider hoisting the `config.dangerous().with_custom_certificate_verifier()` out of the `if`? Might need an explicit `as Arc<dyn ServerCertVerifier>`.

djc (approved):

Issue #83, opened 2023-09-13
SPF validation


wez:
* https://datatracker.ietf.org/doc/html/rfc7208 SPF
* https://datatracker.ietf.org/doc/html/rfc8601 Authentication-Results

NiuBlibing:
Congratulations

PR #306, opened 2024-10-21
SPF: add basic SPF Lua API

djc:
Next step in #83. Currently untested.

r? @wez

### DIFF
--- a/crates/integration-tests/spf-basic.lua
+++ b/crates/integration-tests/spf-basic.lua
@@ -1,0 +1,78 @@
+-- This config acts as a sink that capture all received mail into a maildir
+local kumo = require 'kumo'
+package.path = '../../assets/?.lua;' .. package.path
+local utils = require 'policy-extras.policy_utils'
+
+local TEST_DIR = os.getenv 'KUMOD_TEST_DIR'
+
+kumo.on('init', function()
+  kumo.configure_accounting_db_path ':memory:'
+
+  kumo.start_esmtp_listener {
+    listen = '127.0.0.1:0',
+    relay_hosts = { '0.0.0.0/0' },
+  }
+
+  kumo.configure_local_logs {
+    log_dir = TEST_DIR .. '/logs',
+    max_segment_duration = '1s',
+  }
+
+  kumo.define_spool {
+    name = 'data',
+    path = TEST_DIR .. '/data-spool',
+  }
+
+  kumo.define_spool {
+    name = 'meta',
+    path = TEST_DIR .. '/meta-spool',
+  }
+
+  zones = {
+    [[
+$ORIGIN localhost.
+@        600 TXT "v=spf1 all"
+denied       TXT "v=spf1 -all"
+allowed      TXT "v=spf1 +all"
+]],
+  }
+  kumo.dns.configure_test_resolver(zones)
+end)
+
+kumo.on('get_queue_config', function(_domain, _tenant, _campaign)
+  return kumo.make_queue_config {
+    protocol = {
+      maildir_path = TEST_DIR .. '/maildir',
+    },
+  }
+end)
+
+kumo.on('smtp_server_ehlo', function(domain, conn_meta)
+  local result = kumo.spf.check_host(domain, conn_meta, nil)
+  print('SPF-ehlo', kumo.json_encode_pretty(result))
+  if result.disposition == 'fail' then
+    kumo.reject(550, '5.7.1 SPF EHLO check failed')
+  end
+end)
+
+kumo.on('smtp_server_mail_from', function(sender, conn_meta)
+  local result =
+    kumo.spf.check_host(sender.domain, conn_meta, tostring(sender))
+  print('SPF-mail-from', kumo.json_encode_pretty(result))
+  if result.disposition == 'fail' then
+    kumo.reject(550, '5.7.1 SPF MAIL FROM check failed')
+  end
+end)
+
+kumo.on('smtp_server_message_received', function(msg, conn_meta)
+  msg:set_meta('queue', 'maildir')
+  local result = kumo.spf.check_host(
+    msg:sender().domain,
+    conn_meta,
+    tostring(msg:sender())
+  )
+  local results = {
+    result.result,
+  }
+  msg:add_authentication_results(msg:get_meta 'hostname', results)
+end)
--- a/crates/integration-tests/src/test/spf_basic.rs
+++ b/crates/integration-tests/src/test/spf_basic.rs
@@ -1,0 +1,85 @@
+use crate::kumod::{generate_message_text, KumoDaemon, MailGenParams};
+use rfc5321::{ClientError, Response};
+use std::time::Duration;
+
+#[tokio::test]
+async fn spf_basic() -> anyhow::Result<()> {
+    let mut daemon = KumoDaemon::spawn_with_policy("spf-basic.lua").await?;
+    let mut client = daemon.smtp_client("localhost").await?;
+    let body = generate_message_text(1024, 78);
+
+    // Send mail from `localhost`, which is allowed through `smtp_server_ehlo`
+    // but gets a temporary error from `smtp_server_message_received` because
+    // there no SPF policy is found for the sender's domain, `example.com`.
+
+    let response = MailGenParams {
+        body: Some(&body),
+        ..Default::default()
+    }
+    .send(&mut client)
+    .await?;
+    anyhow::ensure!(response.code == 250);
+
+    // Send mail from `denied.localhost`, which is denied
+
+    match daemon.smtp_client("denied.localhost").await {
+        Ok(_) => panic!("expected rejection"),
+        Err(err) => match err.downcast_ref::<ClientError>() {
+            Some(ClientError::Rejected(Response { code: 550, .. })) => {}
+            _ => panic!("expected ClientError"),
+        },
+    }
+
+    // Send email with an allowed sender
+
+    let response = MailGenParams {
+        body: Some(&body),
+        sender: Some("foo@allowed.localhost"),
+        ..Default::default()
+    }
+    .send(&mut client)
+    .await?;
+    anyhow::ensure!(response.code == 250);
+
+    daemon
+        .wait_for_maildir_count(1, Duration::from_secs(10))
+        .await;
+
+    daemon.stop().await?;
+    println!("Stopped!");
+
+    let delivery_summary = daemon.dump_logs()?;
+    k9::snapshot!(
+        delivery_summary,
+        "
+{
+    Reception: 2,
+    Delivery: 2,
+    Rejection: 1,
+}
+"
+    );
+
+    let mut entries = daemon.maildir().list_new().collect::<Result<Vec<_>, _>>()?;
+    assert_eq!(entries.len(), 2);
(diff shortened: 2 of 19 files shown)


docs/reference/kumo.spf/check_host.md
```diff
@@ -0,0 +1,40 @@
+# `kumo.spf.check_host()`
+
+{{since('2024.10.22-??')}}
```
wez: please use `dev` as the placeholder for the unreleased version; at release time I make a sweep for these and replace with the version of the newly minted tag.

```suggestion
{{since('dev')}}
```

crates/kumod/src/spf.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+impl UserData for CheckHostOutput {
+    fn add_fields<'lua, F: mlua::UserDataFields<'lua, Self>>(fields: &mut F) {
+        fields.add_field_method_get("disposition", |_, this| Ok(this.disposition.as_str()));
+        fields.add_field_method_get("result", |lua, this| {
+            Ok(lua.to_value_with(&this.result, serialize_options())?)
+        });
+    }
+}
```
wez: for this I would be inclined to make the entire output serializable, then `lua.to_value_with` the whole result. It results in less boilerplate here and the resultant native lua table is a bit more flexible to work with in lua if you want to do things using the `pairs` function to iterate it.

```suggestion
#[derive(Serialize)]
struct CheckHostOutput {
    disposition: SpfDisposition,
    result: AuthenticationResult,
}
```

wez:
btw, you can run `docs/build.sh` to run the docs build, and `docs/build.sh serve` to run a local web server to view it in a browser.

djc:
> btw, you can run `docs/build.sh` to run the docs build, and `docs/build.sh serve` to run a local web server to view it in a browser.

Ahh, that presupposes running Docker which I usually try to avoid. Also generates a bunch of diff:

```diff
diff --git a/docs/reference/kumod.openapi.json b/docs/reference/kumod.openapi.json
index db48fbbf..f6440451 100644
--- a/docs/reference/kumod.openapi.json
+++ b/docs/reference/kumod.openapi.json
@@ -6,7 +6,7 @@
     "license": {
       "name": "Apache-2.0"
     },
-    "version": "2024.10.15-3e4b6c13"
+    "version": "2024.10.15-46882b13"
   },
   "paths": {
     "/api/admin/bounce/v1": {
@@ -443,8 +443,7 @@
             "description": "The the preferred filename for the attachment",
             "nullable": true
           }
-        },
-        "additionalProperties": false
+        }
       },
       "BounceV1CancelRequest": {
         "type": "object",
@@ -682,8 +681,7 @@
             "example": "Sales",
             "nullable": true
           }
-        },
-        "additionalProperties": false
+        }
       },
       "Header": {
         "oneOf": [
@@ -744,8 +742,7 @@
           "trace_headers": {
             "$ref": "#/components/schemas/HttpTraceHeaders"
           }
-        },
-        "additionalProperties": false
+        }
       },
       "InjectV1Response": {
         "type": "object",
@@ -909,8 +906,7 @@
               "gender": "male"
             }
           }
-        },
-        "additionalProperties": false
+        }
       },
       "SetDiagnosticFilterRequest": {
         "type": "object",
```

wez:
I think you need to run `make test` so that your kumod is built and will output the same openapi spec as is already checked in

crates/integration-tests/src/test/spf_basic.rs
```diff
@@ -0,0 +1,43 @@
+use crate::kumod::{generate_message_text, KumoDaemon, MailGenParams};
+use std::time::Duration;
+
+#[tokio::test]
+async fn spf_basic() -> anyhow::Result<()> {
+    let mut daemon = KumoDaemon::spawn_with_policy("spf-basic.lua").await?;
```
wez, line 7: I think this lua file is missing from the commit

crates/kumod/src/smtp_server.rs
```diff
@@ -1822,6 +1822,15 @@ impl UserData for ConnectionMetaData {
...
 
+impl<'lua> FromLua<'lua> for ConnectionMetaData {
+    fn from_lua(value: mlua::Value<'lua>, lua: &'lua mlua::Lua) -> mlua::Result<Self> {
+        dbg!(&value);
+        Ok(Self {
+            map: Arc::new(Mutex::new(dbg!(lua.from_value(value)?))),
+        })
+    }
+}
+
```
wez: this impl likely conflicts with the regular UserData handling.  I think you should be fine to remove this?

djc: It didn't seem to work earlier. Will have another look in a few hours.

djc:
@wez I think this is ready for a round of review, seems to be in a decent shape.

crates/integration-tests/spf-basic.lua
```diff
@@ -0,0 +1,74 @@
...
+  kumo.define_spool {
+    name = 'meta',
+    path = TEST_DIR .. '/meta-spool',
+  }
+
+  zones = {}
+  zones[1] = '$ORIGIN localhost.\n'
+    .. '@        600 TXT "v=spf1 all"\n'
+    .. 'denied       TXT "v=spf1 -all"\n'
+    .. 'allowed      TXT "v=spf1 +all"\n'
```
wez: FWIW, for this sort of literal, you can use the `[[ ]]` style string literal:

```suggestion
  zones = {
[[
 $ORIGIN localhost.
 @        600 TXT "v=spf1 all"
 denied       TXT "v=spf1 -all"
 allowed      TXT "v=spf1 +all"
 ]]
 }
```

That would probably look nicer if it was indented; unsure if that is significant for the zone parser?

crates/integration-tests/spf-basic.lua
```diff
@@ -0,0 +1,74 @@
...
+      maildir_path = TEST_DIR .. '/maildir',
+    },
+  }
+end)
+
+kumo.on('smtp_server_ehlo', function(domain, conn_meta)
+  local result = kumo.spf.check_host(domain, conn_meta, nil)
+  print('SPF-ehlo', kumo.json_encode_pretty(result))
+  if result.disposition == 'fail' then
+    kumo.reject(550, '5.7.1 SPF MAIL FROM check failed')
```
wez:
```suggestion
    kumo.reject(550, '5.7.1 SPF EHLO check failed')
```

crates/integration-tests/spf-basic.lua
```diff
@@ -0,0 +1,74 @@
...
+
+kumo.on('smtp_server_message_received', function(msg, conn_meta)
+  msg:set_meta('queue', 'maildir')
+  local result = kumo.spf.check_host(
+    msg:sender().domain,
+    conn_meta,
+    tostring(msg:sender())
+  )
+  results = {}
+  results[1] = result.result
```
wez: I would just directly construct the array style table inline:

```suggestion
  local results = {
    result.result
  }
```

Also: you pretty much always want to use `local` for any variables you define, even those at the file scope, otherwise the variable gets created in the global environment table. From a VM perspective, the `local` variable gets compiled to an integer offset into the register table so you get an `O(1)` micro-optimization boost by avoiding the hash lookup, but the most important benefit of using `local` is that you don't accidentally pollute the global symbol table.

Also also: I literally just learned this about lua 5.4 a couple of days ago, but you can do:

```lua
local results <const> = {
  result.result
}
```

to create the equivalent of a js `const` or rust immutable binding.

I'm not sure how I feel about the readability of this, and it's not worth using here, but just wanted to share this for larger and more complex lua in the future.

crates/integration-tests/src/test/spf_basic.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    let mut entries = daemon.maildir().list_new().collect::<Result<Vec<_>, _>>()?;
+    assert_eq!(entries.len(), 2);
+    for entry in &mut entries {
+        let headers = entry.headers()?;
+        let from = headers.from().unwrap().unwrap().0;
+        let results = headers.authentication_results().unwrap().unwrap().results;
+        match from[0].address.domain.as_str() {
+            "example.com" => {
+                assert_eq!(results.len(), 1);
```
wez, line 71: could use k9::snapshot here? Not necessary, just reduces some manual typing, but may be more verbose than what you want to assert on.

crates/kumod/src/spf.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    spf_mod.set(
+        "check_host",
+        lua.create_async_function(
+            |lua, (domain, meta, sender): (String, UserDataRef<ConnectionMetaData>, Option<String>)| async move {
+                // `received_from` is always set, and always to a value representing a `SocketAddr`.
+                let addr = meta
+                    .get_meta("received_from")
+                    .and_then(|v| SocketAddr::from_str(v.as_str()?).ok())
+                    .unwrap();
```
wez: prefer to explain in an `expect` rather than in a comment ahead of a bare `unwrap`:

```suggestion
                let addr = meta
                    .get_meta("received_from")
                    .and_then(|v| SocketAddr::from_str(v.as_str()?).ok())
                    .expect("`received_from` is always set, and always to a value representing a `SocketAddr`");
```

wez (commented):
Just some minor nits really; this looks good, thank you!

crates/integration-tests/spf-basic.lua
djc: Right, thanks. I was looking for syntax like that but couldn't quickly find something that worked so decided to move on with something slightly uglier that obviously worked (based on other code). This is my first time using Lua, so all this is useful feedback. I do miss static feedback already, though...

crates/integration-tests/spf-basic.lua
djc: Did it without indentation to keep the zone file formatted as expected.

crates/integration-tests/src/test/spf_basic.rs
djc: It doesn't seem worth it? Quite a bit more verbose at seemingly limited benefit...

djc:
(Mostly) addressed your feedback, let me know if there's anything else -- I'll start working on DMARC.

wez:
Looks good, thank you!

wez:
one minor tweak I'm just going to commit and push here is that you can simplify this bit and let mlua handle the type conversion for you:

```diff
diff --git a/crates/mod-dns-resolver/src/lib.rs b/crates/mod-dns-resolver/src/lib.rs
index 823b6f30..2007205c 100644
--- a/crates/mod-dns-resolver/src/lib.rs
+++ b/crates/mod-dns-resolver/src/lib.rs
@@ -193,10 +193,10 @@ pub fn register(lua: &Lua) -> anyhow::Result<()> {

     dns_mod.set(
         "configure_test_resolver",
-        lua.create_function(move |lua, zones: mlua::Value| {
+        lua.create_function(move |_lua, zones: Vec<String>| {
             let mut resolver = TestResolver::default();
-            for zone in lua.from_value::<Vec<String>>(zones)? {
-                resolver = resolver.with_zone(&zone);
+            for zone in &zones {
+                resolver = resolver.with_zone(zone);
             }

             dns_resolver::reconfigure_resolver(resolver);
```

Issue #309, opened 2024-10-29
debug `latest` tag on docker image in release process


wez:
The latest release didn't add `latest` on the merged multiarch image, let's ensure that the next one does do that

wez:
This is also happening to dev builds, so that makes it a bit easier to iterate on and fix

PR #310, opened 2024-10-31
Revert "ci: avoid deprecated syntax"

wez:
This reverts commit 401df300998096ab208417f4f68b75ef1fdc3d03.

The docker merge step is failing:

```
envconfig.Process: assigning PULLREQUEST_DRONE_PULL_REQUEST to Number: converting '' to type int. details: strconv.ParseInt: parsing "": invalid syntax
```

It seems that the compatibility change in
https://github.com/woodpecker-ci/woodpecker/pull/3939 isn't sufficient.

Let's see if we can still run with the environment set manually, or if we'll need to blackjack our own woodpecker plugin for this.

https://github.com/KumoCorp/kumomta/issues/309


wez:
Unfortunately, we can't tell from the PR mode whether this is good or not, so let's merge and see what happens.

PR #312, opened 2024-11-07
change orange.fr to provider.orange in shaping.toml

cai-n:
change orange shaping config to provider to include their business domains with MX smtppromx.orange.fr. that has the same IP as orange.fr MXs.


wez:
Thanks!

PR #313, opened 2024-11-08
Update dependencies

djc:


djc:
doc-image-builder:

```
Step 4/13 : RUN rustup target add x86_64-unknown-linux-musl
 ---> Running in 8de0f5cfee36
info: downloading component 'rust-std' for 'x86_64-unknown-linux-musl'
info: installing component 'rust-std' for 'x86_64-unknown-linux-musl'
Removing intermediate container 8de0f5cfee36
 ---> b8e07d66d0b5
Step 5/13 : RUN cargo build --target x86_64-unknown-linux-musl
 ---> Running in 0db74591f919
error: failed to parse manifest at `/toml2jsonc/Cargo.toml`

Caused by:
  error inheriting `anyhow` from workspace root manifest's `workspace.dependencies.anyhow`

Caused by:
  failed to find a workspace root
The command '/bin/sh -c cargo build --target x86_64-unknown-linux-musl' returned a non-zero code: 101
exit status 101
```

Is this somehow built without the workspace context? Do you want to keep it that way?

djc:
ci/pr/amazon/2 failure in `test`:

```
------------
     Summary [  24.369s] 264 tests run: 262 passed, 2 failed, 0 skipped
        FAIL [  14.672s] integration-tests::bin/integration-tests test::retry_schedule::retry_schedule_singleton_wheel
        FAIL [  13.974s] integration-tests::bin/integration-tests test::retry_schedule::retry_schedule_skiplist
error: test run failed
make: *** [Makefile:26: test] Error 100
```

```
thread 'test::retry_schedule::retry_schedule_skiplist' panicked at crates/integration-tests/src/test/retry_schedule.rs:84:5:
first is 2 but should be ~0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

```
Stopped!
***** event_times: [2, 8]
test test::retry_schedule::retry_schedule_singleton_wheel ... FAILED
```

These tests seem to have passed in other environments... have you seen this before? Any hunch on what could trigger this?

djc:
> if we can keep that one self-contained, I think it will help to keep the corresponding Dockerfile simpler, but I'm not opposed to seeing an alternative viewpoint on that.

That makes sense to me, reverted the change for that crate.

djc:
Some notes from ploughing through the dependencies:

* There are dependencies on slog, log, env_logger in there, should those be moved over to tracing?
* There's async-channel and flume, would it make sense to port async-channel cases over to flume?
* There's both zstd and zstd-safe in there, could we pick just one?
* toml is just a thin wrapper around toml_edit at this point IIRC, maybe skip the wrapper?

wez:
> ```
> Stopped!
> ***** event_times: [2, 8]
> test test::retry_schedule::retry_schedule_singleton_wheel ... FAILED
> ```
> 

> These tests seem to have passed in other environments... have you seen this before? Any hunch on what could trigger this?

Yeah, it's an occasional flake in the CI system; you can ignore it here.

wez:
re: `slog`, I think that isn't actually used directly and can be removed from the message crate
(I think that might have been a hangover from an earlier incarnation of the dkim crate)

re: `log` vs. `tracing`, I think it is just the proxy-server using `log`.  IMO, `tracing` is horrible to setup for an application and I'm happy to stick with `log` in the proxy-server for the time being to keep it nice and simple.
Similar thoughts for the other smaller utilities.

re: rustls version and reqwest, I'm less concerned about that for eg: the kcli usage of reqwest, but I think it would be valuable for the client that we wrap up in kumod's lua http client.

djc:
Current failures:

```
     Summary [  56.446s] 264 tests run: 260 passed, 4 failed, 0 skipped
        FAIL [  26.103s] integration-tests::bin/integration-tests test::log_oob_arf::log_oob_arf
        FAIL [  16.878s] integration-tests::bin/integration-tests test::suspend_delivery_ready_q_and_deliver::suspend_delivery_ready_q_and_deliver
        FAIL [  18.301s] integration-tests::bin/integration-tests test::suspend_delivery_scheduled_q_and_deliver::suspend_delivery_scheduled_q_and_deliver
        FAIL [  11.701s] integration-tests::bin/integration-tests test::temp_fail::temp_fail
error: test run failed
make: *** [Makefile:26: test] Error 100
```

wez:
re: `async-channel` vs `flume`. I have it in my TODO to review https://github.com/chadaustin/batch-channel and potentially standardize on that one, so I'd suggest holding off on making changes here for the time being.

djc:
Okay, so I think this is done for now?

wez:
Yeah, I think this looks good. We're planning on cutting a stable tag probably Mon/Tue next week. I'll merge this after I tag and push that release, just in case there are any unexpected side effects of some of these package bumps.

PR #324, opened 2024-12-18
add send_batch method to kafka mod

cai-n:
Add a new method to the kafka mod to also send with rdkafka's internal batching. The return is a bit strange because I couldn't figure out how to send the failed indexes as LuaError that could then be used in the lua part if it was caught with a pcall...

### DIFF
--- a/crates/mod-kafka/Cargo.toml
+++ b/crates/mod-kafka/Cargo.toml
@@ -12,4 +12,3 @@ mlua = {workspace=true, features=["vendored", "lua54", "async", "send", "seriali
 rdkafka.workspace=true
 serde.workspace=true
 tokio.workspace=true
-tracing.workspace=true
--- a/crates/mod-kafka/Cargo.toml
+++ b/crates/mod-kafka/Cargo.toml
@@ -10,3 +10,6 @@ duration-serde = {path="../duration-serde"}
 mlua = {workspace=true, features=["vendored", "lua54", "async", "send", "serialize"]}
 rdkafka = {workspace=true}
 serde = {workspace=true}
+futures = {workspace=true}
+tokio = {workspace=true, features=["full"]}
+tracing = {workspace=true}
--- a/crates/mod-kafka/Cargo.toml
+++ b/crates/mod-kafka/Cargo.toml
@@ -4,12 +4,12 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-anyhow = {workspace=true}
+anyhow.workspace=true
 config = {path="../config"}
 duration-serde = {path="../duration-serde"}
+futures.workspace=true
 mlua = {workspace=true, features=["vendored", "lua54", "async", "send", "serialize"]}
-rdkafka = {workspace=true}
-serde = {workspace=true}
-futures = {workspace=true}
-tokio = {workspace=true, features=["full"]}
-tracing = {workspace=true}
+rdkafka.workspace=true
+serde.workspace=true
+tokio.workspace=true
+tracing.workspace=true
--- a/crates/mod-kafka/src/lib.rs
+++ b/crates/mod-kafka/src/lib.rs
@@ -1,4 +1,6 @@
 use config::{any_err, get_or_create_sub_module};
+use futures::stream::FuturesOrdered;
+use futures::StreamExt;
 use mlua::prelude::LuaUserData;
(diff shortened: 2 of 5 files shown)


crates/mod-kafka/Cargo.toml
```diff
@@ -10,3 +10,6 @@ duration-serde = {path="../duration-serde"}
 mlua = {workspace=true, features=["vendored", "lua54", "async", "send", "serialize"]}
 rdkafka = {workspace=true}
 serde = {workspace=true}
+futures = {workspace=true}
+tokio = {workspace=true, features=["full"]}
+tracing = {workspace=true}
```
wez: Please sort these deps so that they are alphabetized.
I know that most of these are using `= {workspace = true}`, but I prefer this simpler form if we don't need to specify any additional parameters:

```suggestion
futures.workspace=true
tokio.workspace=true
tracing.workspace
```

for tokio: it is conceptually undesirable for a library crate to enable the full feature set of some other crate.  If we don't need to enable additional features, let's just use `.workspace = true`.  If we do need specific features let's just list the minimum necessary.

`cargo check -p mod-kafka` can be helpful to check whether the right features are enabled for just this specific crate.

crates/mod-kafka/src/lib.rs
```diff
@@ -94,6 +96,69 @@ impl LuaUserData for Producer {
...
+                        failed_indexes.push(index)
+                    }
+                }
+                index += 1;
+            }
+            if failed_indexes.is_empty() {
+                Ok(Value::Nil)
+            } else {
+                Ok(lua.to_value(&failed_indexes)?)
+            }
```
wez: What do you think about doing it this way?

```suggestion
            let mut failed_indexes = vec![];
            let mut errors = vec![];
            let mut index = 1;

            while let Some(result) = tasks.next().await {
                match result {
                    Ok(Ok(_)) => {}
                    Ok(Err((error, _msg))) => {
                        errors.push(format!("{error:#}"));
                        failed_indexes.push(index);
                    }
                    Err(error) => {
                        errors.push(format!("{error:#}"));
                        failed_indexes.push(index);
                    }
                }
                index += 1;
            }
            Ok((failed_indexes, errors))
```

in the lua code you might do something like this:

```lua
local failed_items, errors = producer:send_batch(items)
if #failed_items > 0 then
   -- some items failed
   for i, item_idx in ipairs(failed_items) do
      local error = errors[i]
      print(string.format("item idx %d failed: %s", item_idx, error))
   end
end
```

I like not unilaterally emitting a `tracing::error` for this, because my assumption is that if any one in the batch is likely to fail, all of the batch is likely to fail and you can quickly flood the logs in that case.  Allowing you to control whether you log it from lua feels better.

wez (commented):
I was just wondering how you were getting on with this the other day!

crates/mod-kafka/src/lib.rs
cai-n: thanks for the feedback! I wasn't sure where the error should be logged, and you're right, having it in lua gives the user more control over it

crates/mod-kafka/Cargo.toml
```diff
@@ -4,9 +4,12 @@ version = "0.1.0"
...
 config = {path="../config"}
 duration-serde = {path="../duration-serde"}
+futures.workspace=true
 mlua = {workspace=true, features=["vendored", "lua54", "async", "send", "serialize"]}
-rdkafka = {workspace=true}
-serde = {workspace=true}
+rdkafka.workspace=true
+serde.workspace=true
+tokio.workspace=true
+tracing.workspace=true
```
wez: I think we can trim this out with the API change to return the errors instead of logging them

```suggestion
```

wez (commented):
Looks good!
Please also add docs for this to `docs/reference/kumo.kafka/build_producer.md`

You can use `{{ since('dev') }}` in the section that you add there to indicate that this is available in `dev`.

Thanks!

crates/mod-kafka/Cargo.toml
cai-n: good catch, I'll remove it

docs/reference/kumo.kafka/_index.md
```diff
@@ -1,4 +1,4 @@
-# Module `kumo.amqp`
+# Module `kumo.kafka`
```
wez, line 1: heh, good catch!

wez:
Thank you!

PR #328, opened 2025-01-27
change kafka payload type to vector of bytes

cai-n:
Change the payload to Vec<u8> to enable sending payloads with invalid utf8.


crates/toml2jsonc/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-anyhow = "1.0.86"
-clap = { version = "4.5.8", features = ["derive"] }
-serde_json = "1.0.120"
-toml_edit = "0.22"
+anyhow.workspace=true
```
wez: I'd suggest backing out the changes in this file, so that we don't have to wrangle the workspace into the cut down docker env that we use for docs

cai-n: Yeah, I didn't see that it was used to build a container. I ended up changing the Dockerfile to copy the whole workspace and only build toml2jsonc. The end image should be roughly the same size with the multi stage build though. But I'll totally revert the changes if I missed something else

wez: let's separate these changes into separate PRs; I don't see why we need the toml2json changes here for the kafka change.  I do appreciate tidying up this Cargo.toml, but splitting it out makes things a bit clearer and more self contained!

cai-n: I saw that and thought that I could also change this at the same time but you're right, I'll remove this part and push it in another PR, it will be tidier

wez:
Thank you!

PR #329, opened 2025-01-29
convert arrival-date header for oob

cai-n:
Parse arrival-date header to avoid: 
failed to convert 'Wed, 29 Jan 2025 13:47:40 +0000': input contains invalid characters
    stack traceback:
        [C]: in method 'parse_rfc3464'

### DIFF
--- a/crates/kumo-log-types/src/rfc3464.rs
+++ b/crates/kumo-log-types/src/rfc3464.rs
@@ -540,7 +540,7 @@ Some(
 
     #[test]
     fn rfc3464_4() {
-        let result = Report::parse(include_bytes!("../data/rfc3464/3.eml")).unwrap();
+        let result = Report::parse(include_bytes!("../data/rfc3464/4.eml")).unwrap();
         k9::snapshot!(
             result,
             r#"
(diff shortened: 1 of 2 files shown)


wez:
Ooh, interesting. Can you add a sample message to use as a test case? Simlar to to those in `crates/kumo-log-types/data/rfc3464`?

Also, I think the rfc5965 code may also need to handle this arrival-date header in a similar way

crates/kumo-log-types/src/rfc3464.rs
```diff
@@ -539,7 +540,7 @@ Some(
 
     #[test]
     fn rfc3464_4() {
-        let result = Report::parse(include_bytes!("../data/rfc3464/3.eml")).unwrap();
```
wez, line 542: homer

wez:
Looks like rfc5965 was already doing this properly:
https://github.com/KumoCorp/kumomta/blob/ea8f2be678ba1d591485cf483b4f06dcb2f6415a/crates/kumo-log-types/src/rfc5965.rs#L129-L133

wez:
Thank you!

PR #331, opened 2025-02-05
Add unmask to deb postinst

cai-n:
add unmask to enable switching between installs of kumomta and kumomta-dev and avoid: 
Failed to restart kumomta.service: Unit kumomta.service is masked.
Failed to restart kumo-tsa-daemon.service: Unit kumo-tsa-daemon.service is masked.

Also should the Conflicts field be changed to mail-transport-agent like described in the debian doc for mta? https://www.debian.org/doc/debian-policy/ch-customized-programs.html#mail-transport-delivery-and-user-agents

### DIFF
--- a/assets/build-deb.sh
+++ b/assets/build-deb.sh
@@ -77,6 +77,9 @@ if [ "\$1" = "configure" ]; then
     fi
 
     if [ -x "/usr/bin/deb-systemd-helper" ]; then
+      deb-systemd-helper unmask kumomta.service >/dev/null
+      deb-systemd-helper unmask kumo-tsa-daemon.service >/dev/null
+
       deb-systemd-helper enable kumomta.service >/dev/null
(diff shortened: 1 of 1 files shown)


wez:
> Also should the Conflicts field be changed to mail-transport-agent like described in the debian doc for mta? https://www.debian.org/doc/debian-policy/ch-customized-programs.html#mail-transport-delivery-and-user-agents

Good question. We don't really aim to replace a unix style local MTA service in the same way that sendmail or postfix operate, so I'm not sure how much of that is really applicable.

In particular, if we conflict with one of those, it might prevent someone from being able to configure a unix MDA to relay via a local kumomta installation.

I think it is more flexible for users, even if perhaps a little less "technically correct" from a debian packaging standpoint, to side step those local mail transport examples.

wez:
Thank you!

PR #344, opened 2025-02-21
Modify `kcli` command example to match required arguments

ryanjbonnell:
When executing the `kcli rebind` command, `--reason` is a [required argument](https://docs.kumomta.com/reference/rapidoc/?h=api#post-/api/admin/rebind/v1).

Running the command without a "reason" causes it to fail with an error message.

This copy edit to the documentation aligns the example command with its expected input.


wez:
Thank you!

PR #354, opened 2025-03-10
Complete the missing part of the command

blakethepatton:
I think it'd be helpful to also include over what protocol the rebind actually sends messages back into the queue. 

* Is the rebind talking through the load balancer over the same protocol that the messages was relayed to it ie: http or smtp
* Is there a special means in which it's communicating with the other server or does this command require additional authentication in order to communicate with the server on the other side of the load balancer?

This info might already be available and I just need to rtfm, in which case I'd want to link to that particular doc as well over here.


wez:
How rebinding works is discussed in https://docs.kumomta.com/reference/kcli/rebind/ 
It's really just updating metadata on a message which then influences which named queue it will be inserted into; https://docs.kumomta.com/reference/queues/ discusses how queue naming works.

There is no external communication carried out by the rebind, but if you have implemented a [rebind event handler](https://docs.kumomta.com/reference/events/rebind_message/), you can choose to "do something" extra to influence your decision on what metadata to update, if any.

wez:
Thanks!

PR #357, opened 2025-03-18
Remove Docker compose top-level version property

ryanjbonnell:
A warning message is displayed that the [version element is obsolete](https://docs.docker.com/reference/compose-file/version-and-name/) and should be removed.


ryanjbonnell:
I noticed this warning while debugging the issue in #358.

wez:
Thanks!

PR #361, opened 2025-03-19
Hickory025

wez:
There are a few breaking changes in the API that we need to tackle
here, but they're generally fine.

Since our libunbound crate uses hickory-proto, I had to upgrade
that dep over there and reference the updated git rev here in
this commit; we don't publish that to crates.io.

The main thing that stood out in the upgrade is that the zone
file parser used by the TestResolver now seems to create
records with varying FQDN-ness. This may just be that it is
now passing data through from the underlying zone data, but
it caused a number of the SPF and DKIM test cases to fail
without canonicalizing the names to FQDN.  I chose to do that
in the TestResolver rather than reviewing all the input zone
data, with the rationale being that it is least surprising
to have the test resolver fix that up than to puzzle over
records not resolving due to a missing trailing dot when
more tests are written in the future.

This commit doesn't try to take advantage of the improvements
to DNSSEC that are available in this version of hickory,
it's just upgrading to the API changes.

### DIFF
--- a/crates/dns-resolver/src/resolver.rs
+++ b/crates/dns-resolver/src/resolver.rs
@@ -1,12 +1,12 @@
 use async_trait::async_trait;
-use hickory_resolver::error::{ResolveError, ResolveErrorKind};
+use hickory_proto::ProtoErrorKind;
 use hickory_resolver::proto::op::response_code::ResponseCode;
 use hickory_resolver::proto::rr::rdata::{A, AAAA, MX, PTR, TXT};
 #[cfg(feature = "unbound")]
 use hickory_resolver::proto::rr::DNSClass;
 use hickory_resolver::proto::rr::{LowerName, RData, RecordData, RecordSet, RecordType, RrKey};
 use hickory_resolver::proto::serialize::txt::Parser;
-use hickory_resolver::{Name, TokioAsyncResolver};
+use hickory_resolver::{Name, ResolveError, ResolveErrorKind, TokioResolver};
 #[cfg(feature = "unbound")]
 use libunbound::{AsyncContext, Context};
 use std::collections::BTreeMap;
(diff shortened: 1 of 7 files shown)


wez:
@djc just an FYI; I don't think there's any shocking revelation here about this hickory 0.25 upgrade. The zone parser thing mentioned above might be interesting to consider from an API usability perspective, but I don't consider it to be an issue.

I'm punting considering dane w/hickory until later!

crates/dns-resolver/src/resolver.rs
```diff
@@ -181,14 +197,12 @@ impl TestResolver {
...
-                let mut fqdn = full.clone();
-                fqdn.set_fqdn(true);
-                fqdn
-            }
+        let full_fqdn = {
+            let mut name = full.clone();
+            name.set_fqdn(true);
+            name
         };
+        let mut authority = full_fqdn.clone();
```
djc, line 205: Nit: it doesn't look like `authority` needs to clone `full_fqdn`? I'd also drop the nested scope for the `name` stuff if I were you, doesn't seem worth the syntactic overhead to me.

crates/dns-resolver/src/resolver.rs
```diff
@@ -517,29 +531,32 @@ impl Resolver for HickoryResolver {
...
-                    records: vec![],
-                    nxdomain: *response_code == ResponseCode::NXDomain,
-                    secure: false,
-                    bogus: false,
-                    why_bogus: None,
-                    response_code: *response_code,
-                    expires: Instant::now()
-                        + Duration::from_secs(negative_ttl.unwrap_or(60) as u64),
-                }),
+                ResolveErrorKind::Proto(proto_err) => match proto_err.kind() {
```
djc, line 534: Might be able to profitably use `err.proto()` here?

crates/mod-dns-resolver/src/lib.rs
```diff
@@ -145,8 +147,10 @@ pub fn register(lua: &Lua) -> anyhow::Result<()> {
                 });
             }
 
-            let resolver = TokioAsyncResolver::tokio(r_config, config.options);
-            dns_resolver::reconfigure_resolver(HickoryResolver::from(resolver));
+            let mut builder =
+                TokioResolver::builder_with_config(r_config, TokioConnectionProvider::default());
+            *builder.options_mut() = config.options;
```
djc, line 152: This API is arguably a little idiosyncratic, I think we'll improve on that (see https://github.com/hickory-dns/hickory-dns/pull/2877).

djc (approved):
LGTM!

I'm going to look into whether we need to adjust the parser FQDN-ness aspect of things, that looks potentially iffy.

crates/dns-resolver/src/resolver.rs
wez: ah, yeah, it reduces a bit of boilerplate, thanks!

crates/dns-resolver/src/resolver.rs
wez: authority needs to be fqdn in order to resolve, and we need to clone because we need to preserve an fqdn version of the original `full` parameter in order to resolve correctly below.

djc:
> I'm going to look into whether we need to adjust the parser FQDN-ness aspect of things, that looks potentially iffy.

Looks like we're going to fix this upstream.

PR #363, opened 2025-03-20
Add healthcheck example to single node Docker example

ryanjbonnell:


wez:
Looks good, thank you!

PR #386, opened 2025-06-20
Remove duplicate gmail automation rule

jack-atlas:
"gmail.com".automation --> regex = "Our system has detected an unusual rate"   

This item is repeated, although it will be override .

The reason for keeping the second one is to better protect ip reputation.


wez:
Thank you!

PR #390, opened 2025-06-30
add lookup_ptr

kayozaki:
Adding a function to use through lua to query PTR records.
This can be used to match against RBLs or create blocking policies against reverse domain.
Tested against both IPv4 and IPv6

### DIFF
--- a/crates/integration-tests/dns-ptr.lua
+++ b/crates/integration-tests/dns-ptr.lua
@@ -12,9 +12,7 @@ end
 -- check if we're able to resolve back to localhost
 local ok, a = pcall(kumo.dns.lookup_ptr, '127.0.0.1')
 assert(ok, 'expected localhost for 127.0.0.1 ptr')
-if ok then
-  assert(contains(a, 'localhost.'), 'expected localhost.')
-end
+assert(contains(a, 'localhost.'), 'expected localhost.')
 
 -- see if we're able to do resolve for ipv6
 local ok, a = pcall(kumo.dns.lookup_ptr, '::1')
--- a/crates/integration-tests/dns-ptr.lua
+++ b/crates/integration-tests/dns-ptr.lua
@@ -19,9 +19,7 @@ end
 -- see if we're able to do resolve for ipv6
 local ok, a = pcall(kumo.dns.lookup_ptr, '::1')
 assert(ok, 'expected localhost for ::1 ptr')
-if ok then
-  assert(contains(a, 'localhost.'), 'expected localhost.')
-end
+assert(contains(a, 'localhost.'), 'expected localhost.')
 
 local ok, a = pcall(kumo.dns.lookup_ptr, '::2')
 assert(not ok, '::2 should not have ptr')
--- a/crates/integration-tests/dns-ptr.lua
+++ b/crates/integration-tests/dns-ptr.lua
@@ -1,0 +1,28 @@
+local kumo = require 'kumo'
+
+local function contains(list, t)
+    for key, value in pairs(list) do
+        if value == t then
+            return true
+        end
+    end
+    return false
+end
+
+-- check if we're able to resolve back to localhost
+local ok, a = pcall(kumo.dns.lookup_ptr, '127.0.0.1')
+assert(ok, 'expected localhost for 127.0.0.1 ptr')
+if ok then
+  assert(contains(a, 'localhost.'), 'expected localhost.')
+end
+
+-- see if we're able to do resolve for ipv6
+local ok, a = pcall(kumo.dns.lookup_ptr, '::1')
+assert(ok, 'expected localhost for ::1 ptr')
+if ok then
+  assert(contains(a, 'localhost.'), 'expected localhost.')
+end
+
+local ok, a = pcall(kumo.dns.lookup_ptr, '::2')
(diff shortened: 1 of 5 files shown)


crates/integration-tests/dns-ptr.lua
wez, line 1: Let's move this to `crates/mod-dns-resolver/test-dns-ptr.lua`, then get the test infra to run it:

```patch
diff --git a/assets/run-lua-test b/assets/run-lua-test
index 36b04f67..765fe749 100755
--- a/assets/run-lua-test
+++ b/assets/run-lua-test
@@ -36,6 +36,7 @@ test_module 'policy-extras.sources'
 test_module 'policy-extras.typing'

 dofile "crates/mod-memoize/test.lua"
+dofile "crates/mod-dns-resolver/test-dns-ptr.lua"

 end)
 EOT
```

crates/integration-tests/dns-ptr.lua
```diff
@@ -0,0 +1,28 @@
...
+    end
+    return false
+end
+
+-- check if we're able to resolve back to localhost
+local ok, a = pcall(kumo.dns.lookup_ptr, '127.0.0.1')
+assert(ok, 'expected localhost for 127.0.0.1 ptr')
+if ok then
+  assert(contains(a, 'localhost.'), 'expected localhost.')
+end
```
wez: The `assert(ok)` above means that we don't really need the `if` block around this:

```suggestion
assert(contains(a, 'localhost.'), 'expected localhost.')
```

crates/integration-tests/dns-ptr.lua
```diff
@@ -0,0 +1,28 @@
...
+
+-- see if we're able to do resolve for ipv6
+local ok, a = pcall(kumo.dns.lookup_ptr, '::1')
+assert(ok, 'expected localhost for ::1 ptr')
+if ok then
+  assert(contains(a, 'localhost.'), 'expected localhost.')
+end
```
wez:
```suggestion
assert(contains(a, 'localhost.'), 'expected localhost.')
```

crates/integration-tests/dns-ptr.lua
```diff
@@ -0,0 +1,28 @@
...
+
+local function contains(list, t)
+    for key, value in pairs(list) do
+        if value == t then
+            return true
+        end
+    end
+    return false
+end
+
```
wez: This test is technically dependent upon the local DNS service in terms of its test expectations, and also also the availability of functional network and DNS service.

To remove that dependency, we can put the resolver into a test mode where the zones are explicitly provided, as shown below.

A consequence of this is that the test case for `::2` below will no longer raise an error, but will return an empty table.

```suggestion
local zones = {
  [[
$ORIGIN 0.0.127.in-addr.arpa.
1 30 IN PTR localhost.
  ]],
  [[
$ORIGIN 1.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.ip6.arpa.
@ 30 IN PTR localhost.
  ]],
}
kumo.dns.configure_test_resolver(zones)
```

crates/integration-tests/dns-ptr.lua
```diff
@@ -0,0 +1,28 @@
...
+-- see if we're able to do resolve for ipv6
+local ok, a = pcall(kumo.dns.lookup_ptr, '::1')
+assert(ok, 'expected localhost for ::1 ptr')
+if ok then
+  assert(contains(a, 'localhost.'), 'expected localhost.')
+end
+
+local ok, a = pcall(kumo.dns.lookup_ptr, '::2')
+assert(not ok, '::2 should not have ptr')
```
wez: With the test resolver configuration suggested above, this PTR lookup returns an empty list of records rather than an error, just because the test resolver prefers to return empty results rather than an error in that case.

I don't think it is strictly necessary to have this test case here, but it is something that you might want to consider when you build your larger integration as it depends on the zone being queried and the DNS server configuration whether you will experience an error or an empty result set.

docs/reference/kumo.dns/lookup_ptr.md
```diff
@@ -0,0 +1,20 @@
+# lookup_ptr
+
+```lua
+kumo.dns.lookup_ptr(IP)
+```
+
+Resolve PTR records for the requested `IP`.
+
+Raises an error if the domain doesn't exist.
```
wez:
```suggestion
Raises an error if there was an issue resolving the record.
```

docs/reference/kumo.dns/lookup_ptr.md
```diff
@@ -0,0 +1,20 @@
...
+
+Returns a lua array-style table with the list of A records returned from DNS.
```
wez:
```suggestion
Returns a lua array-style table with the list of PTR records returned from DNS.  The table may be empty.
```

wez (commented):
Thanks for this; I have some suggestions around the test coverage and implementation!

kayozaki:
Thank you for the review, made proposed changes

wez:
FYI: I rebased your PR to pick up some changes that I think will fix up the CI when running on non-kumocorp-member pull requests

PR #391, opened 2025-06-30
Adding support for presenting Client cert as part of make_egress_path

kayozaki:
Adding an option for presenting client side certificate during delivery to next hop MTA.
Parameter has been added to support both openssl and rustls.
If supplied cert/key is invalid or options are missing, it'll default to not presenting any client side cert.
The input parameter supports KeySource struct, so using Vault and Data should also work.


crates/kumod/src/smtp_dispatcher.rs
```diff
@@ -33,6 +34,10 @@ lruttl::declare_cache! {
 static BROKEN_TLS_BY_SITE: LruCacheWithTtl<String, ()>::new("smtp_dispatcher_broken_tls", 64 * 1024);
 }
 
+lruttl::declare_cache! {
+static CLIENT_CERT: LruCacheWithTtl<KeySource, Option<Vec<u8>>>::new("client_certificate", 1024);
```
wez: A couple of things to consider here:

1. Let's make the cache name more specific to this module in case we need to do something similar in another module in the future.
2. Let's wrap up the data in an `Arc` so that we can avoid making additional extraneous copies of the data.  I'm suggesting `Arc<Box<[u8]>>` (also known as a boxed slice) for this, as we don't need to mutate the `Vec` after we've produced it.  You can convert `Vec<u8>` into that form by using `Arc::new(vec.into_boxed_slice())`.  We use boxed slices in the message crate if you want some examples of its usage.
3. Negative caching. If there's an error resolving the KeySource, it's likely to persist until the underlying problem (likely a configuration error) has been resolved.  It can be expensive to keep repeating the load and parse, especially in a heavily loaded instance.  So let's store `Result<DATA, String>` (eg: a `Result` where the Error value is a `String`) so that we can cheaply propagate the error value.  You can see an example of how we do negative caching in the `dns-resolver` crate around the `MX_CACHE`.

```suggestion
static CLIENT_CERT: LruCacheWithTtl<KeySource, Result<Option<Arc<Box<[u8]>>, String>>::new("smtp_dispatcher_client_certificate", 1024);
```

crates/kumod/src/smtp_dispatcher.rs
```diff
@@ -438,6 +443,41 @@ impl SmtpDispatcher {
         let mut dane_tlsa = vec![];
         let mut mta_sts_eligible = true;
 
+        let mut certificate_from_pem: Option<Vec<u8>> = None;
+        let mut private_key_from_pem: Option<Vec<u8>> = None;
+
+        if let Some(pem) = &path_config.tls_certificate {
```
wez, line 488: There's a couple of things on my mind around this:

1. There's some repitition between the two code blocks in here, and in particular, reuse of a hard coded TTL duration. I think there should be a little helper function:

```rust
fn resolve_cached_client_cert(source: &KeySource) -> anyhow::Result<Option<...>> {
```

that function can internalize mapping and propagating the errors. Per the comment elsewhere about negative caching, the cache should store the error variant as a `String` (rather than `anyhow::Error`, which is awkward to store in a global due to a missing `Sync` trait), so the function is responsible for mapping both key source and pem loading errors into a string to store in the cache, and mapping a cached string back into an anyhow error for propagation, which leads into the next point...

2. Error propagation: I think that an error to load/parse the cert should prevent sending. Right now it just prints an error to the diagnostic log and continues. Raising an error using the `?` operator will prevent the send and log some context, which feels like a more desirable outcome.

3. is a bit more vague, but: it would be great if `kumod --validate` could surface errors (such as an incorrect path, or similar) with the cert/key.  Right now there is no infrastructure for this, so it would be unfair to ask you to do something about it in this PR.  I think that being well-factored is sufficient here, so that it is easier to later introduce this. TBH, I think that means that the cache and the function to resolve the cert/key and should live in `egress_path.rs` and be exposed as accessor methods, but the containing `kumo-api-types` crate is supposed to be a relatively light/neutral type-only crate, so that isn't ideal.  So let's just make things nice here in this file for the time being.

crates/kumod/src/smtp_dispatcher.rs
```diff
@@ -438,6 +443,41 @@ impl SmtpDispatcher {
         let mut dane_tlsa = vec![];
         let mut mta_sts_eligible = true;
 
+        let mut certificate_from_pem: Option<Vec<u8>> = None;
+        let mut private_key_from_pem: Option<Vec<u8>> = None;
```
wez: stylistically, I tend to omit the types from this sort of local declaration unless the compiler insists that they are needed, as it makes the code less dense and is a bit easier to read:

```suggestion
        let mut certificate_from_pem = None;
        let mut private_key_from_pem = None;
```

crates/rfc5321/src/client.rs
```diff
@@ -1031,6 +1032,23 @@ impl TlsOptions {
...
+            let key = PKey::private_key_from_pem(&key_data)?;
+            builder.set_private_key(key.as_ref())?;
+        }
+
+        match builder.check_private_key() {
+            Ok(_) => {}
+            Err(err) => {
+                tracing::error!("client side certificate failed to validate: {err:#}");
+            }
+        };
```
wez: I think this should propagate the error as well:

```suggestion
        builder.check_private_key()?;
```

crates/rfc5321/src/tls.rs
```diff
@@ -45,6 +50,18 @@ impl std::hash::Hash for RustlsCacheKey {
...
+                pem.hash(hasher);
+            }
+            _ => {}
+        }
+        match &self.private_key_from_pem {
+            Some(pem) => {
+                pem.hash(hasher);
+            }
+            _ => {}
+        }
```
wez: stylistic suggestion:

```suggestion
        if let Some(pem) = &self.certificate_from_pem {
            pem.hash(hasher);
        }
        if let Some(pem) = &self.private_key_from_pem {
           pem.hash(hasher);
        }
```

crates/rfc5321/src/tls.rs
```diff
@@ -117,18 +137,110 @@ impl TlsOptions {
...
+            None => {}
+            Some(pem) => match self.load_private_key(&pem).await {
+                Ok(key) => {
+                    rustls_private_key = Some(Arc::new(key));
+                }
+                Err(err) => {
+                    tracing::error!("failed to load private key: {err:#}");
+                }
+            },
+        }
```
wez: I think it is simpler when phrased this way, which also propagates the error:

```suggestion
        if let Some(pem) = &self.private_key_from_pem {
            rustls_private_key = Some(Arc::new(self.load_private_key(&pem).await?));
        }
```

please also adopt this for the certificate logic below!

crates/rfc5321/src/tls.rs
```diff
@@ -117,18 +137,110 @@ impl TlsOptions {
...
+            .dangerous()
+            .with_custom_certificate_verifier(verifier.clone());
+        let config = match (&rustls_certificate, &rustls_private_key) {
+            (Some(certs), Some(key)) => match builder
+                .clone()
+                .with_client_auth_cert(certs.as_ref().clone(), key.as_ref().clone_key())
+            {
+                Ok(cfg) => Arc::new(cfg),
+                Err(err) => {
+                    tracing::error!("invalid client side certificate: {err:#}");
```
wez: we should propagate the error here too

crates/integration-tests/test-client-cert.lua
```diff
@@ -0,0 +1,133 @@
+local kumo = require 'kumo'
```
wez: from a test perspective, we should use the existing framework for this, otherwise the tests will never be run by anyone other than you!

Take a look at `crates/integration-tests/src/test/tls_opportunistic_fail.rs` for an example of an integration test that does something (small) with TLS options.

I think the approach to be taken here is to add `crates/integration-tests/src/test/tls_client_cert.rs` and define a test function for each of the cases you are testing here.  You'll need to add an appropriate `mod` line to `crates/integration-tests/src/test/mod.rs` to reference that in the build.

That will allow using `cargo nextest run` to launch a specific case during investigation, as well as to safely spawn multiple integration tests in parallel without conflicting on the listener ports.  `make test` is how I generally run all of the tests.

For certificate generation, rather than using the openssl cli, it is preferable to use a rust crate as we can control all of the certificate parameters more directly and without any surprising results from how the system is configured.

We use the `rcgen` crate for our existing automatic self-signed-cert generation inside kumo.  What I think you might want to do here is to use `rcgen` in the integration test module that you add (eg: `crates/integration-tests/src/test/tls_client_cert.rs`) to generate the different variations of cert/key and put them into the isolated test directory used for the spawned kumo daemon.

You might need to do a bit of plumbing in `crates/integration-tests/src/kumod.rs` to facilitate that.

For the policy file that is run by that kumod, there are two possible approaches you could take:

1. Introduce some environment variables that influence `crates/integration-tests/source.lua` to adopt the appropriate certs/keys in the appropriate contexts
2. Define a separate policy file and spawn the daemon using that instead; you can see that we do something like that for `tsa_source.lua` in some of the tests.

My suggestion as to which to use is whichever is simplest to follow in the end!

wez (changes requested):
Thank you for this!
I have some suggestions, and the biggest thing to tackle is around making the tests part of the integration test suite.
I've left some pointers; please let me know if you have questions!

crates/rfc5321/src/tls.rs
kayozaki: Did quite a bit of restructuring. Hope it looks ok

kayozaki:
Latest change
----
     Summary [  69.759s] 342 tests run: 342 passed, 0 skipped

crates/integration-tests/source.lua
```diff
@@ -301,8 +301,20 @@ kumo.on('get_egress_path_config', function(domain, source_name, _site_name)
     try_next_host_on_transport_error = (
       (os.getenv 'KUMOD_TRY_NEXT_HOST_ON_TRANSPORT_ERROR') and true
     ) or false,
+    tls_prefer_openssl = ((os.getenv 'PREFER_OPENSSL') and true) or false,
```
wez: nit: prefix these env vars with `KUMOD_` for the sake of consistency with the various other env vars being passed down

crates/integration-tests/src/test/tls_client_certificate.rs
```diff
@@ -0,0 +1,164 @@
...
+use std::env;
+use std::time::Duration;
+
+// These test do not confirm the client side certificates are being sent to the server.
+// They confirm the client side certificate paramters are being honored and loaded. Confirmed
+// they're valid before proceeding with delivery. Failure in validation would result in temp failure..
+// Add unit test for server side client side certificate handling once that feature is implemented.
+
+#[tokio::test]
+#[serial]
```
wez: why do we need this? Can we remove this requirement? It is very undesirable because of the impact it has on running tests

crates/integration-tests/src/test/tls_client_certificate.rs
```diff
@@ -0,0 +1,164 @@
...
+    env::remove_var("CLIENT_PRIVATE_KEY");
+    return r;
+}
+
+async fn tls_client_certificate(expected: DeliverySummary) -> anyhow::Result<()> {
+    let mut daemon = DaemonWithMaildir::start_with_env(vec![
+        // The default is OpportunisticInsecure; tighten it up a bit
+        // so that we fail to deliver to the sink, and verify
+        // the result of that attempt below
+        ("KUMOD_ENABLE_TLS", "OpportunisticInsecure"),
```
wez: ah, I think the `serial` stuff is because you're using `env:set_var`?  We really don't want to use that; it's super dangerous (is marked as `unsafe` in newer/upcoming versions of rust!) and not needed.  You can pass down the list of env vars here; I think you can simpy change the signature of `tls_client_certificate` to accept this vec of vars and pass it down from each of the tests alongside the expected delivery summary

crates/rfc5321/src/client.rs
```diff
@@ -712,98 +716,107 @@ impl SmtpClient {
...
-                    Ok(ip) => ServerName::IpAddress(ip.into()),
-                    Err(_) => ServerName::try_from(self.hostname.clone())
-                        .map_err(|_| ClientError::InvalidDnsName(self.hostname.clone()))?,
-                };
+                _ => Box::new(ssl_stream),
+            }
+        } else {
+            tls_info.provider_name = "rustls".to_string();
+            let connector = options.build_tls_connector().await.map_err(|error| {
+                ClientError::InvalidClientCertificate {
```
wez: is this particular error case guaranteed to be caused by the client certificate?

Edit: I see. I think that this should be a `ClientError::FailedToBuildConnector` or something that is a bit more generic, as it isn't necessarily due to an invalid client certificate, and it might be misleading, especially as/when we evolve `build_tls_connector` in the future, to claim that it is due to the cert.  The error string that we propagate out here should be sufficient context to explain what actually happened.

crates/rfc5321/src/tls.rs
```diff
@@ -45,6 +50,18 @@ impl std::hash::Hash for RustlsCacheKey {
...
+                pem.as_ref().clone().into_vec().hash(hasher);
+            }
+            _ => {}
+        }
+        match &self.private_key_from_pem {
+            Some(pem) => {
+                pem.as_ref().clone().into_vec().hash(hasher);
+            }
+            _ => {}
+        }
```
wez:
```suggestion
        if let Some(pem) = &self.certificate_from_pem {
                pem.as_ref().clone().into_vec().hash(hasher);
        }
        if let Some(pem) = &self.private_key_from_pem {
                pem.as_ref().clone().into_vec().hash(hasher);
        }
```

crates/rfc5321/src/tls.rs
```diff
@@ -117,17 +137,93 @@ impl TlsOptions {
...
+                let rsa_keys: Vec<PrivateKeyDer<'static>> = {
+                    let mut reader = BufReader::new(data.as_slice());
+                    rustls_pemfile::rsa_private_keys(&mut reader)
+                        .into_iter()
+                        .map(|r| r.map(PrivateKeyDer::Pkcs1))
+                        .collect::<Result<Vec<PrivateKeyDer<'static>>, std::io::Error>>()?
+                };
+
+                if !rsa_keys.is_empty() {
+                    return Ok(Some(Arc::new(rsa_keys.into_iter().next().unwrap())));
```
wez: can we eliminate the `unwrap` from here?  We should be able to translate it into an Err response and avoid raising a panic, which is difficult to catch and respond to.

Same comment for other uses of `unwrap`.

crates/tls-probe/src/main.rs
```diff
@@ -36,6 +39,10 @@ struct ProbeCommand {
     #[arg(long, value_parser=clap::builder::ValueParser::new(find_suite))]
     rustls_cipher_suites: Vec<SupportedCipherSuite>,
     #[arg(long)]
+    certificate: Option<String>,
```
wez, line 43: the option name is ambiguous and suggests that it might be the certificate itself, rather than being a path to it; let's spell it out in the `--help`  output by adding a doc comment.

```suggestion
    /// path to a certificate file
    certificate: Option<String>,
```

crates/tls-probe/src/main.rs
```diff
@@ -36,6 +39,10 @@ struct ProbeCommand {
...
+    #[arg(long)]
+    private_key: Option<String>,
```
wez, line 46:
```suggestion
    /// path to a private key file
    private_key: Option<String>,
```

docs/reference/kumo/make_egress_path/tls_certificate.md
```diff
@@ -0,0 +1,49 @@
+# tls_certificate
+
```
wez, line 2: Let's add the version info in here too:

```suggestion

{{since('dev')}}

```

docs/reference/kumo/make_egress_path/tls_private_key.md
```diff
@@ -0,0 +1,47 @@
+# tls_private_key
+
```
wez, line 2:
```suggestion

{{since('dev')}}

```

crates/integration-tests/src/test/tls_client_certificate.rs
kayozaki: Yes this is cause when i run the test the tests are running parallel and env variable defined in one test are evading into another test. it's mostly due to how crates/integration-tests/source.lua is using env variable for decision making.

Ah, ok understood, let me change and remove serial

crates/rfc5321/src/tls.rs
kayozaki: Since i'm checking is_empty before, i take unwrap would not return an error. 
Would 
```
return Ok(pkcs8_keys.into_iter().next().map(Arc::new));
```
be ok ? In each step, if key is None, subsequent key type is checked.
At the end if all keys types are exhausted, we'll return an error

kayozaki:
Added recommended changes.
Rebased from main.
Test looks good

Summary [  69.841s] 347 tests run: 347 passed, 0 skipped

wez:
Thank you!

PR #392, opened 2025-07-01
WIP: Aes block feature

dariomaiocchi:
different as the previous one this is for aes-cbc enablement

### DIFF
--- a/crates/mod-crypto/Cargo.toml
+++ b/crates/mod-crypto/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "mod-crypto"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+anyhow = {workspace=true}
+config = {path="../config"}
+data-encoding = {workspace=true}
+mlua = {workspace=true, features=["serialize"]}
+ring = {workspace=true}
+serde_json = {workspace=true}
--- a/crates/mod-crypto/Cargo.toml
+++ b/crates/mod-crypto/Cargo.toml
@@ -8,5 +8,9 @@ anyhow = {workspace=true}
 config = {path="../config"}
 data-encoding = {workspace=true}
 mlua = {workspace=true, features=["serialize"]}
+data-loader = {path="../data-loader", default-features=false}
 ring = {workspace=true}
 serde_json = {workspace=true}
+aes-gcm = "0.10.3" 
+rand = { version = "0.8", features = ["std"] } 
+base64 = "0.22.1"
--- a/crates/mod-crypto/Cargo.toml
+++ b/crates/mod-crypto/Cargo.toml
@@ -11,6 +11,7 @@ mlua = {workspace=true, features=["serialize"]}
 data-loader = {path="../data-loader", default-features=false}
 ring = {workspace=true}
 serde_json = {workspace=true}
-aes-gcm = "0.10.3" 
-rand = { version = "0.8", features = ["std"] } 
-base64 = "0.22.1"
+aes = "0.8.4"
+block-modes = "0.9.1"
+block-padding = "0.3.3"
+cbc = "0.1.2"
--- a/crates/mod-crypto/src/lib.rs
+++ b/crates/mod-crypto/src/lib.rs
@@ -1,84 +1,125 @@
 use config::get_or_create_sub_module;
+use mlua::{Error as LuaError}; 
 use mlua::Lua;
-use std::fmt;
+use data_loader::KeySource;
+
+
+use aes_gcm::{
+    aead::{Aead, AeadCore, KeyInit, OsRng},
+    Aes256Gcm, Nonce, Key,
+};
+use std::fs::File;
+use std::io::{Read};
+// Import the base64 crate Engine trait anonymously so we can
+// call its methods without adding to the namespace.
+use base64::{engine::general_purpose::STANDARD, Engine as _};
+
 
 // TODO 
 // use https://github.com/KumoCorp/kumomta/blob/main/crates/data-loader/src/lib.rs#L13 KeySource
 // add unit-tests
-
-
-#[derive(Debug)]
-pub enum AesError {
-    InvalidKeyLength,
-    InternalError,
+pub struct AesParam {
+pub aes_key_source: Option<KeySource>,
 }
 
 
-impl std::error::Error for AesError {}
-impl fmt::Display for AesError {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        match self {
-            AesError::InternalError => write!(f, "Internal Erorr occurred"),
-            AesError::InvalidKeyLength => write!(f, "Invalid AES key length. Must be 16, 24, or 32 bytes (128, 192, or 256 bits)."),
-        }
+
+
+/// aes_encrypt
+/// 1) TODO: use the Keystore input instead of enc_key
+/// 2) TODO: depending on the key bits use the correct Chiper creation
+pub fn aes_encrypt(value: String, enc_key: String, key_length_bits: usize) -> Result< String, aes_gcm::Error> {
+    // Determine expected key size in bytes
+    let key_size_bytes = match key_length_bits {
+        128 => 16,
+        192 => 24,
+        256 => 32,
+        _ => panic!("Unsupported key size: {key_length_bits} bits"),
+    };
+
+    // Read the key from the file
+    let mut file = File::open(enc_key).map_err(|_| aes_gcm::Error)?;
+    let mut key_bytes = vec![0u8; key_size_bytes];
+    file.read_exact(&mut key_bytes).map_err(|_| aes_gcm::Error)?;
+
+    // 
+    let key = Key::<Aes256Gcm>::from_slice(&key_bytes);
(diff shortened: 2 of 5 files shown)


crates/mod-crypto/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "mod-crypto"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+anyhow = {workspace=true}
```
wez, line 7: stylistically, prefer `crate.workspace = true` when simply pulling in a reference to a workspace crate with no other option overrides

```suggestion
anyhow.workspace=true
```

crates/mod-crypto/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+config = {path="../config"}
+data-encoding = {workspace=true}
+mlua = {workspace=true, features=["serialize"]}
+data-loader = {path="../data-loader"}
+ring = {workspace=true}
```
wez, line 12: if `ring` is unused here, please remove this dep. FIPS stuff is coming up on our radar and early indications are that ring is potentially problematic for simply being in the build dependency graph in some scan reports.

I'm anticipating that our FIPS support effort will be to replace any direct and indirect `ring` deps with `aws-lc`:

https://docs.rs/aws-lc-rs/1.13.1/aws_lc_rs/#motivation

crates/mod-crypto/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+serde_json = {workspace=true}
+serde = {workspace=true}
+aes = "0.8.4"
```
wez, line 15: I'm concerned by this security warning in the `aes` crate:

https://docs.rs/aes/latest/aes/#-security-warning-hazmat

Can we replace the usage of the `aes` crate with `aws-lc-rs`?

https://docs.rs/aws-lc-rs/1.13.1/aws_lc_rs/cipher/index.html

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+    match params.key {
+        AesKey::Aes128(k) => {
+            let cipher = Aes128CbcEnc::new((&k).into(), (&iv).into());
+            match cipher.encrypt_padded_mut::<Pkcs7>(&mut buf, plaintext_bytes.len()) {
+                Ok(ct) => {
+                    let mut output = iv.to_vec(); // IV is 16 bytes
+                    output.extend_from_slice(ct);
+                    Ok(output)
+                }
+                Err(_) => Err("Encryption failed or padding error"),
```
wez, line 82: This appears to be ignoring the underlying error and replacing it with some hard coded error text.  In some cases the original error type is useless and this is warranted, but in most other cases this is the sort of thing that produces obscure and hard to debug support issues, especially in this situation where it could be two different things going wrong.

So, it is preferable to include the original context in the error that gets bubbled up!

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+    crypto.set(
+        "aes_decrypt_cbc",
+        lua.create_async_function(|lua, params: Value| async move {
+            let params: LuaCfgDecrypt = from_lua_value(&lua, params)?;
+            let aes_k = match params.key {
+                Some(key) => key
+                    .get()
+                    .await
+                    .map_err(|e| LuaError::external(format!("key.get() failed: {}", e)))?,
+                None => return Err(LuaError::external("No AES key was provided for decryption")),
```
wez, line 156: I think `params.key` should simply not be declared as `Option`al, so that the `from_lua_value` handles this check as part of its own parsing.  It makes things more declarative and straightforward overall.

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+            };
+
+            let aes_key =
+                AesKey::from_bytes(&aes_k).map_err(|e| LuaError::external(e.to_string()))?;
```
wez, line 160: add `use config::any_err;` to the imports at the top, and you can simplify this:

```suggestion
                AesKey::from_bytes(&aes_k).map_err(any_err)?;
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+                let mut arr = [0u8; 32];
+                arr.copy_from_slice(bytes);
+                Ok(AesKey::Aes256(arr))
+            }
+            _ => Err("Key length must be 16 or 32 bytes"),
+        }
+    }
+}
+
+fn aes_encrypt_cbc(plaintext: &str, params: AesParams) -> Result<Vec<u8>, &'static str> {
```
wez, line 65: `AesParams` contains just `AesKey`; can we simplify these functions to directly pass `AesKey` instead?

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+    )?;
+    crypto.set(
+        "aes_decrypt_cbc",
+        lua.create_async_function(|lua, params: Value| async move {
+            let params: LuaCfgDecrypt = from_lua_value(&lua, params)?;
+            let aes_k = match params.key {
+                Some(key) => key
+                    .get()
+                    .await
+                    .map_err(|e| LuaError::external(format!("key.get() failed: {}", e)))?,
```
wez, line 155: I feel like `key` should be included in the error context here to make things easier to debug if this error condition is ever met

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+        let correct_params = AesParams {
+            key: correct_key.clone(),
+        };
+        let wrong_params = AesParams { key: wrong_key };
+
+        let plaintext = "This message won't decrypt correctly with the wrong key";
+        let ct = aes_encrypt_cbc(plaintext, correct_params.clone()).expect("Encryption failed");
+
+        let result = aes_decrypt_cbc(&ct, wrong_params);
+        assert!(result.is_err());
```
wez, line 209: Can make this a bit more proactive by using `unwrap_err` to ensure that we get an error variant, then compare that with an expected error:

```suggestion
        assert_eq!(result.unwrap_err(), SOMETHING);
```

simple_policy.lua
```diff
@@ -384,3 +384,16 @@ kumo.on('smtp_server_auth_plain', function(authz, authc, password)
...
+  key = '/tmp/kumo/aes_key_256.bin',
+  value = 'helloworld',
+}
+print(f)
+local b = kumo.crypto.aes_decrypt_cbc {
+  key = '/tmp/kumo/aes_key_256.bin',
+  decrypted = f,
+}
+local original = string.char(table.unpack(b))
+print(original)
```
wez, line 399: nobody should really be using `simple_policy.lua` for anything :)

To test the lua bindings themselves, look at `test_memoize` in `crates/mod-memoize/src/lib.rs`.  You can use that approach as a test driven from rust in your `mod-crypto` crate

wez (commented):
I'm not sure about the lua interface itself; I think that perhaps it could be made a bit more generic and extensible to other cipher types, but I don't want to focus too much on that at the moment.  The larger concern is around the crypto code itself; I'd feel more comfortable adopting a higher level crate as the basis of the crypto code, and ideally via a dependency that has a FIPS option.

I think that the aws-lc-rs crate is probably the right choice for this, but I've only briefly glanced at it.

I'm hoping that that reduces the bulk of the code added here and makes it simpler overall to read and review!

dariomaiocchi:
@wez  thx for review i will adress your comments and ping back once i made progress. thank you for your time.

simple_policy.lua
dariomaiocchi: thx resolving as adressed in new pr

crates/mod-crypto/src/lib.rs
dariomaiocchi: oki thx i have updated thx for remark

crates/mod-crypto/src/lib.rs
dariomaiocchi: updated with new params

crates/mod-crypto/src/lib.rs
dariomaiocchi: refactored

crates/mod-crypto/src/lib.rs
dariomaiocchi: updated thx

crates/mod-crypto/Cargo.toml
dariomaiocchi: thx updated in new pr

crates/mod-crypto/Cargo.toml
dariomaiocchi: updated in new pr

crates/mod-crypto/src/lib.rs
dariomaiocchi: yop make sense i have added it in my next pr thx

crates/mod-crypto/src/lib.rs
dariomaiocchi: this is solved

PR #394, opened 2025-07-03
add support for compression body in http injection api

dschaaff:


crates/kumod/Cargo.toml
```diff
@@ -67,6 +67,7 @@ throttle = {path="../throttle"}
 timeq = {path="../timeq"}
 tokio = {workspace=true, features=["full", "tracing"]}
 tokio-rustls = {workspace=true}
+tower-http = {workspace=true, features=["decompression-full"]}
```
wez: Can probably simplify to this and avoid pulling in some unused deps?

```suggestion
tower-http.workspace=true
```

dschaaff:
@wez I moved the implementation to the common http server module as requested. It should be ready for review now.

crates/kumod/src/http_server/inject_v1.rs
```diff
@@ -1232,4 +1232,86 @@ Ok(
 "#
         );
     }
+
+    #[tokio::test]
```
wez: FWIW, these newly added tests in this file don't (shouldn't, AFAICT) be affected by any compression settings, as compression is a transport thing and these tests are directly creating the structs that the requests are parsed into.

I don't think we need these, as I don't think they are exercising anything different from the other tests?

wez (commented):
Thanks! Looks good!
One question about a couple of the tests, but this is looking just about ready to merge!

crates/kumod/src/http_server/inject_v1.rs
dschaaff: You are spot on. I've removed those. I think the integration tests are sufficient.

PR #395, opened 2025-07-07
Aes block - CBC and ECB mode -- feature with module "aws lc rs"

dariomaiocchi:
Aes block - CBC and ECB mode -- feature with module "aws lc rs"

### DIFF
--- a/.github/workflows/builder-images.yml
+++ b/.github/workflows/builder-images.yml
@@ -101,6 +101,6 @@ jobs:
           context: .
           file: Dockerfile.builder
           platforms: ${{ matrix.platform }}
-          tags: ghcr.io/kumocorp/builder-for${{ matrix.REPOARM }}-${{ matrix.IMAGE }}
+          tags: ghcr.io/kumocorp/builder-for${{ matrix.REPOARM }}-${{ matrix.IMAGE }}:${{ matrix.VERSION }}
           push: ${{ github.event_name != 'pull_request' }}
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.24.2"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
  "gimli",
 ]
@@ -125,12 +125,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "android-tzdata"
-version = "0.1.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e999941b234f3131b00bc13c22d06e8c5ff726d1b6318ac7eb276997bbb4fef0"
-
-[[package]]
 name = "android_system_properties"
 version = "0.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -147,9 +141,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.20"
+version = "0.6.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3ae563653d1938f79b1ab1b5e668c87c76a9930414574a6583a7b7e11a8e6192"
+checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
 dependencies = [
  "anstyle",
  "anstyle-parse",
@@ -162,9 +156,9 @@ dependencies = [
 
 [[package]]
 name = "anstyle"
-version = "1.0.11"
+version = "1.0.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "862ed96ca487e809f1c8e5a8447f6ee2cf102f846893800b20cebdf541fc6bbd"
+checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
 
 [[package]]
 name = "anstyle-parse"
@@ -196,16 +190,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "any_ascii"
-version = "0.3.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "90c6333e01ba7235575b6ab53e5af10f1c327927fd97c36462917e289557ea64"
-
-[[package]]
 name = "anyhow"
-version = "1.0.99"
+version = "1.0.100"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b0674a1ddeecb70197781e945de4b3b8ffb61fa939a5597bcf48503737663100"
+checksum = "a23eb6b1614318a8071c9b2521f36b424b2c83db5eb3a0fead4a6c0809af6e61"
 
 [[package]]
 name = "arc-swap"
@@ -233,9 +221,9 @@ dependencies = [
 
 [[package]]
 name = "async-compression"
-version = "0.4.30"
+version = "0.4.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "977eb15ea9efd848bb8a4a1a2500347ed7f0bf794edf0dc3ddcf439f43d36b23"
+checksum = "5a89bce6054c720275ac2432fbba080a66a2106a44a1b804553930ca6909f4e0"
 dependencies = [
  "compression-codecs",
  "compression-core",
@@ -266,7 +254,7 @@ checksum = "13f937e26114b93193065fd44f507aa2e9169ad0cdabbb996920b1fe1ddea7ba"
 dependencies = [
  "async-channel",
  "async-executor",
- "async-io 2.5.0",
+ "async-io 2.6.0",
  "async-lock 3.4.1",
  "blocking",
  "futures-lite 2.6.1",
@@ -305,20 +293,20 @@ dependencies = [
 
 [[package]]
 name = "async-io"
-version = "2.5.0"
+version = "2.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "19634d6336019ef220f09fd31168ce5c184b295cbf80345437cc36094ef223ca"
+checksum = "456b8a8feb6f42d237746d4b3e9a178494627745c3c56c6ea55d92ba50d026fc"
 dependencies = [
- "async-lock 3.4.1",
+ "autocfg",
  "cfg-if",
  "concurrent-queue",
  "futures-io",
  "futures-lite 2.6.1",
  "parking",
- "polling 3.10.0",
- "rustix 1.0.8",
+ "polling 3.11.0",
+ "rustix 1.1.2",
  "slab",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -371,7 +359,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c7c24de15d275a1ecfd47a380fb4d5ec9bfe0933f309ed5e705b775596a3574d"
 dependencies = [
  "proc-macro2 1.0.101",
- "quote 1.0.40",
+ "quote 1.0.41",
  "syn 2.0.106",
 ]
 
@@ -388,7 +376,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9035ad2d096bed7955a320ee7e2230574d28fd3c3a0f186cbea1ff3c7eed5dbb"
 dependencies = [
  "proc-macro2 1.0.101",
- "quote 1.0.40",
+ "quote 1.0.41",
  "syn 2.0.106",
 ]
 
@@ -415,9 +403,9 @@ checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
 name = "aws-lc-rs"
-version = "1.13.3"
+version = "1.14.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5c953fe1ba023e6b7730c0d4b031d06f267f23a46167dcbd40316644b10a17ba"
+checksum = "879b6c89592deb404ba4dc0ae6b58ffd1795c78991cbb5b8bc441c48a070440d"
 dependencies = [
  "aws-lc-sys",
  "untrusted 0.7.1",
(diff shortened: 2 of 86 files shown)


dariomaiocchi:
@wez  feel free to take a look once time thx!

crates/mod-crypto/test.lua
```diff
@@ -0,0 +1,39 @@
...
+  },
+  value = 'second message here!!!',
+  algorithm = 'ecb',
+}
+local b2 = kumo.crypto.aes_decrypt_block {
+  key = {
+    -- read binary key
+    key_data = hex_key,
+  },
+  decrypted = f2,
```
dariomaiocchi: i will rename this to `value` as this param name is confusing

dariomaiocchi:
i will add also table support as digest module

crates/mod-crypto/test.lua
dariomaiocchi: updated

dariomaiocchi:
updated . once we get a new cycle of review i will add the docs

crates/mod-crypto/test.lua
```diff
@@ -0,0 +1,41 @@
...
+local f = kumo.crypto.aes_encrypt_block {
+  key = '/tmp/kumo/aes_key_256.bin',
+  value = 'some value here world',
+  algorithm = 'cbc',
+}
+-- 'f' is now a raw binary string representing the encrypted data.
+-- 'b' will be a DecryptResult object.
+local b = kumo.crypto.aes_decrypt_block {
+  key = '/tmp/kumo/aes_key_256.bin',
+  value = f,
```
dariomaiocchi: i have some doubts here if the `f` input for decrypting if we should allow some sort of multiple formats

( hex, base64, binary etc)

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
+use anyhow::anyhow;
+use aws_lc_rs::cipher::{
+    AES_128, AES_256, DecryptionContext, PaddedBlockDecryptingKey, PaddedBlockEncryptingKey,
+    UnboundCipherKey,
+};
+
```
wez: Prefer to avoid blank lines between the imports so that they all are sorted and maintained by the code formatting.

```suggestion
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+    BASE32, BASE32HEX, BASE32HEX_NOPAD, BASE32_NOPAD, BASE64, BASE64URL, BASE64URL_NOPAD,
+    BASE64_NOPAD, HEXLOWER,
+};
+use config::{any_err, from_lua_value, get_or_create_sub_module};
+use mlua::prelude::LuaUserData;
+use mlua::{Lua, MetaMethod, UserDataFields, UserDataMethods, Value, Error as LuaError};
+use data_loader::KeySource;
+use hex::decode;
+use std::str;
+
```
wez:
```suggestion
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+#[derive(Deserialize, Clone, Debug)]
+pub struct AesParams {
+    pub algorithm: AesAlgo,
+    pub key: AesKey,
+}
+
+#[derive(Deserialize, Clone, Debug)]
+pub enum AesAlgo {
+    Ecb(),
+    Cbc(),
```
wez: These variants don't need the parens; let's remove them so that the code is easier to read by having less punctuation throughout!

```suggestion
    Ecb,
    Cbc,
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+                _ => Err(anyhow!("Unexpected IV context for encrypting in CBC mode")),
+            }
+        }
+    }
+}
+
+
+struct DecryptResult(Vec<u8>);
+impl LuaUserData for DecryptResult {
+    fn add_fields<F: UserDataFields<Self>>(fields: &mut F) {
```
wez: I realize that this is a precedent set by the digest module; I think it is convenient to do this sort of thing, but we should probably factor it out so that it can be re-used.

I'd suggest that we rename the digest module's `DigestResult` to `BinaryResult` and make it public, then re-use it in this module in place of `DecryptResult` so that we only have a single implementation that can be consistently used throughout this and similar future crates.

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+pub enum AesAlgo {
+    Ecb(),
+    Cbc(),
+}
+
+#[derive(Deserialize, Clone, Debug)]
+pub struct LuaCfg {
+    pub key: KeySource,
+    pub value: String,
+    pub algorithm: String,
```
wez: if we type `algorithm`, then the `Deserialize` logic will parse the enum name for us (especially together with the removal of the parens from the variants above).

It will be case sensitive, which is fine and is consistent with the other type handling throughout kumo.

```suggestion
    pub algorithm: AesAlgo,
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+}
+
+#[derive(Deserialize, Clone, Debug)]
+pub struct LuaCfgDecrypt {
+    pub key: KeySource,
+    pub value: Vec<u8>,
+    pub algorithm: String,
```
wez:
```suggestion
    pub algorithm: AesAlgo,
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+            let algo = match params.algorithm.to_lowercase().as_str() {
+                "cbc" => AesAlgo::Cbc(),
+                "ecb" => AesAlgo::Ecb(),
+                _ => {
+                    return Err(LuaError::external(format!(
+                        "Invalid algorithm '{}'. Expected 'cbc' or 'ecb'",
+                        params.algorithm
+                    )));
+                }
+            };
```
wez: This whole block should go away once `algorithm` is changed from `String` to `AesAlgo`.

```suggestion
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
```
wez: This block also can also be eliminated by typing `params.algorithm` as shown above

```suggestion
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+    hex_str.chars().all(|c| c.is_digit(16)) && hex_str.len() % 2 == 0
+}
+
+impl AesKey {
+    pub fn from_bytes(bytes: &[u8]) -> Result<Self, String> {
+        let decoded_bytes = if is_hex(bytes) {
+            decode(bytes).map_err(|e| format!("Failed to decode hex: {}", e))?
+        } else {
+            bytes.to_vec()
+        };
```
wez: this sort of thing results in ambiguity; what if I have data that looks like hex (eg: the word `deadbeef`) but which is not hex?  This logic will result in it being treated as hex and interpreted incorrectly.

Can we remove support for implicitly hex encoded keys? That will make things simpler to reason about here in the code.

If you have and need to support hex encoded keys, can we introduce a way to explicitly indicate that the data is hex encoded into the API/parameters?

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+        match decoded_bytes.len() {
+            16 => Ok(AesKey::Aes128(decoded_bytes.try_into().unwrap())),
+            32 => Ok(AesKey::Aes256(decoded_bytes.try_into().unwrap())),
```
wez: the `unwrap` calls here will cause a panic, which is undesirable when we can raise an error instead.  Can we simply use the `?` operator here to propagate the error?

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+            // context contain IV vector
+            let context = enc_key.encrypt(&mut buf_ciphertext)?;
+
+            match context {
+                DecryptionContext::Iv128(iv) => {
+                    let mut result = iv.as_ref().to_vec();
+                    result.extend_from_slice(&buf_ciphertext); // Append the actual ciphertext after the IV
+                    Ok(result)
+                }
+                _ => Err(anyhow!("Unexpected IV context for encrypting in CBC mode")),
```
wez:
```suggestion
                unsupported => anyhow::bail!("Unexpected IV context {unsupported:?} for encrypting in CBC mode"),
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+pub struct LuaCfg {
+    pub key: KeySource,
+    pub value: String,
+    pub algorithm: String,
+}
+
+#[derive(Deserialize, Clone, Debug)]
+pub struct LuaCfgDecrypt {
+    pub key: KeySource,
+    pub value: Vec<u8>,
```
wez: I *think* that having this type being declared as `Vec<u8>` is a little bit inflexible.

`mlua::String` supports representing both binary and UTF-8 string values, and I think that it would be a better choice to use for the encoded data in the interface to lua; that would allow passing strings that are binary (non-UTF-8).

I'm not sure that you can successfully use `mlua::String` with `Deserialize` here, so you might need to change the encode/decode function signatures in lua so that you explicitly pass the data to be encrypted/decrypted separately from the other configuration parameters.

Maybe something like this?

```lua
local f2 = kumo.crypto.aes_encrypt_block('Ecb', 'second message here!!',  {
     key_data = hex_key,
   })
```

There should be a test case that uses binary non-utf8 data as the input to validate that it is possible to use with these functions.

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+        };
+        match decoded_bytes.len() {
+            16 => Ok(AesKey::Aes128(decoded_bytes.try_into().unwrap())),
+            32 => Ok(AesKey::Aes256(decoded_bytes.try_into().unwrap())),
+            _ => Err("Key length must be 16 or 32 bytes".to_string()),
+        }
+    }
+}
+
+fn aes_encrypt_block(plaintext: &str, params: AesParams) -> Result<Vec<u8>, anyhow::Error> {
```
wez: similar to the comments above, having the plaintext defined as `str` (or `String`) here requires that the data to encode be UTF-8, which may be undesirable and inflexible.  This should probably be switched to a byte slice here and the string vs bytes handling at the interface from lua.

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+    let mut buf_ciphertext = plaintext.as_bytes().to_vec();
+    let enc_key: PaddedBlockEncryptingKey;
```
wez: the scoping of this is not very idiomatic, I'd suggest removing it from here

```suggestion
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    match params.algorithm {
+        AesAlgo::Ecb() => {
+            enc_key = match params.key {
```
wez:
```suggestion
            let enc_key = match params.key {
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+                    let key = UnboundCipherKey::new(&AES_256, &k)?;
+                    PaddedBlockEncryptingKey::ecb_pkcs7(key)?
+                }
+            };
+            // don't return IV vector for ecb
+            enc_key.encrypt(&mut buf_ciphertext)?;
+            Ok(buf_ciphertext)
+        }
+        AesAlgo::Cbc() => {
+            enc_key = match params.key {
```
wez:
```suggestion
            let enc_key = match params.key {
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+    fn add_methods<M: UserDataMethods<Self>>(methods: &mut M) {
+        methods.add_meta_method(MetaMethod::ToString, |_, this, _: ()| {
+            Ok(HEXLOWER.encode(&this.0))
+        });
+    }
+}
+
+fn aes_decrypt_block(ciphertext_buf: &[u8], params: AesParams) -> anyhow::Result<DecryptResult>  {
+    let mut in_out_buffer = ciphertext_buf.to_vec();
+    let dec_key: PaddedBlockDecryptingKey;
```
wez:
```suggestion
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+
+    match params.algorithm {
+        AesAlgo::Ecb() => {
+            dec_key = match params.key {
```
wez:
```suggestion
            let dec_key = match params.key {
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+
+            match dec_key.decrypt(&mut in_out_buffer, DecryptionContext::None) {
+                //     Ok(DigestResult(digest.as_ref().to_vec()))
+                Ok(plaintext) => Ok(DecryptResult(plaintext.to_vec())),
+                Err(_) => Err(anyhow!("Decryption failed with AES key, mode ecb block")),
+            }
+        }
+
+        AesAlgo::Cbc() => {
+            dec_key = match params.key {
```
wez:
```suggestion
            let dec_key = match params.key {
```

crates/mod-crypto/src/lib.rs
```diff
@@ -0,0 +1,375 @@
...
+                return Err(anyhow!(
+                    "Ciphertext is too short to contain an IV (expected at least 16 bytes for IV)"
+                ));
+            }
+
+            let (iv_bytes, actual_ciphertext) = ciphertext_buf.split_at(16); // Split into IV and actual ciphertext
+            let iv: [u8; 16] = iv_bytes
+                .try_into()
+                .map_err(|_| anyhow!("IV length is incorrect, expected 16 bytes"))?;
+
```
wez: Lots of "magic number" usage here with `16`. Is there a constant that we can reference from the underlying crypto crate instead?

If not, let's define something for ourselves:

```rust
const CBC_IV_LEN: usize = 16;
```

then we can use it throughout:

```suggestion
            anyhow::ensure!(ciphertext_buf.len() >= CBC_IV_LEN,
                    "Ciphertext is too short to contain an IV (expected at least {CBC_IV_LEN} bytes for IV)"
                );
                
            let (iv_bytes, actual_ciphertext) = ciphertext_buf.split_at(CBC_IV_LEN); // Split into IV and actual ciphertext
            let iv: [u8; CBC_IV_LEN] = iv_bytes
                .try_into()
                .map_err(|_| anyhow!("IV length is incorrect, expected {CBC_IV_LEN} bytes"))?;

```

crates/mod-crypto/test.lua
```diff
@@ -0,0 +1,41 @@
+local kumo = require 'kumo'
+
+-- 128 key
+local hex_key = '2b7e151628aed2a6abf7158809cf4f3c'
+
+-- showcase encryption/decryption with aes_cbc -- 01 file
+local f = kumo.crypto.aes_encrypt_block {
+  key = '/tmp/kumo/aes_key_256.bin',
```
wez: Please use `os.tmpname()` to produce a temporary file name if you need to use scratch paths in this sort of test, so that multiple tests from multiple repos can be run concurrently without fear of colliding.

crates/mod-crypto/test.lua
wez: IMO, simplicity is best here: just accepting binary/strings without encoding is easiest to reason about.  The caller can compose/decompose/decode data using other functions that either already exist or can be added as needed.

crates/mod-crypto/test.lua
```diff
@@ -0,0 +1,41 @@
...
+}
+local dr = kumo.crypto.aes_decrypt_block {
+  key = {
+    key_data = hex_key,
+  },
+  value = f2,
+  algorithm = 'ecb',
+}
+
+print(string.format("Hex: %s", dr.hex))
```
wez, line 85: this test doesn't really assert anything.  It should probably literally use `assert()` to validate that the output matches expectations!

dariomaiocchi:
thx wez will work through your review

crates/mod-crypto/src/lib.rs
dariomaiocchi: updatew without thx

crates/mod-crypto/src/lib.rs
dariomaiocchi: updated thx

crates/mod-crypto/test.lua
dariomaiocchi: ok updated

crates/mod-crypto/test.lua
dariomaiocchi: updated thx

crates/mod-crypto/test.lua
dariomaiocchi: oki sounds good

crates/mod-crypto/src/lib.rs
dariomaiocchi: yes , i think make sense.. will work on this

crates/mod-crypto/src/lib.rs
dariomaiocchi: updated this one thx

crates/mod-crypto/src/lib.rs
dariomaiocchi: done on last commit

dariomaiocchi:
@wez  i think the pr should be good to be reviewed. let me know for a 2nd round.. thx in advance :)

PR #401, opened 2025-07-26
Support for hostname-based proxy server configuration in KumoMTA's egress sources

pankajrathi95:
## Summary
This PR adds support for hostname-based proxy server configuration in KumoMTA's egress sources. Previously, `ha_proxy_server` and `socks5_proxy_server` only accepted SocketAddr format (e.g., "127.0.0.1:8080"). Now they support both SocketAddr and hostname formats, with hostnames being resolved to IP addresses at runtime when connections are established.

### crates/kumod/src/egress_source.rs:
- Added `SocketAddrOrHostname` enum to support both SocketAddr and hostname formats
- Updated `EgressSource` struct to use `SocketAddrOrHostname` for proxy server fields
- Enhanced `resolve_proxy_protocol` method to handle hostname resolution asynchronously
- Added comprehensive test coverage for the new functionality

## Documentation updates
- `docs/reference/kumo/make_egress_source/ha_proxy_server.md`: Updated to document hostname support with inline version indicator
- `docs/reference/kumo/make_egress_source/socks5_proxy_server.md`: Updated to document hostname support with inline version indicator
- `docs/userguide/operation/proxy.md`: Added examples showing both SocketAddr and hostname usage


crates/kumod/src/egress_source.rs
```diff
@@ -24,6 +24,68 @@ use std::time::{Duration, Instant};
...
+/// A type that can represent either a SocketAddr or a hostname string
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SocketAddrOrHostname {
+    SocketAddr(SocketAddr),
+    Hostname(String),
+}
+
+impl SocketAddrOrHostname {
+    /// Resolve to a SocketAddr, either directly or by resolving the hostname
+    pub async fn resolve(&self) -> anyhow::Result<SocketAddr> {
```
wez, line 36: This `resolve` method has a few issues:

1. It doesn't account for multiple candidate addresses being returned.  How should it decide what is appropriate to return if it is going to return only one?  If it returns multiple, what is a reasonable semantic for dealing with multiple? Does the caller try multiple in a loop? How does it decide whether it should try a candidate address after the first? Does it decide based solely on transport errors, or also on protocol errors? That introduces multiple attempts to establish the proxy connection; is that desirable? Is that something that should be configurable?
 
2. It doesn't use the resolver that was centrally configured in the `dns-resolver` crate, which might result in divergent/unexpected behavior if the system resolver is configured different from the kumo resolver.  Please use `dns_resolver::ip_lookup` to do the actual resolution; you'll need to convert its results into `SockAddr`'s with port numbers.

Note that you cannot add the `dns-resolver` crate as a dependency to `kumo-address` because the dependency graph already goes the other way.

I'd suggest adding a free function to `dns-resolver`, alongside the existing `resolve_a_or_aaaa` function, that works in a similar way but receives a `&DnsResolvableHostOrSocketAddr` instead of a `&str`.  Returning `Vec<ResolvedAddress>` would be good (rather than just `SocketAddr`), as that additional context (the original hostname) is useful to show in logs.

I'd leave any policy around selecting a specific address to the logic in `egress_source.rs`.

crates/kumod/src/egress_source.rs
```diff
@@ -24,6 +24,68 @@ use std::time::{Duration, Instant};
 use tokio::io::{AsyncReadExt, AsyncWriteExt};
 use tokio::net::{TcpSocket, TcpStream};
 
+/// A type that can represent either a SocketAddr or a hostname string
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SocketAddrOrHostname {
+    SocketAddr(SocketAddr),
+    Hostname(String),
```
wez, line 31: Since this allows any arbitrary string to be used as a fallback, this new type adds no real value to the configuration parser, because any typo in the string will be allowed by that `Hostname` variant.

I think it would be better to make the `Hostname` variant be typed as `hickory_proto::rr::domain::Name` so that it can validate that the input string is actually a compliant DNS name.

I'd prefer it this type was added to the `kumo-address` crate, instead of inlined here, with a name like `DnsResolvableHostOrSocketAddr`.

wez (commented):
Thanks for this!

Could you expand on your motivation for this change, as I think that will shape how to address the policy questions I've raised in this review?

For example, is the goal to just avoid using an IP address for the sake of readability, or is the goal to use some kind of DNS or IP based load balancing technique?

Please share more about the use case!

crates/kumod/src/egress_source.rs
pankajrathi95: I was about to close this PR and I did create a new PR with the same changes. I'll work on your suggestions mentioned.

The motivation behind the change is to call tha proxy server with hostname which is configured at our end in such a way. DNS resolver to A record would give us the IP address which we can pass it further.

crates/kumod/src/egress_source.rs
pankajrathi95: I did add validation to check if hostname is in the desired format. I pushed it to another #402 PR. I'll consolidate and make the changes and add it.
Thanks

PR #402, opened 2025-07-26
Support for hostname-based proxy server configuration in KumoMTA's egress sources

pankajrathi95:
## Summary
This PR adds support for hostname-based proxy server configuration in KumoMTA's egress sources. Previously, `ha_proxy_server` and `socks5_proxy_server` only accepted SocketAddr format (e.g., "127.0.0.1:8080"). Now they support both SocketAddr and hostname formats, with hostnames being resolved to IP addresses at runtime when connections are established.

### crates/kumod/src/egress_source.rs:
- Added `SocketAddrOrHostname` enum to support both SocketAddr and hostname formats
- Updated `EgressSource` struct to use `SocketAddrOrHostname` for proxy server fields
- Enhanced `resolve_proxy_protocol` method to handle hostname resolution asynchronously
- Added comprehensive test coverage for the new functionality

## Documentation updates
- `docs/reference/kumo/make_egress_source/ha_proxy_server.md`: Updated to document hostname support with inline version indicator
- `docs/reference/kumo/make_egress_source/socks5_proxy_server.md`: Updated to document hostname support with inline version indicator
- `docs/userguide/operation/proxy.md`: Added examples showing both SocketAddr and hostname usage


pankajrathi95:
@wez - Can you take a look at this PR.

wez:
I asked a number of higher-level motivating questions in https://github.com/KumoCorp/kumomta/pull/401, as well as about multiple ip addresses that haven't been responded to; the answers really shape what we need from the product and that will inform the direction that we need to take.

There's a lot of code in this PR that doesn't feel like it is in the right place, but we can tackle that once we have a clearer picture of things.

Let's start by tackling the motivating reasons: I understand that you want to be able to specify the hostname, but why? What is the purpose of using the name rather than the IP? Is it for load balancing purposes? Is it because the server might change IP? Is it because it might be dynamically instantiated?  Let's also tackle the questions from the other end: why do you NOT want to specify the IP address?  These answers really help to understand how best to proceed.

pankajrathi95:
@wez Sorry for replying late as I missed the comment on this one. Let me provide you the reason why I want to use the hostname. Yes the proxy server when it comes up, it's IP is dynamic and can change during a restart / deployment, So we have added a hostname so that we can always resolve to the right IP address. 

This is the reason why we can't use IP address directly. I did found a way to do the same thing in lua though, but I thought if this is provided out of the box it would be great and the PR is all about that.

PR #403, opened 2025-08-06
add epoch with millisec

kayozaki:
Adding two functions

- return epoch with milliseconds 
- return the time difference between now and given epoch millisec

primary usage is to be able to track and log the time duration of given task.

### DIFF
--- a/assets/run-lua-test
+++ b/assets/run-lua-test
@@ -39,6 +39,7 @@ dofile "crates/mod-memoize/test.lua"
 dofile "crates/mod-dns-resolver/test-dns-ptr.lua"
 dofile "crates/mod-mpsc/test.lua"
 dofile "crates/mod-filesystem/test.lua"
+dofile "crates/mod-time/test.lua"
 
 end)
 EOT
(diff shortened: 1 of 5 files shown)


wez:
Thanks for this; I've been holding off looking too hard at this because time types can have a lot of nuance; within Rust we have many types from which to choose:

* `SystemTime`
* `Instant` for monotonic time points (both `std` and `tokio` have a version of this)
* `Duration` for intervals
* chrono's `DateTime` (with timezone spec!) and `Duration`

I want to think a little bit about what makes the most sense in the appropriate contexts before proceeding with this PR or similar functionality. I will note that we're currently missing exposing the `Date` header within the lua bindings to the mime parser because we don't have a convenient lua wrapper around the chrono `DateTime<Utc>` type that it parses into, and we also have some tension in resolving the granularity of timestamps in the log record (see #405).  The tension is that the desire to expose higher granularity introduces a backwards compatibility concern unless we decide to bloat the the record with an additional higher resolution field.  Needs some more thought!

For tracking timing in particular, I'd like to point you to the existing https://docs.kumomta.com/reference/kumo.time/start_timer/ function, which can be used to label a span of execution and return the duration between the start/stop, while also populating a histogram that you can feed into eg: prometheus.  If that doesn't fit your use case, please let me know what the gap is!

What I'm leaning towards is standardizing on `DateTime<Utc>` for the date time object that we expose to lua (rather than `SystemTime`), then exposing the most useful `chrono` methods, as well as a reasonable `Duration` type.

PR #404, opened 2025-08-06
adding crc32 into mod-digest

kayozaki:
Adding crc32 onto digest function.
For my purpose, I wanted to be able to generate a short string representation of the execution environment.


wez:
Thanks!

PR #418, opened 2025-09-16
Initial DMARC support

sophiajt:
Begin laying the groundwork for DMARC support.

Please note:
- some of this carries over from SPF support so that we can align DMARC code-wise with the structure we had before. I suspect this will change as it matures
- the only implemented logic thus far is the SPF and DKIM alignment checks.
- lua registration is in this PR, but lua testing was moved out for simplicity.

### DIFF
--- a/crates/kumo-dmarc/src/lib.rs
+++ b/crates/kumo-dmarc/src/lib.rs
@@ -10,6 +10,7 @@ use crate::types::results::DmarcResultWithContext;
 use dns_resolver::Resolver;
 use hickory_resolver::proto::rr::RecordType;
 use hickory_resolver::Name;
+use std::collections::BTreeMap;
 use std::net::IpAddr;
 use std::str::FromStr;
 use std::time::SystemTime;
@@ -17,32 +18,36 @@ use std::time::SystemTime;
 pub use types::results::DmarcResult;
 
 pub struct CheckHostParams {
+    /// Domain of the sender in the "From:"
+    pub from_domain: String,
+
     /// Domain that provides the sought-after authorization information.
     ///
-    /// Initially, the domain portion of the "MAIL FROM" or "HELO" identity.
-    pub domain: String,
-
     /// The "MAIL FROM" email address if available.
-    pub sender: Option<String>,
+    pub mail_from_domain: Option<String>,
 
     /// IP address of the SMTP client that is emitting the mail (v4 or v6).
     pub client_ip: IpAddr,
+
+    /// The results of the DKIM part of the checks
+    pub dkim: Vec<BTreeMap<String, String>>,
 }
 
 impl CheckHostParams {
     pub async fn check(self, resolver: &dyn Resolver) -> DmarcResultWithContext {
         let Self {
-            domain,
-            sender,
+            from_domain,
+            mail_from_domain,
             client_ip,
+            dkim,
         } = self;
 
-        let sender = match sender {
-            Some(sender) => sender,
-            None => format!("postmaster@{domain}"),
-        };
-
-        match DmarcContext::new(&sender, &domain, client_ip) {
+        match DmarcContext::new(
+            &from_domain,
+            mail_from_domain.as_ref().map(|x| x.as_str()),
+            client_ip,
+            &dkim[..],
+        ) {
             Ok(cx) => cx.check(resolver).await,
             Err(result) => result,
         }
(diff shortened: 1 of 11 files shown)


crates/kumo-dmarc/src/lib.rs
```diff
@@ -1,3 +1,138 @@
 #![allow(dead_code)]
 
 mod types;
+
+#[cfg(test)]
+mod tests;
+
+use crate::types::record::Record;
```
wez, line 3: my preference is for the imports to be at the top of the file (or as close as possible in the face of the `#!` at the top of this particular file), because part of my workflow in my terminal is to grab suggested `use xxx;` statements from the cargo check/build output via wezterm's quickselect feature, then switch to vim, `gg` to jump to the top and then paste the import; it's a quick sequence of keys to get the suggestion applied.  I'll then use `make fmt` to put the import in the right place when I'm ready to tidy up and commit/amend.

crates/kumo-dmarc/src/lib.rs
```diff
@@ -1,3 +1,138 @@
...
+            Ok(name) => name,
+            Err(_) => {
+                return DmarcResultWithContext {
+                    result: DmarcResult::Fail,
+                    context: format!("invalid domain name: {}", self.from_domain),
+                }
+            }
+        };
+
+        let initial_txt = match resolver.resolve(name, RecordType::TXT).await {
```
wez: Hmm, it nets out the same, but this highlights a little wart in the trait's evolution; we have a `resolve_txt` method that internally performs the conversion to Name and the resolve call.  Implementors of the trait are not technically required to implement it by making the obvious call down to the `resolve` method, although all implementations do currently, and I have no plans to change that.

I'd suggest switching this to call `resolver.resolve_txt` because you can eliminate the explicit `Name` conversion above and reduce the line count here.

crates/kumo-dmarc/src/types/record.rs
```diff
@@ -1,15 +1,20 @@
+use dns_resolver::Resolver;
+
```
wez: for better or worse, I prefer to keep all the imports contiguous without blank lines separating crate vs. std vs. crates.io imports.  That way `make fmt` will keep things ordered the way I like.

```suggestion
```

crates/kumo-dmarc/src/types/record.rs
```diff
@@ -18,6 +23,73 @@ pub struct Record {
...
+        &self,
+        cx: &DmarcContext<'_>,
+        _resolver: &dyn Resolver,
+    ) -> DmarcResultWithContext {
+        match self.align_dkim {
+            Mode::Relaxed => {
+                for dkim in cx.dkim {
+                    if let Some(result) = dkim.get("header.d") {
+                        if !cx.from_domain.ends_with(&format!(".{}", result))
+                            && cx.from_domain != result
```
wez: I feel like these terms should be swapped so that the simple equality check could succeed and short-circuit before we do the format+allocate for the suffix check?

```suggestion
                        if cx.from_domain != result && !cx.from_domain.ends_with(&format!(".{}", result))
```

Since there is a similar check in the spf code below, perhaps this is worth pulling out into a tiny `is_domain_or_subdomain` helper function?

crates/kumo-dmarc/src/types/record.rs
```diff
@@ -18,6 +23,73 @@ pub struct Record {
...
+                    if let Some(result) = dkim.get("header.d") {
+                        if cx.from_domain != result {
+                            return DmarcResultWithContext {
+                                result: DmarcResult::Fail,
+                                context: "DMARC: DKIM strict check failed".into(),
+                            };
+                        }
+                    }
+                }
+            }
```
wez, line 67: should there be some logic here to handle the "dkim should be strict, but there was no dkim signature" case?  If not, perhaps a comment that says something like `It is OK for there to be no dkim signatures when in strict mode` would be good as a sign post of sorts here?  I can't see an obvious statement about missing dkim signature and its effect on the alignment in the RFC, so I'm not sure which way this should go!

Would be good to have a test for this case if you don't already (I'm currently still in my first pass through, so I don't yet know if that exists!)

crates/kumod/src/dmarc.rs
```diff
@@ -0,0 +1,106 @@
+use std::collections::BTreeMap;
+use std::net::SocketAddr;
+use std::str::FromStr;
+
```
wez:
```suggestion
```

crates/kumod/src/dmarc.rs
```diff
@@ -0,0 +1,106 @@
...
+pub fn register<'lua>(lua: &'lua Lua) -> anyhow::Result<()> {
+    let dmarc_mod = get_or_create_sub_module(lua, "dmarc")?;
+
+    dmarc_mod.set(
+        "verify",
+        lua.create_async_function(
+            |lua,
+             (msg, dkim_result, meta): (
+                UserDataRef<Message>,
+                UserDataRef<Vec<AuthenticationResult>>,
```
wez: I'm not sure if `UserDataRef<Vec<AuthenticationResult>>` will work at runtime because we're actually using the LuaSerdeExt support in mlua to serde-serialize the `AuthenticationResult` into a native lua table when we do the dkim verify logic.

So while this function definition will compile, I think mlua's dispatching logic will fail to call it at runtime when passing on the results from a dkim verify in lua.

I think you may need to switch the handling of this parameter to something like:

```rust
   dkim_results: mlua::Value // use generic lua value as type
```

and then deserialize it using the helper function from the `config` crate, which has slightly better error messaging than the raw `LuaSerdeExt` trait, but also can handle interop with the memoize crate and a couple of other weird situations:

```rust
   let dkim_results: Vec<AuthenticationResult> = config::from_lua_value(lua, dkim_results)?;
```

crates/kumod/src/dmarc.rs
```diff
@@ -0,0 +1,106 @@
...
+                    .get_meta("received_from")
+                    .and_then(|v| SocketAddr::from_str(v.as_str()?).ok())
+                    .expect("`received_from` is always set, and always to a value representing a `SocketAddr`");
+
+                let resolver = dns_resolver::get_resolver();
+
+                // MAIL FROM
+                let msg_sender = msg.sender();
+
+                let mail_from_domain = msg_sender.ok().map(|x| x.to_string());
```
wez: Probably want just the domain here?

```suggestion
                let mail_from_domain = msg_sender.ok().map(|x| x.domain().to_string());
```

crates/kumod/src/dmarc.rs
```diff
@@ -0,0 +1,106 @@
...
+                        from_domain.to_string()
+                    } else {
+                        return Ok(lua.to_value_with(
+                            &CheckHostOutput {
+                                disposition: DmarcResult::Fail,
+                                result: AuthenticationResult {
+                                    method: "dmarc".to_string(),
+                                    method_version: None,
+                                    result: "Only single 'From:' header supported".to_string(),
+                                    reason: Some("Only single 'From:' header supported".to_string()),
```
wez: I think these error strings are copypasta from the case below?

wez (commented):
Looking good; mostly just stylistic comments, but there are a couple of questions and a recommendation for receiving the auth results from the dkim verify code!

crates/kumo-dmarc/src/types/record.rs
sophiajt: Good catch. This would be important logic to make sure we're working with valid DKIM as well. Per the DKIM spec: 

"d= The SDID claiming responsibility for an introduction of a message
      into the mail stream (plain-text; REQUIRED)."
      
 My read is that it must exist for each dkim signature. I'll add the check.

PR #422, opened 2025-10-04
support ip4, ip6 with no subnet

kayozaki:
Add support to handle ip4, ip6 directive with no subnet being defined.


wez:
Thank you!

PR #424, opened 2025-10-06
Add support for dmarc pct

sophiajt:
This adds initial support for `pct=` to the DMARC support. Currently, we pass the message with a message that it was skipped, though we may want to explore a status that lets us know the message was skipped.

Testing is fairly simple, using a large sampling test to ensure that the pct is kicking in and only testing x percent of the messages. This can fail if it finds an outlier run, which currently would lead to a spurious failure.

### DIFF
--- a/crates/kumo-dmarc/Cargo.toml
+++ b/crates/kumo-dmarc/Cargo.toml
@@ -9,7 +9,7 @@ dns-resolver = {path="../dns-resolver"}
 instant-xml = {workspace=true}
 kumo-spf = { path = "../kumo-spf" }
 serde = {workspace=true}
-rand = "0.9"
+rand = {workspace=true}
 
 [dev-dependencies]
 k9 = {workspace=true}
--- a/crates/kumo-dmarc/Cargo.toml
+++ b/crates/kumo-dmarc/Cargo.toml
@@ -9,6 +9,7 @@ dns-resolver = {path="../dns-resolver"}
 instant-xml = {workspace=true}
 kumo-spf = { path = "../kumo-spf" }
 serde = {workspace=true}
+rand = "0.9"
 
 [dev-dependencies]
 k9 = {workspace=true}
--- a/crates/kumo-dmarc/src/tests.rs
+++ b/crates/kumo-dmarc/src/tests.rs
@@ -132,6 +132,34 @@ async fn dmarc_spf_strict_subdomain() {
     k9::assert_equal!(result.result, DmarcResult::Fail);
 }
 
+#[tokio::test]
+async fn dmarc_pct_rate() {
+    let mut total_failures = 0;
+
+    for _ in 0..10000 {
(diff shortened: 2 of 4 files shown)


crates/kumo-dmarc/src/tests.rs
```diff
@@ -132,6 +132,34 @@ async fn dmarc_spf_strict_subdomain() {
...
+async fn dmarc_pct_rate() {
+    let mut total_failures = 0;
+
+    for _ in 0..10000 {
+        let resolver = TestResolver::default().with_zone(EXAMPLE_COM).with_txt(
+            "example.com",
+            "v=DMARC1; p=reject; aspf=s; pct=50; \
+                rua=mailto:dmarc-feedback@example.com"
+                .to_string(),
+        );
```
wez: a little bit of micro-optimization here to avoid creating so many identical resolvers

```suggestion
        let resolver = TestResolver::default().with_zone(EXAMPLE_COM).with_txt(
            "example.com",
            "v=DMARC1; p=reject; aspf=s; pct=50; \
                rua=mailto:dmarc-feedback@example.com"
                .to_string(),
        );
        for _ in 0..10000 {
```

crates/kumo-dmarc/Cargo.toml
```diff
@@ -9,6 +9,7 @@ dns-resolver = {path="../dns-resolver"}
 instant-xml = {workspace=true}
 kumo-spf = { path = "../kumo-spf" }
 serde = {workspace=true}
+rand = "0.9"
```
wez:
```suggestion
rand.workspace = true
```

crates/kumo-dmarc/src/types/record.rs
```diff
@@ -28,6 +28,12 @@ impl Record {
         cx: &DmarcContext<'_>,
         _resolver: &dyn Resolver,
     ) -> DmarcResultWithContext {
+        if rand::random_range(0..100) >= self.rate {
+            return DmarcResultWithContext {
+                result: DmarcResult::Pass,
+                context: "DMARC: message skipped".into(),
```
wez: How about using `sampled_out` as the terminology here; that's used in the spec in reporting PolicyOverrideTypes in reporting, so there is precedent for that term to refer to this sort of thing.

I don't know that we need to include `DMARC` as a prefix here?

So something like this?

```suggestion
                context: format!("sampled_out due to pct={}", self.rate),
```

crates/kumo-dmarc/src/types/record.rs
```diff
@@ -37,7 +43,7 @@ impl Record {
                         {
                             return DmarcResultWithContext {
                                 result: DmarcResult::Fail,
-                                context: "DMARC: DKIM relaxed check failed".into(),
+                                context: format!("sampled_out due to pct={}", self.rate),
```
wez: I think you meant the one above? :)

sophiajt: Oops, fixed!

PR #426, opened 2025-10-09
rspamd: add native integration with file header support

vstakhov:
Add native Rspamd integration using rspamd-client crate v0.4.0.

Features:
- File header optimization for local scanning (avoids message body transfer)
- ZSTD compression enabled by default
- HTTPCrypt encryption support
- Automatic metadata extraction (IP, HELO, sender, recipient)
- Flexible action handling (reject, tag, quarantine)

New components:
- mod-rspamd: Rust module with Lua bindings
- policy-extras/rspamd.lua: high-level Lua helper with setup() function
- examples/rspamd-integration.lua: comprehensive usage examples

The file header optimization could be particularly beneficial when Rspamd runs on the same host as KumoMTA, avoiding transmission of message bodies over the socket for significant performance improvement.


assets/policy-extras/rspamd.lua
```diff
@@ -0,0 +1,280 @@
...
+
+  -- Scan the message with file path optimization
+  local result = rspamd.scan_message(client, msg, conn_meta, {use_file_path = true})
+
+  -- Custom action handling
+  if result.score > 15 then
+    kumo.reject(550, '5.7.1 Message rejected as spam')
+  elseif result.score > 5 then
+    msg:set_first_named_header('X-Spam-Flag', 'YES')
+    msg:set_first_named_header('X-Spam-Score', tostring(result.score))
```
wez:
```suggestion
    msg:prepend_header('X-Spam-Flag', 'YES')
    msg:prepend_header('X-Spam-Score', tostring(result.score))
```

assets/policy-extras/rspamd.lua
```diff
@@ -0,0 +1,280 @@
...
+local client = kumo.rspamd.build_client {
+  base_url = 'http://localhost:11333',
+  timeout = '10s',
+  password = os.getenv('RSPAMD_PASSWORD'),
+  zstd = true,  -- Compress message bodies (default: true)
+  encryption_key = os.getenv('RSPAMD_ENCRYPTION_KEY'),  -- HTTPCrypt encryption
+  retries = 2,  -- Retry failed requests
+}
+
+kumo.on('smtp_server_message_received', function(msg, conn_meta)
```
wez: In `main`, I would recommend doing this in the new https://docs.kumomta.com/reference/events/smtp_server_data/ event that you probably didn't see yet because I only pushed it the other day.

The new event happens just once for an incoming batch, whereas `smtp_server_message_received` operates on essentially the same content for each `Message` that results from potentially splitting the message up.

assets/policy-extras/rspamd.lua
```diff
@@ -0,0 +1,280 @@
...
+  if conn_meta:get_meta('authn_id') then
+    return
+  end
+
+  -- Scan the message with file path optimization
+  local result = rspamd.scan_message(client, msg, conn_meta, {use_file_path = true})
+
+  -- Custom action handling
+  if result.score > 15 then
+    kumo.reject(550, '5.7.1 Message rejected as spam')
```
wez: For a bit more nuance, if someone wanted a policy that allowed per-user spam thresholds, you'd want to scan in `smtp_server_data` (so that you only scan the content once) but defer taking an action like rejecting until `smtp_server_message_received` where the recipient list has been finalized--it's possible that someone might have a policy that adds, removes or otherwise edits the recipient list.

assets/policy-extras/rspamd.lua
```diff
@@ -0,0 +1,280 @@
...
+      451,
+      string.format('4.7.1 Greylisted, please try again (score: %.2f)', result.score)
+    )
+  end,
+
+  ['rewrite subject'] = function(msg, result)
+    local subject = msg:get_first_named_header_value 'Subject' or ''
+    msg:set_first_named_header('Subject', '***SPAM*** ' .. subject)
+    msg:set_first_named_header('X-Spam-Flag', 'YES')
+    msg:set_first_named_header('X-Spam-Score', tostring(result.score))
```
wez:
```suggestion
    msg:prepend_header('X-Spam-Flag', 'YES')
    msg:prepend_header('X-Spam-Score', tostring(result.score))
```

assets/policy-extras/rspamd.lua
```diff
@@ -0,0 +1,280 @@
...
+  ['soft reject'] = function(msg, result)
+    kumo.reject(
+      451,
+      string.format('4.7.1 Greylisted, please try again (score: %.2f)', result.score)
+    )
+  end,
+
+  ['rewrite subject'] = function(msg, result)
+    local subject = msg:get_first_named_header_value 'Subject' or ''
+    msg:set_first_named_header('Subject', '***SPAM*** ' .. subject)
```
wez: We don't have a one-liner for this sort of change at the moment; the header modification methods on `Message` are relatively lightweight and targeted towards "cheap" operations like inserting a new instance of a header.

What we'd need to do for this is remove all instances of `Subject`, then insert a new one:

```suggestion
    msg:remove_all_named_headers('Subject')
    msg:prepend_header('Subject, '***SPAM*** ' .. subject)
```

assets/policy-extras/rspamd.lua
```diff
@@ -0,0 +1,280 @@
...
+    msg:set_first_named_header('X-Spam-Flag', 'YES')
+    msg:set_first_named_header('X-Spam-Score', tostring(result.score))
+  end,
+
+  ['add header'] = function(msg, result)
+    msg:set_first_named_header('X-Spam-Flag', 'YES')
+    msg:set_first_named_header('X-Spam-Score', tostring(result.score))
```
wez:
```suggestion
    msg:prepend_header('X-Spam-Flag', 'YES')
    msg:prepend_header('X-Spam-Score', tostring(result.score))
```

assets/policy-extras/rspamd.lua
```diff
@@ -0,0 +1,280 @@
...
+  if msg then
+    metadata.from = tostring(msg:sender())
+    metadata.queue_id = msg:id()
+
+    -- Get recipients
+    local recipients = {}
+    local recip = msg:recipient()
+    if recip then
+      table.insert(recipients, tostring(recip))
+    end
```
wez: Some changes landed since you first started looking at this; previously we'd split every incoming batch into 1 Message per recipient, but now in `main` we offer control over batch splitting and a message can now have multiple recipients.

You probably want to use the new https://docs.kumomta.com/reference/message/recipient_list/ method here to capture an array of recipients:

```suggestion
    for _, recip in ipairs(msg:recipient_list()) do
      table.insert(recipients, tostring(recip))
    end
```

assets/policy-extras/rspamd.lua
```diff
@@ -0,0 +1,280 @@
...
+    -- Optionally use file path for local scanning (File header optimization)
+    -- This avoids sending the message body over the socket
+    -- Only use this if Rspamd is on localhost or unix socket
+    if options.use_file_path then
+      local spool_path = msg:get_meta 'spool_path'
+      if spool_path then
+        metadata.file_path = spool_path
+      end
+    end
+  end
```
wez: at the stage where this policy runs, the message exists only in memory.  In addition, if we are  scanning a message that has been spooled, the spool that we recommend folks to use is rocksdb which doesn't reserve a file per message; the contents are part of a structured database file that employs compression.

I don't think we can reasonably leverage the file path in this way.

What's the best way to interoperate with rspamd in this situation?  It seems to me that passing the compressed bytes in the underlying http request is probably best?  I'd like to avoid doing file IO for this if we can, as part of the "secret" behind our performance is minimizing file IO as much as possible!

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+
+/// Configuration for Rspamd client (Lua-friendly wrapper)
+#[derive(Deserialize, Debug, Clone)]
+struct RspamdClientConfig {
+    /// Base URL for Rspamd server (e.g., "http://localhost:11333")
+    base_url: String,
+
+    /// Optional password for Rspamd authentication
+    #[serde(default)]
+    password: Option<String>,
```
wez: for credential information, I'd suggest looking at `data_loader::KeySource` for this.

That maps to https://docs.kumomta.com/reference/keysource/ in the docs and allows abstracting out secret storage.

https://docs.kumomta.com/rustapi/data_loader/enum.KeySource.html#method.get is the primary method you'd consume here in this crate; its an async `get` method that returns the bytes from the underlying storage

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+    /// Enable ZSTD compression (default: true)
+    /// Automatically compresses message bodies for faster transmission
+    #[serde(default = "default_true")]
+    zstd: bool,
+
+    /// Optional HTTPCrypt encryption key (base32 format)
+    /// When set, enables encrypted communication with Rspamd
+    /// Generate with: rspamadm keypair
+    #[serde(default)]
+    encryption_key: Option<String>,
```
wez: Should probably use KeySource for this as well

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+use rspamd_client::scan_async;
+use serde::Deserialize;
+use std::collections::HashMap;
+use std::sync::Arc;
+
+/// Configuration for Rspamd client (Lua-friendly wrapper)
+#[derive(Deserialize, Debug, Clone)]
+struct RspamdClientConfig {
+    /// Base URL for Rspamd server (e.g., "http://localhost:11333")
+    base_url: String,
```
wez: Consider typing the various url parameters in this struct as `url::Url` so that we have stronger validation of the config file

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+    #[serde(default)]
+    encryption_key: Option<String>,
+
+    /// Optional proxy URL (e.g., "http://proxy.example.com:8080")
+    #[serde(default)]
+    proxy_url: Option<String>,
+
+    /// Optional proxy username
+    #[serde(default)]
+    proxy_username: Option<String>,
```
wez: can and probably should also use KeySource for this as well; it's not as critical as the key material but it does mean that both the identity and key can be managed by eg: vault that way

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+    #[serde(default)]
+    tls_cert_path: Option<String>,
+
+    /// Optional TLS key path
+    #[serde(default)]
+    tls_key_path: Option<String>,
+
+    /// Optional TLS CA path
+    #[serde(default)]
+    tls_ca_path: Option<String>,
```
wez: we do also have a precedent of using KeySource for tls parameters like this as well

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+            rcpt: table.get("rcpt").unwrap_or_default(),
+            ip: table.get("ip").ok(),
+            user: table.get("user").ok(),
+            helo: table.get("helo").ok(),
+            hostname: table.get("hostname").ok(),
+            queue_id: table.get("queue_id").ok(),
+            file_path: table.get("file_path").ok(),
+            additional_headers: table.get("additional_headers").unwrap_or_default(),
+        })
+    }
```
wez: I'd suggest using `#[derive(FromLua)]` on `EnvelopeDataLua` to have mlua provide this boilerplate for you. See https://docs.rs/mlua/latest/mlua/derive.FromLua.html and https://docs.rs/mlua/latest/mlua/trait.FromLua.html

You probably would need to tweak `additional_headers` to be `Option<HashMap<String, String>>` and handle defaulting that when you convert to `EnvelopeData`.

assets/policy-extras/rspamd.lua
```diff
@@ -0,0 +1,280 @@
...
+  if handler then
+    handler(msg, result)
+  else
+    -- Default: add headers for unknown actions
+    kumo.log_warn(
+      string.format(
+        "Unknown Rspamd action '%s', adding headers only",
+        action
+      )
+    )
```
wez: For this sort of thing, I'd prefer to catch an unknown action when we set up the module, and surface it as an error, so that we can avoid filling the diagnostic logs with warnings.

We have a validation concept that can be triggered by running `kumod --validate` which we suggest folks to trigger in their validation/deployment pipelines.

https://github.com/KumoCorp/kumomta/blob/f582d601598ce086af481d164f1e7415bcf68851/assets/policy-extras/queue.lua#L551 shows one fairly complex example of a validator.

You don't have to implementation validation right now; you could just catch the invalid handler during setup and flag it as an error.

Also in that queue.lua is an example of using our typing helper that you might consider using to represent your lua config struct.  It's similar to the python typing module.

assets/policy-extras/rspamd.lua
wez, line 1: I have a few comments here on this lua file, but after making a pass through it, I think that it might simplify things if this was all implemented in the mod-rspamd module in Rust.  It would sidestep some marshaling of data to/from lua and simplify some validation.

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,212 @@
...
+            additional_headers: self.additional_headers,
+        }
+    }
+}
+
+/// Register the rspamd module with Lua
+pub fn register(lua: &Lua) -> anyhow::Result<()> {
+    let rspamd_mod = get_or_create_sub_module(lua, "rspamd")?;
+
+    rspamd_mod.set(
```
wez, line 307: as mentioned as a comment on the lua file as a whole, I think that we could probably simplify and streamline this integration a bit if we exposed a lua function here:

```lua
kumo.rspamd.scan_message(config, msg)
```

where `config` is the `RspamdClientConfig`, `msg` is a `message::Message` (will need to add a dep on the `message` crate for this.

In here, you'd do something like:

```rust
rspamd_mod.set("scan_message",
   lua.create_async_function(|lua, (config, msg): (mlua::Value, Message)| async move {
       let config: RspamdClientConfig = from_lua_value(lua, options)?;

       let envelope_data = EnvelopeData {
           // populate directly from msg here.
           // see https://docs.kumomta.com/rustapi/message/message/struct.Message.html
            // for the underlying rust methods you can use to extra metadata, headers, envelope and so on
            from: msg.sender().map(|s| s.to_string()).ok(),
            rcpt: msg.recipient_list_string().unwrap_or_default(),
            ip: msg.get_meta_string('received_from').map_err(any_err)?,
            user: msg.get_meta_string('authn_id').map_err(any_err)?,
            helo: msg.get_meta_string('ehlo_domain').map_err(any_err)?,
            // note that this is *our* hostname, not what we think the hostname is of the peer
            hostname: msg.get_meta_string('hostname').map_err(any_err)?,
            // as mentioned elsewhere, we don't really have a file path that we
            // can pass to rspamd in most cases
            file_path: None, 
            // maybe we configure a list of additional header and/or meta names to pass here via RspamdClientConfig?
            additional_headers: HashMap::default(),
       };

       // `make_client_config` is logically equivalent to RspamdClient::new, but
       // returns the config object instead of making a client object
       let client_config = config.make_client_config()?;

       let reply = scan_async(client_config, msg.get_data(), envelope_data).await.map_err(any_err)?;

       // if we wanted to do some default actions, we could apply them here.
       // See additional comments below

       // since RspamdScanReply is Serialize, we can serialize it into
       // a lua object as part of the return value
       lua.to_value(reply)
})?)?;
```

For the default actions, we try not to do too much by default, but for scanning I think there is value to making it easy to do the common options.

What I'd suggest is adding a couple of flags to `RspamdClientConfig`:

```rust
struct RspamdClientConfig {
   // If true, add the X-Spam-XXX headers
   #[serde(default = "default_true")]
   add_headers: bool,

   // If true, prefix the Subject header with SPAM
   #[serde(default)]
   prefix_subject: bool,

   // if true, reject spam
   #[serde(default)]
   reject_spam: bool,

   // if true, and reject_spam is true, use a 4xx rather than 5xx rejection
   #[serde(default)]
   reject_soft: bool,
   ...
}
```

then we could react to those before we serialize the scan reply.

But! There is a wrinkle. To do the equivalent of `kumo.reject` we'd need access to the `RejectError` type that we cannot access directly because of the dependency graph shape.

There are two ways around that:

1. Factor it out into a separate crate
2. Look up `kumo.reject` via `lua.globals()` and call it via the lua runtime!  See https://docs.rs/mlua/latest/mlua/struct.Lua.html#method.globals. You can then do the two-step resolve for `kumo` then `reject`.  If you `Table::get::<mlua::Function>()` to resolve `reject`, you can then call it via https://docs.rs/mlua/latest/mlua/struct.Function.html#method.call

I'd suggest trying the second one of these for now for the sake of expediency.  It's not ideal, but I don't want you to have to move mountains just to get this workable.

wez (commented):
Thank you for this!
I have quite a few comments/suggestions on how to integrate this better, and some are because of changes that only got pushed this week, likely while you weren't looking!

For testing, I'd suggest looking at the redis and amqp integrations.

For amqp, we leverage https://docs.rs/testcontainers-modules/0.13.0/testcontainers_modules/rabbitmq/struct.RabbitMq.html to run a server in a docker container:

https://github.com/KumoCorp/kumomta/blob/f582d601598ce086af481d164f1e7415bcf68851/crates/integration-tests/src/rabbit.rs#L97-L101

For redis, we launch an isolated copy of the redis server ourselves (no need for a container), on a port that we allocate dynamically:

https://github.com/KumoCorp/kumomta/blob/f582d601598ce086af481d164f1e7415bcf68851/crates/mod-redis/src/test.rs#L31-L49

What I think we should have for rspamd testing is similar to those: if we find rspamd installed in a reasonable path, then we can spawn it with an isolated configuration from a unit test, then run with a policy file that will trigger rspamd testing appropriately.  Look at:

https://github.com/KumoCorp/kumomta/blob/f582d601598ce086af481d164f1e7415bcf68851/crates/integration-tests/src/test/queue_ndr.rs#L13-L17

for an example of an integration test that uses a custom policy. That particular one loads in https://github.com/KumoCorp/kumomta/blob/main/crates/integration-tests/ndr.lua

crates/mod-rspamd/src/lib.rs
vstakhov: Yeah, I'll try the second approach for now, as it will be like one extra lua pcall for a message that is not so expensive comparing to everything else.

crates/integration-tests/src/rspamd.rs
```diff
@@ -0,0 +1,545 @@
...
+            }
+            _ => {
+                if attempt == 10 {
+                    anyhow::bail!("Rspamd did not become ready in time");
+                }
+                eprintln!("Waiting for Rspamd to be ready (attempt {})", attempt);
+                tokio::time::sleep(Duration::from_secs(1)).await;
+            }
+        }
+    }
```
wez: Since this block is used verbatim (or very close to verbatim!) in the other tests in this file, let's factor it out into a little helper:

```rust
struct Rspamd {
    container: Container,
    url: String,
}

impl Rspamd {
    pub async fn start() -> anyhow::Result<Self> {
       // essentially the code in the highlighted section,
       // just capture the pertinent bits at the end
       INSERT CODE HERE!
       Ok(Self { container: rspamd_container, url: rspamd_url})
    }

    pub fn url(&self) -> &str {
        &self.url
    }
}
```

then each test can simplify the startup to something like:

```rust
let rspamd = Rspamd::start().await?;
```

and see my comments below about the policy script handling and passing the URL into it

crates/integration-tests/src/rspamd.rs
```diff
@@ -0,0 +1,545 @@
...
+  kumo.define_spool {{
+    name = 'meta',
+    path = '/tmp/kumo-test-spool/meta',
+  }}
+end)
+
+-- Use smtp_server_data for per-batch scanning
+kumo.on('smtp_server_data', function(msg)
+  local config = {{
+    base_url = '{rspamd_url}',
```
wez: The pattern that we use for this sort of parameter is to pass it down via the environment.  Here's an example with our traffic shaping automation tests:

https://github.com/KumoCorp/kumomta/blob/aac842439fe980ee99a95353f4d224f86056feee/crates/integration-tests/tsa_source.lua#L8-L12

What I think we can do here is a bit simpler:

```lua
base_url = os.getenv 'KUMOD_TEST_RSPAMD_URL',
```

that env var can be set when we spawn the daemon; I will comment at the appropriate place

crates/integration-tests/src/rspamd.rs
```diff
@@ -0,0 +1,545 @@
...
+                eprintln!("Waiting for Rspamd to be ready (attempt {})", attempt);
+                tokio::time::sleep(Duration::from_secs(1)).await;
+            }
+        }
+    }
+
+    // Create temporary Lua config file
+    let lua_config = format!(
+        r#"
+local kumo = require 'kumo'
```
wez: We generally prefer to keep the lua files as plain files on the filesystem in the repo because it is a bit easier to edit the lua logic when your editor can apply highlighting, and we have some linting and code formatting tools that will apply to them as well.

We tend to place those lua policy files in the `crates/integration-tests` directory.

When there are slightly different scenarios to test, we tend to make the policy file capable of exercising all of them, but have the client trigger alternate policies based on things that are simple to control at injection time.

For example, you can embed a significant identifier in the local part of the sender or recipient email address and put some conditional logic around that to call the scan with the appropriate parameters if needed.

crates/integration-tests/src/rspamd.rs
```diff
@@ -0,0 +1,545 @@
...
+    listen = '0.0.0.0:0',
+  }}
+
+  kumo.start_http_listener {{
+    listen = '127.0.0.1:0',
+  }}
+
+  kumo.define_spool {{
+    name = 'data',
+    path = '/tmp/kumo-test-spool/data',
```
wez: there's a potential test isolation (and thus flakeyness) issue here because this path can potentially conflict with other potential concurrent test runs.

Please adopt the technique used over here for this and the meta spool below:

https://github.com/KumoCorp/kumomta/blob/aac842439fe980ee99a95353f4d224f86056feee/crates/integration-tests/source.lua#L65

crates/integration-tests/src/rspamd.rs
```diff
@@ -0,0 +1,545 @@
...
+"#
+    );
+
+    let config_file = std::env::temp_dir().join("rspamd_test_config.lua");
+    std::fs::write(&config_file, lua_config).context("write lua config")?;
+
+    let mut daemon = crate::kumod::DaemonWithMaildirOptions::new()
+        .policy_file(config_file.to_string_lossy().to_string())
+        .start()
+        .await?;
```
wez, line 195: Per my comment above about passing down the rspamd URL, combined with factoring out the container launching, something like this here would do the job:

```rust
 let mut daemon = DaemonWithMaildirOptions::new()
        // set the path to whatever you choose to move the inlined lua logic into
        .policy_file("rspamd.lua")
        .env("KUMOD_TEST_RSPAMD_URL", rspamd.url())
        .start()
        .await?;
```

crates/integration-tests/src/rspamd.rs
```diff
@@ -0,0 +1,545 @@
...
+  }}
+
+  local result = kumo.rspamd.scan_message(config, msg)
+
+  -- Store scan results in message metadata
+  msg:set_meta('rspamd_score', result.score)
+  msg:set_meta('rspamd_action', result.action)
+
+  -- Log for debugging
+  print(string.format('Rspamd scan: score=%.2f action=%s', result.score, result.action))
```
wez: I'd recommend doing something like this:

```lua
kumo.log_info('Rspam scan', kumo.serde.json_encode_pretty(result))
```

for this sort of debug print; that way you can see a dump of the whole scan result.

crates/integration-tests/src/rspamd.rs
```diff
@@ -0,0 +1,545 @@
...
+    eprintln!("Message successfully delivered after Rspamd scan");
+
+    daemon.stop_both().await.context("stop_both")?;
+    eprintln!("Test completed successfully");
+
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_rspamd_reject_spam() -> anyhow::Result<()> {
```
wez, line 254: the name of this test gives me the expectation that it will test a rejection, but it looks otherwise the same as the test case above, just with a smaller message.

Should this be renamed, or should the input be something like a GTUBE payload?

crates/integration-tests/src/rspamd.rs
```diff
@@ -0,0 +1,545 @@
...
+  msg:set_meta('rspamd_score', result.score)
+  msg:set_meta('rspamd_action', result.action)
+
+  -- Verify headers were added
+  local spam_flag = msg:get_first_named_header_value('X-Spam-Flag')
+  local spam_score = msg:get_first_named_header_value('X-Spam-Score')
+  local spam_action = msg:get_first_named_header_value('X-Spam-Action')
+
+  print(string.format('Headers: Flag=%s Score=%s Action=%s',
+    spam_flag or 'nil', spam_score or 'nil', spam_action or 'nil'))
```
wez: let's move the header assertions to the rust code in the integration test.  The pattern for this sort of thing is to parse the message(s) out of the maildir and look for the headers.

Here's an example:

https://github.com/KumoCorp/kumomta/blob/aac842439fe980ee99a95353f4d224f86056feee/crates/integration-tests/src/test/end_to_end.rs#L64-L71

Moving the assertions to the rust code makes it easier to reuse the same test policy file with the other test cases and reduces the amount of similar code, which makes things easier for human brains in the long run!

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+    #[serde(default)]
+    tls_certificate: Option<KeySource>,
+
+    /// Optional TLS private key
+    #[serde(default)]
+    tls_private_key: Option<KeySource>,
+
+    /// Optional TLS CA certificate
+    #[serde(default)]
+    tls_ca_certificate: Option<KeySource>,
```
wez: if these must be paths, then I retract my earlier suggestion to use KeySource for these; they can be simple strings here

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+
+fn default_true() -> bool {
+    true
+}
+
+impl RspamdClientConfig {
+    /// Build rspamd-client Config from this configuration
+    fn make_client_config(&self) -> anyhow::Result<Config> {
+        // Helper to extract file paths from KeySource (for TLS certificates)
+        let extract_path = |key_source: &KeySource, param_name: &str| -> anyhow::Result<String> {
```
wez: as mentioned on the struct fields above, if you need paths then using KeySource here is too forced and will be a bit surprising to have it raise errors in some of the cases below.

Let's just switch to using strings for the options that must be paths in the rspamd client and keep things simpler!

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+                    KeySource::Data { key_data, .. } => Ok(key_data.clone()),
+                    KeySource::Vault { .. } => {
+                        anyhow::bail!(
+                            "{} does not support Vault secrets yet. \
+                            Consider using a file or inline data instead.",
+                            param_name
+                        )
+                    }
+                }
+            };
```
wez: I think you can/should remove this block and/or just use `key_source.get().await` to get at the bytes from the secret, so that any future KeySource variants will be handled appropriately.

If you require a string rather than bytes, you can operate on the returned bytes, so something like:

```rust
let password = match &self.password {
    Some(pw) => Some(String::from_utf8(&pw.get().await?)?),
    None => None,
};
```

Wrapping this up in a lambda like the highlighted block to make it easier and less repeaty and have better error context is welcomed/fine; the main objection is recreating a partial implementation of the `KeySource::get` operation!

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,378 @@
...
+                // We don't have a reliable file path in most cases
+                file_path: None,
+                body_block: false,
+                additional_headers: HashMap::default(),
+            };
+
+            // Build client config and scan
+            // Convert Arc<Box<[u8]>> to Vec<u8> for scan_async
+            let message_data = msg.get_data();
+            let message_bytes: Vec<u8> = message_data.as_ref().to_vec();
```
wez: I'd assume that this would be sufficient for this. Was there a nuance that required breaking it up?

```suggestion
            let message_bytes = msg.get_data().to_vec();
```

That said, it's highly desirable to avoid making this separate copy, as this doubles the memory used to handle message payloads in a receiving setup, at least for the duration of a scan.  Reducing this reduces the DoS potential.

Could we do something here using `Bytes::from_owner(msg.get_data())`?

https://docs.rs/bytes/1.10.1/bytes/struct.Bytes.html#method.from_owner

wez (commented):
This is looking really good, thank you!
A few more comments below!

crates/mod-rspamd/src/lib.rs
vstakhov: As I've mentioned, I'm not quite sure that we should even add TLS stuff, as we have native httpcrypt support that covers all use cases. I'll probably just remove TLS from the whole thing as it will just confuse users.

wez: My suggestion would be to retain those options, but bias the documentation towards httpcrypt if that works out simplest.  There are kumomta users that have FIPS and other similar requirements that might render httpcrypt unsuitable for deployment due to it not being a widely adopted and understood protocol: it may not fall within organizational security compliance requirements.

vstakhov: It's not supported by Rspamd anyway.

wez:
> @wez it seems there might be an issue with headers rewriting in the Rust. Or maybe I don't understand the original logic...
> 
> ```diff
> --- a/crates/kumod/src/smtp_server.rs
> +++ b/crates/kumod/src/smtp_server.rs
> @@ -2619,9 +2619,12 @@ impl SmtpServerSession {
>                      )
>                  };
> 
> -                let mut body = Vec::with_capacity(data.len() + received.len());
> +                // Use base_message.get_data() instead of original data to preserve
> +                // any modifications made in smtp_server_data hook (e.g., Rspamd headers)
> +                let base_data = base_message.get_data();
> +                let mut body = Vec::with_capacity(base_data.len() + received.len());
>                  body.extend_from_slice(received.as_bytes());
> -                body.extend_from_slice(&data);
> +                body.extend_from_slice(&base_data);
>                  Arc::new(body.into_boxed_slice())
>              } else {
>                  base_message.get_data()
> ```
> 
> This change looks logical to me but please tell me if I'm wrong.

Oh, good catch! I pushed a fix for this in 42b65b22852db6ccbc92e6954657a7412764952d
Thanks!

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+                file_path: None,
+                body_block: false,
+                additional_headers: HashMap::default(),
+            };
+
+            // Build client config and scan
+            // scan_async accepts Into<Bytes>, so we convert Arc<Box<[u8]>> to Vec<u8>
+            // which then gets wrapped in Bytes. This minimizes intermediate allocations.
+            let client_config = config.make_client_config().await.map_err(any_err)?;
+            let reply = scan_async(&client_config, (*msg.get_data()).to_vec(), envelope_data)
```
wez: Should be able to do something like this to get zero-copy of the data bytes:

```suggestion
           struct DataWrapper(Arc<Box<[u8]>>);
           impl AsRef<[u8]> for DataWrapper {
               fn as_ref(&self) -> &[u8] {
                   &*self.0
                }
            } 
            let data = Bytes::from_owner(DataWrapper(msg.get_data()));

            let client_config = config.make_client_config().await.map_err(any_err)?;
            let reply = scan_async(&client_config, data, envelope_data)
```

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+                        .unwrap_or("4.7.1 Greylisted, please try again later");
+
+                    reject.call::<()>((451, message))?;
+                }
+                "add header" => {
+                    // Just add headers (already done above) and deliver
+                }
+                "rewrite subject" => {
+                    // Rewrite subject if prefix_subject is enabled
+                    if config.prefix_subject {
```
wez: Can reduce nesting by joining the condition with the match like this:

```suggestion
                "rewrite subject" if config.prefix_subject => {
                    // Rewrite subject if prefix_subject is enabled
                   
```

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+    add_headers: bool,
+
+    /// If true, prefix the Subject header with "***SPAM*** " when action is "rewrite subject" (default: false)
+    /// Note: This only applies when Rspamd returns action "rewrite subject"
+    #[serde(default)]
+    prefix_subject: bool,
+
+    /// If true, reject messages when Rspamd action is "reject" (default: false)
+    /// If false, messages with action "reject" will still be delivered with spam headers
+    /// Note: Action "soft reject" always results in 451 temporary failure (greylisting)
```
wez: is there a straight forward way for a user to prevent rspamd returning a soft reject?  I can see this sort of unilateral policy decision being something that one of our users will ask for a way to override so that they can handle it for themselves in a possible weird/inventive way!

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,340 @@
...
+    fn get_config(&self) -> mlua::Result<&Config> {
+        Ok(&self.config)
+    }
+
+    async fn scan(
+        &self,
+        message: String,
+        envelope: EnvelopeDataLua,
+    ) -> anyhow::Result<RspamdScanReply> {
+        let config = self.get_config().map_err(|e| anyhow::anyhow!("{}", e))?;
```
wez: nit: prefer this form of format string expansion, as it is less prone to drifting in unexpected ways across future refactoring!

```suggestion
        let config = self.get_config().map_err(|e| anyhow::anyhow!("{e}"))?;
```

wez (commented):
Just submitting a couple of comments for now; I can see that you're active so I'll review again a little later!

crates/mod-rspamd/src/lib.rs
vstakhov: Soft reject is very special and it's efficiently returned merely by `greylist` or `ratelimit`. In fact, a user should explicitly enable all of this to do soft rejection. And I assume that there is no other way to act properly for MTA aside of emitting of the temporary fail. If users are relying on greylisting/ratelimiting by MTA they should not use Rspamd modules for that, so I guess it's safe for all the use cases.

crates/mod-rspamd/src/lib.rs
vstakhov: I've totally overlooked that `to_vec` indeed.

vstakhov: I need to update rspamdclient-rs to make it happen as well.

vstakhov:
Any updates?

crates/integration-tests/src/rspamd.rs
```diff
@@ -0,0 +1,413 @@
...
+                anyhow::bail!("Expected X-Spam-Action header to be present");
+            }
+        }
+        Ok(self)
+    }
+}
+
+#[tokio::test]
+async fn test_rspamd_scan_message() -> anyhow::Result<()> {
+    if std::env::var("KUMOD_TESTCONTAINERS").unwrap_or_else(|_| String::new()) != "1" {
```
wez: I think you can simplify a little like this:

```suggestion
    if std::env::var("KUMOD_TESTCONTAINERS").unwrap_or_default() != "1" {
```

crates/integration-tests/src/rspamd.rs
```diff
@@ -0,0 +1,413 @@
...
+        return Ok(());
+    }
+
+    let rspamd = RspamdContainer::new().await?;
+
+    let policy_file = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("rspamd.lua");
+
+    let mut daemon = crate::kumod::DaemonWithMaildirOptions::new()
+        .policy_file(policy_file.to_string_lossy().to_string())
```
wez: this is more consistent with how the other tests do this:

```suggestion
        .policy_file("rspamd.lua")
```

crates/integration-tests/src/rspamd.rs
```diff
@@ -0,0 +1,413 @@
...
+        .wait_for_maildir_count(1, Duration::from_secs(10))
+        .await;
+
+    // Extract and verify Rspamd headers were added
+    // Read from source daemon's maildir (not sink) since we're delivering locally
+    let mut messages = vec![];
+    let md = daemon.source.maildir();
+    for entry in md.list_new() {
+        messages.push(entry?);
+    }
```
wez: This had me wondering why `daemon.extract_maildir_messages()` wasn't being used, and it took me a bit to realize that it is because the source is writing directly to the maildir.

Typically the `DaemonWithMaildir` test helper runs with both a source and a sink MTA, but the way things are rigged up here there is an extraneous sink that is unused.

I think there are a couple of ways to go to reconcile this:

1. Switch the test to spawning just a single `KumoDaemon` directly. This is relatively high effort because `DaemonWithMaildirOptions` has some nice helpers for setting up the test parameters and consuming data, and there isn't currently an analog for all of that in `KumoDaemon`.  I don't recommend this here; it's too much "off topic" effort.
2. Refactor `crates/integration-tests/src/kumod.rs` to add `KumoDaemon::extract_maildir_messages` with the guts of the current `DaemonWithMaildir::extract_maildir_messages`, then adjust the latter to call down to it as `self.sink.extract_maildir_messages()`.  This would enable you to use `daemon.source.extract_maildir_messages()` here.
3. Use `daemon.extract_maildir_messages()` here to consume from the sink. Adjust the `rspam.lua` integration test policy to relay messages on to the sink as is done in `crates/integration-tests/source.lua`.

I have a slight preference for the third option, as it makes things more similar to the other integration tests, but I'll accept the second option as well.

crates/integration-tests/src/rspamd.rs
wez: I'm still unclear what this case is testing?

crates/integration-tests/rspamd.lua
```diff
@@ -0,0 +1,102 @@
...
+
+  kumo.define_spool {
+    name = 'meta',
+    path = TEST_DIR .. '/meta-spool',
+  }
+end)
+
+-- Determine per-recipient spam threshold
+local function get_spam_threshold_for_user(recipient)
+  if recipient:match '@vip%.example%.com$' then
```
wez: Can use the `.domain` field to make this is a bit easier to consume:

```suggestion
  if recipient.domain == 'vip.example.com' then
```

https://docs.kumomta.com/reference/address/

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,345 @@
...
+                msg.prepend_header(
+                    Some("X-Spam-Flag"),
+                    if is_spam { "YES" } else { "NO" },
+                );
+                msg.prepend_header(Some("X-Spam-Score"), &reply.score.to_string());
+                msg.prepend_header(Some("X-Spam-Action"), &reply.action);
+
+                // Add symbols if available
+                if !reply.symbols.is_empty() {
+                    let symbols: Vec<String> = reply.symbols.keys().cloned().collect();
```
wez: Since we're going to join things into a string, I think you can make this collect `&str` rather than copying the individual symbol names here?

```suggestion
                    let symbols: Vec<&str> = reply.symbols.keys().collect();
```

crates/mod-rspamd/src/lib.rs
```diff
@@ -0,0 +1,345 @@
...
+
+                    reject.call::<()>((451, message))?;
+                }
+                "add header" => {
+                    // Just add headers (already done above) and deliver
+                }
+                "rewrite subject" if config.prefix_subject => {
+                    if let Ok(Some(subject)) = msg.get_first_named_header_value("Subject") {
+                        msg.remove_all_named_headers("Subject").map_err(any_err)?;
+                        msg.prepend_header(Some("Subject"), &format!("***SPAM*** {}", subject));
```
wez:
```suggestion
                        msg.prepend_header(Some("Subject"), &format!("***SPAM*** {subject}"));
```

assets/policy-extras/rspamd.lua
wez: I think this file can be removed now that all the work is done in Rust?

examples/rspamd-integration.lua
```diff
@@ -0,0 +1,672 @@
+--[[
```
wez, line 1: This will need to be updated to match the rust-centric implementation

crates/mod-rspamd/README.md
```diff
@@ -0,0 +1,282 @@
+# mod-rspamd: Native Rspamd Integration for KumoMTA
```
wez, line 1: this will need to be updated to match the rust-centric implementation

wez (commented):
Thanks for this; sorry for the delay in reviewing since the last time, we've had a lot going on!

Most of my comments are rather nit-picky around the tests, but the core looks good.
There's a bit of tidying up to do to match the current rust-centric implementation as compared to the original more lua-centric approach.

There's something that I think we should do to enable "connection pooling" (or rather, client-object-caching), but I think that we can do that in a follow-up PR: I'd like to get this one buttoned up so that the caching can be in its own smaller PR.

dandigangi:
Following! We'd be interested in having this at Postmark when it gets into Kumo.

vstakhov:
Ok, I have fixed various issues around this PR and added connection pooling.

PR #430, opened 2025-10-15
Add psl to dmarc relaxed checks

sophiajt:
This moves the aspf and adkim checks in dmarc to use the organizational domain rather than substring matching, via the psl crate.

Also added a few more tests to ensure we're getting proper failures in both cases and that we handle deep subdomains in both cases.

### DIFF
--- a/crates/kumo-dmarc/src/tests.rs
+++ b/crates/kumo-dmarc/src/tests.rs
@@ -29,6 +29,54 @@ async fn dmarc_dkim_relaxed_subdomain() {
 }
 
 #[tokio::test]
+async fn dmarc_dkim_relaxed_subdomain_deep() {
+    let resolver = TestResolver::default()
+        .with_zone(EXAMPLE_COM)
+        .unwrap()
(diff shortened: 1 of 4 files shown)


crates/kumo-dmarc/src/tests.rs
```diff
@@ -28,6 +28,54 @@ async fn dmarc_dkim_relaxed_subdomain() {
...
+        );
+
+    let result = evaluate_ip(
+        Ipv4Addr::LOCALHOST,
+        "a.b.c.sample.example.com",
+        "a.b.c.sample.example.com",
+        &[Some("example.com")],
+        &resolver,
+    )
+    .await;
```
wez, line 50: as an aside, I feel like this argument list is long enough and similar enough that it isn't easy to tell at a glance, especially in a PR, what each parameter is for.  Probably worth switching `evaluate_ip` to receive a struct so that we can have "named parameters" and clear that up.  Doesn't (shouldn't!) have to be in this PR!

PR #432, opened 2025-10-15
Move dmarc tests to use struct

sophiajt:
For readability in non-lsp environments, move dmarc tests to use a struct to effectively name params.


wez:
Thank you!

PR #433, opened 2025-10-17
Add dmarc p/sp support

sophiajt:
This is an initial go at implementing p/sp tag support for DMARC. To do so, we're adding the following:

- Both from domain and organizational domain lookups (4.3)
- Move to searching for `_dmarc` prefix when searching for domains (6.1)
- New Disposition value that can represent success and all policy forms of failure
- Corresponding updates to tests, including testing of the `sp` tag in the case where it would apply

### DIFF
--- a/crates/kumo-dmarc/src/lib.rs
+++ b/crates/kumo-dmarc/src/lib.rs
@@ -52,9 +52,13 @@ impl CheckHostParams {
     }
 }
 
-pub enum SenderLocation {
-    Domain,
-    Subdomain,
+pub(crate) enum SenderDomainAlignment {
+    /// Sender domain is an exact match to the dmarc record
+    Exact,
+
+    /// Sender domain has no exact matching dmarc record
+    /// but its organizational domain does
+    OrganizationalDomain,
 }
 
 struct DmarcContext<'a> {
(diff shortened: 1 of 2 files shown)


crates/kumo-dmarc/src/lib.rs
```diff
@@ -81,46 +86,74 @@ impl<'a> DmarcContext<'a> {
...
+    // TXT records can contain all sorts of stuff, let's walk through
+    // the set that we retrieved and take the first one that parses
+    for txt in initial_txt {
+        if txt.starts_with("v=DMARC1;") {
+            if let Ok(record) = Record::from_str(&txt) {
+                records.push(record);
+            }
         }
     }
+
```
wez, line 163: I think it's worth doing something like this here, to avoid unexpected weirdness if the DNS didn't have any dmarc records for this domain:

```suggestion
if records.is_empty() {
   return None;
}
```

crates/kumo-dmarc/src/lib.rs
```diff
@@ -52,6 +52,11 @@ impl CheckHostParams {
     }
 }
 
+pub enum SenderLocation {
+    Domain,
+    Subdomain,
+}
```
wez: The naming feels not quite there. How about something like this that more closely matches the nomenclature in the RFC?
I'm also open to alternatives!

```suggestion
pub(crate) enum SenderDomainAlignment {
    /// Sender domain has an exactly matching dmarc record
    Exact,
    /// Sender domain has no exact matching dmarc record
    /// but its organizational domain does
    OrganizationalDomain,  // Or `OrgDomain` for brevity?
}
```

PR #434, opened 2025-10-23
Feature/traffic-gen custom headers

cristian-porta:
Allows running targeted tests on specific queues (e.g., tenant, campaign), and measuring behaviors or latency impacts introduced by custom header manipulation

### DIFF
--- a/crates/traffic-gen/src/main.rs
+++ b/crates/traffic-gen/src/main.rs
@@ -133,7 +133,10 @@ fn parse_header(arg: &str) -> Result<MailHeader<'static>, String> {
     if consumed < arg.len() {
         let remaining = &arg[consumed..];
         if !remaining.trim().is_empty() {
-            return Err(format!("Unexpected trailing data after header: {:?}", remaining));
+            return Err(format!(
+                "Unexpected trailing data after header: {:?}",
+                remaining
+            ));
         }
     }
     Ok(header.to_owned())


crates/traffic-gen/src/main.rs
```diff
@@ -118,6 +126,28 @@ struct Opt {
...
+    }
+
+    let value = {
+        let v = v.trim();
+        if v.contains(['\r', '\n']) {
+            v.replace(['\r', '\n'], " ")
+        } else {
+            v.to_string()
+        }
+    };
```
wez: Can just roll this together unconditionally, something like this:

```suggestion
    let value = v.replace(['\r', '\n'], " ").trim().to_string();
```

crates/traffic-gen/src/main.rs
```diff
@@ -366,15 +396,20 @@ impl Opt {
...
              Subject: test {datestamp}\r\n\
              Message-Id: {id}\r\n\
-             X-Mailer: KumoMta traffic-gen\r\n\
-             \r\n\
-             {body}"
-        )
+             X-Mailer: KumoMta traffic-gen\r\n"
+        );
+        for (name, value) in self.headers.iter() {
+            message.push_str(&format!("{name}: {value}\r\n"));
```
wez: I can't help but think: "oh, this is undoing the parsing of the --header parameter" reading this.

Is it good or bad?  I think the answer depends.  Defining the `parse_header` function lets you validate and catch bad input, which is good, but a consequence of it is that the `replace` call changes the whitespace.  That's generally neutral, but if you have a complicated multi-line header that transformation will cause it all to be put on a single physical line which might then run into line length limits in an SMTP session.

What I think might be nice is to capture the input header verbatim and use that here.  It would be great to preserve the header validation though.

How about having the arg parser and validation capture `Header` instances?

You could use https://docs.kumomta.com/rustapi/mailparsing/struct.Header.html#method.parse to parse into a `Header` and raise an error for invalid content, then when you're building out the message here you could use either `Header::write_header` or `Header::to_header_string()` to get at that original value.

wez (commented):
Thanks for this! A couple of quick thoughts below!

crates/traffic-gen/src/main.rs
cristian-porta: Right, it worked for my use case, but I was too quick and didnt pay enough attention when creating the PR :)

So... checking your comment ( including the first )

- Add `parse_header` using `MailHeader::parse`, rejecting non-whitespace trailing and returning `MailHeader<'static>` (keeps original folding/whitespace)
- Switch CLI arg `--header` to `Vec<MailHeader<'static>>` with `#[arg(value_parser=ValueParser::new(parse_header))]` so invalid headers fail early at parse time
- Serialize headers via `Header::to_header_string()`

Done (?): implemented as suggested by wez with MailHeader::parse and to_header_string()

crates/traffic-gen/src/main.rs
cristian-porta: Tried updating as suggested, so

- Add `parse_header` using `MailHeader::parse`, rejecting non-whitespace trailing and returning `MailHeader<'static>` (keeps original folding/whitespace)
- Switch CLI arg `--header` to `Vec<MailHeader<'static>>` with `#[arg(value_parser=ValueParser::new(parse_header))]` so invalid headers fail early at parse time

cristian-porta:
Sorry, I forgot to pipe `make fmt`, done

PR #438, opened 2025-10-27
Preload dkim keys to speed up signing process

naetraga:
Before we had around 180 msgs / sec on > 10000 dkim keys
After this change we're at 5xx msg / sec

### DIFF
--- a/assets/policy-extras/dkim_sign.lua
+++ b/assets/policy-extras/dkim_sign.lua
@@ -265,6 +265,22 @@ local function load_dkim_data(dkim_data_files, no_compile)
   return data
 end
 
+local function make_signer(params, algo)
+  algo = algo or 'sha256'
+
+  if algo == 'sha256' then
+    return kumo.dkim.rsa_sha256_signer(params)
+  end
+
+  if algo == 'ed25519' then
+    return kumo.dkim.ed25519_signer(params)
+  end
+
+  error(
+    string.format("invalid algo '%s' for domain '%s'", algo, params.domain)
+  )
+end
+
 local function do_dkim_sign(msg, data)
   local from_header = msg:from_header()
   if not from_header then
@@ -382,25 +398,18 @@ local function do_dkim_sign(msg, data)
   end
 end
 
-local function make_signer(params, algo)
-  algo = algo or 'sha256'
-
-  if algo == 'sha256' then
-    return kumo.dkim.rsa_sha256_signer(params)
-  end
-
-  if algo == 'ed25519' then
-    return kumo.dkim.ed25519_signer(params)
-  end
-
-  error(
-    string.format("invalid algo '%s' for domain '%s'", algo, params.domain)
-  )
-end
-
-function mod:setup(options)
+function mod:setup(files, options)
   if mod.CONFIGURED then
     error 'dkim_sign module has already been configured'
+  end
+
+  if not options then
+    options = {
+      name = 'dkim_signing_data',
+      ttl = '5 minutes',
+      capacity = 10,
+      invalidate_with_epoch = true,
+    }
   end
 
   local cached_load_data = kumo.memoize(load_dkim_data, {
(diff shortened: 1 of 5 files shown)


assets/policy-extras/dkim_sign.lua
```diff
@@ -192,6 +227,34 @@ local function load_dkim_data(dkim_data_files, no_compile)
...
+      error(
+        ("dkim domain '%s' missing selector and no base.selector"):format(
+          domain_name
+        )
+      )
+    end
+
+    local path = domain.filename
+      or ('%s/%s/%s.key'):format(DKIM_PATH, domain_name, selector)
+    local f, err = io.open(path, 'rb')
```
wez, line 242: This will cause every single lua context creation to block the IO scheduler on the current thread to do disk IO for every single DKIM key.

In addition, preloading the key data causes the cache key that is associated with the signer to occupy more RAM, which will put more pressure on the system, also likely leading to degraded performance.

It seems highly unlikely to me that this part of the PR improves performance

naetraga: i understand the breaking change for the setup method. if you don't mind, I will rework on that part by adding a 2nd argument to mod:setup?

what I'm currently working on is that my customer(s) have massive performance problems on receiving messages when we have a lot of dkim keys. everything is slowing down. in that specific case we have 38870 dkim keys. the system is under highly pressure.

i have here a snapshot before my changes during receiving with a setup of 128 GB ram 32 cores and nvme disks
Monosnap Image 2025-10-27 16-53-31

and here the same after the adjustments:
Monosnap Image 2025-10-27 16-52-26

both mailings had around the same size of recipients. 

from the technical point of view (implementation) feel free to tell me optimizations for loading the keys into the cache. if you dont mind we can also make that optional.

wez: While I trust that this change did make an improvement for your specific circumstance, just sharing the raw throughput is not sufficient information to understand what that actual bottleneck is or was.  I'd strongly recommend that you consider becoming a sponsor so that we can take the appropriate amount of time to analyze and resolve this issue.

naetraga: @wez i did some further research and wanted to let you know about the latest results.

setup:
 - fresh installed kumod
 - using same dkim keys as before
 - use the original dkim_sign.lue
 - trace down with strace / strace -fs 100 -p [pid of kumod]
 - created some charts in grafana for lruttl_cache_*
 - pre_init settings 
 
```
  kumo.set_lruttl_cache_capacity('dkim_signer_cache', 40000)
  kumo.set_lruttl_cache_capacity('dkim_key_cache', 40000)
```

checking grafana / metrics:
during receive i checked the cache hits on dkim_keys - which is always 0, while miss increases
=> insert count is increasing
=> expire count starts increasing
=> cache size is "stable"
 
Monosnap Image 2025-10-31 10-15-39

i filtered on strace with dkim and see endless file openings
Monosnap Image 2025-10-31 09-53-32

=> i would say the cache_key logic seems not to be working in my setup

naetraga: is there anything else i could trace down or did i something wrong?

wez: it's hard to interpret and reason about those metrics because they mostly show percentages rather than absolute numbers. What's on my mind is:

* is memory low? If so, all caches will be purged every 30 seconds until the headroom recovers and increases above 0. See https://docs.kumomta.com/reference/memory/
* is the cache capacity being exceeded, and the cache thrashed? is the cache size set via `kumo.set_lruttl_cache_capacity` taking effect? It should, but its not clear from what is being shown/shared
* are your traffic patterns causing many concurrent sessions to request the same signer and keys at the same time? eg: do you see the same key paths showing up in your strace output around the same time? I did notice that the dkim cache logic was using the older interface that didn't support thundering herd protection. I've pushed a commit to update that, and it may help. Note that the `lruttl_populated_count` metric will increment instead of `lruttl_insert_count` when operating this way.

Issue #419, opened 2025-09-16
Add authentication switches to traffic-gen


tommairs:
The `traffic-gen` utility is a very helpful tool for volume testing, but it is lacking the ability to test with smtp_auth.

It would be handy to add the ability to provide a username and password to the command string.
Borrowing from other popular tools, it could look something like this:
```/opt/kumomta/sbin/traffic-gen --target <your.sink.server>:25 --auth plain -au <USERNAME> -ap <PASSWORD> ```

We currently only support AUTH PLAIN, so the --auth switch is not really necessary at this time, but could be made default in the case we add other formats in the future.
The -au (--auth-username) and -ap (--auth-password) could potentially be used also for basic HTTP auth using the --http switch that already exists.

When documenting, you will need to note that `--starttls` will also be needed when using this.

PR #439, opened 2025-10-27
#419 - Add authentication switches to traffic-gen + docs

cristian-porta:
Add a selectable authentication mechanism with username/password options, validate credential handling, and apply those credentials to both SMTP (including STARTTLS re-EHLO and AUTH PLAIN negotiation) and HTTP injection requests in traffic-gen

- cli: new `--auth` option (ValueEnum) with `plain` value, `--auth-username` and `--auth-password`.
- SMTP flow: stable EHLO domain, STARTTLS + re-EHLO per RFC 3207, check AUTH capability post-TLS, perform `AUTH PLAIN`
- New helpers in `mod smtp_auth`:
  - `require_auth_mech_in(mechs: &str, mech: AuthMechanism)`; verify server advertises requested mech
  - `do_smtp_auth(client, mech, mechs, creds)`; central dispatch for mechanism specific auth logic (currently handles PLAIN)
  Helpers are intentionally isolated to avoid impacting the HTTP path
- HTTP: uses `auth_credentials()?` for reqwest `basic_auth`
- Documented the new authentication workflow, including the requirement to pair credentials with `--starttls` and support for HTTP basic auth

Locally tested with both:
```
kumo.on('smtp_server_auth_plain', function(authz, authc, password, conn_meta)
  local password_database = {
    ['daniel'] = 'tiger',
  }
  if password == '' then
    return false
  end
  return password_database[authc] == password
end)

-- Use this to lookup and confirm a user/password credential
-- used with the http endpoint
kumo.on('http_server_validate_auth_basic', function(user, password)
  local password_database = {
    ['daniel'] = 'tiger',
  }
  if password == '' then
    return false
  end
  return password_database[user] == password
end)
```

via:

```
./target/release/traffic-gen --target http://<your.sink.server>:8000 --concurrency 1 --message-count 1 --body-size 1000 --domain example.com --domain-suffix '' --header "$(echo -e 'X-Test: a\r\n b\r\n')" --header "X-Tenant: 7006175" --auth plain --auth-username daniel --auth-password tiger --starttls --http

./target/release/traffic-gen --target <your.sink.server>:25 --concurrency 1 --message-count 1 --body-size 1000 --domain example.com --domain-suffix '' --header "$(echo -e 'X-Test: a\r\n b\r\n')" --header "X-Tenant: 7006175" --auth plain --auth-username daniel --auth-password tiger --starttls
```
and other combinations

### DIFF
--- a/crates/traffic-gen/src/main.rs
+++ b/crates/traffic-gen/src/main.rs
@@ -1,7 +1,7 @@
 use anyhow::{anyhow, Context};
 use chrono::Utc;
 use clap::builder::ValueParser;
-use clap::{Parser, ValueEnum};
+use clap::Parser;
 use futures_util::stream::FuturesUnordered;
 use futures_util::StreamExt;
 use hdrhistogram::sync::Recorder;
@@ -23,11 +23,6 @@ use throttle::ThrottleSpec;
 use uuid::Uuid;
 
 const DOMAINS: &[&str] = &["aol.com", "gmail.com", "hotmail.com", "yahoo.com"];
-
-#[derive(Copy, Clone, Debug, Eq, PartialEq, ValueEnum)]
-enum AuthMechanism {
-    Plain,
-}
 
 #[derive(Clone, Debug, Parser)]
 #[command(about = "SMTP traffic generator")]
@@ -58,24 +53,19 @@ struct Opt {
     #[arg(long)]
     message_count: Option<usize>,
 
-    /// Whether to use STARTTLS for submission. Required when
+    /// Whether to use STARTTLS for submission. Often required when
     /// authenticating to SMTP targets.
     #[arg(long)]
     starttls: bool,
 
-    /// Authentication mechanism to use when credentials are supplied.
-    /// Currently only `plain` is supported for SMTP AUTH and HTTP basic auth.
-    #[arg(long, value_enum)]
-    auth: Option<AuthMechanism>,
-
     /// Username to authenticate with when using SMTP AUTH or HTTP basic auth.
     /// When targeting SMTP, `--starttls` must also be supplied.
-    #[arg(short = 'u', long = "auth-username", alias = "au")]
+    #[arg(long, requires = "auth-password")]
     auth_username: Option<String>,
 
     /// Password to authenticate with when using SMTP AUTH or HTTP basic auth.
     /// When targeting SMTP, `--starttls` must also be supplied.
-    #[arg(short = 'p', long = "auth-password", alias = "ap")]
+    #[arg(long, requires = "auth-username")]
     auth_password: Option<String>,
 
     /// Take the message contents from the specified file
@@ -336,38 +326,13 @@ impl Opt {
         self.domain = adjusted_domains;
     }
 
-    /// Decide which authentication mechanism (if any) is requested.
-    fn requested_auth(&self) -> Option<AuthMechanism> {
-        if let Some(mech) = self.auth {
-            // Explicit mechanism provided by CLI/config
-            Some(mech)
-        } else if self.auth_username.is_some() || self.auth_password.is_some() {
-            // Credentials present without an explicit mechanism, assume PLAIN
-            Some(AuthMechanism::Plain)
-        } else {
-            // No auth requested
-            None
-        }
-    }
-
-    /// Validate and return authentication credentials
     fn auth_credentials(&self) -> anyhow::Result<Option<(String, String)>> {
-        // if no auth is requested, nothing to validate, return
-        if self.requested_auth().is_none() {
-            return Ok(None);
-        }
-        // Validate presence of both username and password
-        match (self.auth_username.as_ref(), self.auth_password.as_ref()) {
-            (Some(u), Some(p)) => Ok(Some((u.clone(), p.clone()))),
-            (Some(_), None) => {
-                anyhow::bail!("--auth-password is required when specifying --auth-username")
-            }
-            (None, Some(_)) => {
-                anyhow::bail!("--auth-username is required when specifying --auth-password")
-            }
-            (None, None) => {
-                anyhow::bail!("--auth requires both --auth-username and --auth-password")
-            }
+        match (self.auth_username.as_deref(), self.auth_password.as_deref()) {
+            (Some(user), Some(pass)) => Ok(Some((user.to_owned(), pass.to_owned()))),
+            (None, None) => Ok(None),
+            _ => Err(anyhow!(
+                "both --auth-username and --auth-password must be provided together"
+            )),
         }
     }
 
@@ -510,13 +475,6 @@ impl Opt {
         let timeouts = SmtpClientTimeouts::default();
         let connect_timeout = timeouts.connect_timeout;
 
-        let requested_auth = self.requested_auth();
-        let auth_credentials = self.auth_credentials()?;
-
-        if requested_auth.is_some() && !self.starttls {
-            anyhow::bail!("--starttls must be specified when authenticating to SMTP targets");
-        }
-
         let mut client =
             tokio::time::timeout(connect_timeout, SmtpClient::new(&self.target, timeouts))
                 .await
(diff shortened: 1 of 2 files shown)


crates/traffic-gen/src/main.rs
```diff
@@ -54,10 +58,26 @@ struct Opt {
...
     starttls: bool,
 
+    /// Authentication mechanism to use when credentials are supplied.
+    /// Currently only `plain` is supported for SMTP AUTH and HTTP basic auth.
+    #[arg(long, value_enum)]
+    auth: Option<AuthMechanism>,
+
+    /// Username to authenticate with when using SMTP AUTH or HTTP basic auth.
+    /// When targeting SMTP, `--starttls` must also be supplied.
+    #[arg(short = 'u', long = "auth-username", alias = "au")]
```
wez: I try to avoid short flags as they can be (or become, over time) ambiguous or obscure; I think the slightly more verbose form saves more value in terms of mental energy than a shorter flag would save in typing energy!

In addition, I think you can have clap perform the mutual requirement check for username and password by specifying it here like this.

```suggestion
    #[arg(long, requires="auth-password")]
```

crates/traffic-gen/src/main.rs
```diff
@@ -54,10 +58,26 @@ struct Opt {
...
+    auth_username: Option<String>,
+
+    /// Password to authenticate with when using SMTP AUTH or HTTP basic auth.
+    /// When targeting SMTP, `--starttls` must also be supplied.
+    #[arg(short = 'p', long = "auth-password", alias = "ap")]
```
wez:
```suggestion
    #[arg(long, requires="auth-username")]
```

crates/traffic-gen/src/main.rs
```diff
@@ -54,10 +58,26 @@ struct Opt {
     #[arg(long)]
     message_count: Option<usize>,
 
-    /// Whether to use STARTTLS for submission
+    /// Whether to use STARTTLS for submission. Required when
+    /// authenticating to SMTP targets.
```
wez:
```suggestion
    /// Whether to use STARTTLS for submission. Often required when
    /// authenticating to SMTP targets.
```

crates/traffic-gen/src/main.rs
```diff
@@ -442,28 +502,41 @@ impl Opt {
...
+        let connect_timeout = timeouts.connect_timeout;
+
+        let requested_auth = self.requested_auth();
+        let auth_credentials = self.auth_credentials()?;
 
-        let stream =
-            tokio::time::timeout(timeouts.connect_timeout, TcpStream::connect(&self.target))
+        if requested_auth.is_some() && !self.starttls {
+            anyhow::bail!("--starttls must be specified when authenticating to SMTP targets");
+        }
```
wez: This is actually a server side policy and I think it is overly restrictive to encode this check here.  For example, we often find that folks use traffic-gen against their existing non-kumo infrastructure when evaluating and performing comparisons, and those non-kumo deployments may have no such policy in place.

It feels more flexible to me to avoid encoding this sort of policy in this particular client
```suggestion
```

crates/traffic-gen/src/main.rs
```diff
@@ -20,11 +20,15 @@ use std::sync::atomic::{AtomicUsize, Ordering};
...
-use tokio::net::TcpStream;
 use uuid::Uuid;
 
 const DOMAINS: &[&str] = &["aol.com", "gmail.com", "hotmail.com", "yahoo.com"];
 
+#[derive(Copy, Clone, Debug, Eq, PartialEq, ValueEnum)]
+enum AuthMechanism {
+    Plain,
+}
+
```
wez: TBH, I feel like this makes things more complex than we need at the moment, so I'd propose just removing it.

```suggestion
```

crates/traffic-gen/src/main.rs
```diff
@@ -54,10 +58,26 @@ struct Opt {
...
+    /// Whether to use STARTTLS for submission. Required when
+    /// authenticating to SMTP targets.
     #[arg(long)]
     starttls: bool,
 
+    /// Authentication mechanism to use when credentials are supplied.
+    /// Currently only `plain` is supported for SMTP AUTH and HTTP basic auth.
+    #[arg(long, value_enum)]
+    auth: Option<AuthMechanism>,
+
```
wez:
```suggestion
```

crates/traffic-gen/src/main.rs
```diff
@@ -311,6 +336,41 @@ impl Opt {
...
+            Some(mech)
+        } else if self.auth_username.is_some() || self.auth_password.is_some() {
+            // Credentials present without an explicit mechanism, assume PLAIN
+            Some(AuthMechanism::Plain)
+        } else {
+            // No auth requested
+            None
+        }
+    }
+
```
wez:
```suggestion
```

crates/traffic-gen/src/main.rs
```diff
@@ -311,6 +336,41 @@ impl Opt {
...
+                anyhow::bail!("--auth-password is required when specifying --auth-username")
+            }
+            (None, Some(_)) => {
+                anyhow::bail!("--auth-username is required when specifying --auth-password")
+            }
+            (None, None) => {
+                anyhow::bail!("--auth requires both --auth-username and --auth-password")
+            }
+        }
+    }
```
wez: With clap validating the parameters, this can simplify down a lot:

```suggestion
    fn auth_credentials(&self) -> Option<(&str, &str)> {
        match (self.auth_username.as_deref(), self.auth_password.as_deref()) {
           (Some(user), Some(pass)) => Some((user, pass)),
           (None, None) => None,
           _ => unreachable!("clap validation should ensure that both username and pass are set"),
        }
      }
```

crates/traffic-gen/src/main.rs
```diff
@@ -442,28 +502,41 @@ impl Opt {
...
+            auth: self.auth_credentials()?,
         })
     }
 
     async fn make_smtp_client(&self) -> anyhow::Result<SmtpClient> {
         let timeouts = SmtpClientTimeouts::default();
+        let connect_timeout = timeouts.connect_timeout;
+
+        let requested_auth = self.requested_auth();
+        let auth_credentials = self.auth_credentials()?;
```
wez:
```suggestion
```

crates/traffic-gen/src/main.rs
```diff
@@ -442,28 +502,41 @@ impl Opt {
...
+        let helo_domain = self.pick_a_domain();
+        // EHLO pre-TLS
+        let mut caps = client.ehlo(&helo_domain).await?;
 
-        if self.starttls && caps.contains_key("STARTTLS") {
+        if self.starttls {
+            anyhow::ensure!(
+                caps.contains_key("STARTTLS"),
+                "target does not advertise STARTTLS"
+            );
```
wez:
```suggestion
```

crates/traffic-gen/src/main.rs
```diff
@@ -473,6 +546,32 @@ impl Opt {
...
+            caps = client.ehlo(&helo_domain).await?;
+        }
+
+        // AUTH (only if explicitly requested)
+        if let Some(mech) = requested_auth {
+            // Extract username/password if present, or fail with an error
+            let (username, password) = auth_credentials
+                .as_ref()
+                .map(|(u, p)| (u.as_str(), p.as_str()))
+                .context("authentication credentials must be supplied with --auth-username and --auth-password")?;
```
wez:
```suggestion
        if let Some((username, password)) = self.auth_credentials() {
```

crates/traffic-gen/src/main.rs
```diff
@@ -473,6 +546,32 @@ impl Opt {
...
+                let mechs = caps
+                    .get("AUTH")
+                    .and_then(|cap| cap.param.as_deref())
+                    .context("server does not advertise AUTH capability/mechanisms")?;
+                mechs.to_owned()
+            };
+
+            // Check capability and perform AUTH for the selected mechanism
+            smtp_auth::require_auth_mech_in(&mechs_owned, mech)?;
+            smtp_auth::do_smtp_auth(&mut client, mech, &mechs_owned, (username, password)).await?;
```
wez: I think we can simplify this a lot, something like this:

```suggestion
            let mechs: Vec<&str> = caps.get("AUTH").and_then(|cap| cap.param.as_deref()).map(|mechs| mechs.split_ascii_whitespace().collect()).unwrap_or_else(Vec::new);

            if !mechs.iter().any(|mech| mech.eq_ignore_ascii_case("PLAIN")) {
               anyhow::bail!("AUTH PLAIN not advertised by target");
            }
            
            client
                    .auth_plain(username, Some(password))
                    .await
                    .context("failed to authenticate with AUTH PLAIN")?;
```

crates/traffic-gen/src/main.rs
```diff
@@ -698,3 +797,46 @@ impl Rates {
...
+                    .await
+                    .context("failed to authenticate with AUTH PLAIN")?;
+            } // any other implementation:
+              // AuthMechanism::Login => client.auth_login(user, pass).await?,
+              // AuthMechanism::CramMd5 => client.auth_cram_md5(user, pass).await?,
+              // ...
+        }
+        Ok(())
+    }
+}
```
wez: I'd leave this out; it adds more complexity than we need

```suggestion
```

wez (commented):
Thanks for this!  I think making it a bit simpler is the way to go for now; we don't support other mechanisms yet, and I feel like if/when we do add other mechanisms, we'll want to change the code around a bit to accommodate those.

For now taking out the explicit mechanism selection helps keep things a bit lighter and easier to read.

crates/traffic-gen/src/main.rs
```diff
@@ -442,39 +467,110 @@ impl Opt {
...
+                    .await?;
+                // re-EHLO post TLS
+                caps = client.ehlo(&helo_domain).await?;
+            } else {
+                // "hard"
+                // if self.require_starttls { anyhow::bail!("STARTTLS not advertised by target"); }
+                if self.auth_username.is_some() {
+                    anyhow::bail!(
+                        "Server doesn't advertise STARTTLS; refusing to send AUTH without TLS"
+                    );
```
wez: This is another instance where we're enforcing a local policy. I think there are situations where you might want to proceed anyway, so we'd either need an additional flag to bypass this policy choice, or just remove this client side check.

I'd lean towards the latter: this is a testing tool and it feels to me like it should just do what you ask it to without too much friction.

crates/traffic-gen/src/main.rs
```diff
@@ -442,39 +467,110 @@ impl Opt {
...
+                                anyhow::bail!(
+                                    "failed to authenticate with AUTH PLAIN - {} ({})",
+                                    content,
+                                    code
+                                );
+                            }
+                        }
+                        other => {
+                            return Err(other).context("failed to authenticate with AUTH PLAIN")
+                        }
```
wez: This feels like quite a lot of code for this one error case; I'd just use the `?` to bubble the error up to the central error logging and keep the code smaller and easier to read.

crates/traffic-gen/src/main.rs
```diff
@@ -311,6 +326,16 @@ impl Opt {
         self.domain = adjusted_domains;
     }
 
+    fn auth_credentials(&self) -> anyhow::Result<Option<(String, String)>> {
+        match (self.auth_username.as_deref(), self.auth_password.as_deref()) {
+            (Some(user), Some(pass)) => Ok(Some((user.to_owned(), pass.to_owned()))),
```
wez: Feels redundant to copy the strings out here, only to dance around and turn them back into str references in the caller.

```suggestion
    fn auth_credentials(&self) -> anyhow::Result<Option<(&str, &str)>> {
        match (self.auth_username.as_deref(), self.auth_password.as_deref()) {
            (Some(user), Some(pass)) => Ok(Some((user, pass))),
```

crates/traffic-gen/src/main.rs
```diff
@@ -311,6 +326,16 @@ impl Opt {
...
+            (None, None) => Ok(None),
+            _ => Err(anyhow!(
+                "both --auth-username and --auth-password must be provided together"
```
wez: Since this condition is validated by clap, we could simplify this with the `expect` I suggested previously and remove the Result layer from the args, simplifying the code overall

wez (commented):
I think we can simplify this a bit more; there's more code than I think is strictly necessary here!

crates/traffic-gen/src/main.rs
cristian-porta: ok, forgot this block of code

crates/traffic-gen/src/main.rs
cristian-porta: Agreed and replaced the manual match with simple propagation plus a minimal context

crates/traffic-gen/src/main.rs
cristian-porta: removed the anyhow::Result layer and introduced expect

PR #440, opened 2025-10-29
Add support for proxy protocol

cai-n:
Add support for proxy protocol. If proxy protocol is enable, the header is parsed and ips in meta are updated accordingly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4687,6 +4687,8 @@ dependencies = [
  "chrono",
  "config",
  "criterion",
+ "k9",
+ "memchr",
  "mlua",
  "phf 0.13.1",
  "regex",
--- a/Makefile
+++ b/Makefile
@@ -32,7 +32,7 @@ test: build test-lua
 	RUST_BACKTRACE=1 cargo nextest run --no-fail-fast
 
 test-adhoc: build
-	cargo nextest run --no-fail-fast --no-capture -p integration-tests -- xfer
+	cargo nextest run --no-fail-fast --no-capture -p integration-tests -- disconnect_in_data_try_next
 
 test-kumod:
 	cargo nextest run --no-fail-fast -p kumod
--- a/crates/dns-resolver/src/lib.rs
+++ b/crates/dns-resolver/src/lib.rs
@@ -559,6 +559,8 @@ impl MailExchanger {
 #[derive(Debug, Clone, Serialize)]
 pub enum ResolvedMxAddresses {
     NullMx,
+    /// The list of addresses to which to connect, expressed
+    /// in LIFO order
     Addresses(Vec<ResolvedAddress>),
 }
 
--- a/crates/integration-tests/source.lua
+++ b/crates/integration-tests/source.lua
@@ -358,6 +358,13 @@ kumo.on('get_egress_path_config', function(domain, source_name, _site_name)
     tls_prefer_openssl = ((os.getenv 'KUMOD_PREFER_OPENSSL') and true)
       or false,
     max_recipients_per_batch = tonumber(MAX_RECIPIENTS_PER_BATCH),
+
+    -- Skip IPv6 addresses that come back for eg: localhost.
+    -- For the most part the integration tests don't care about this,
+    -- but the disconnect_XXX tests do make some assertions on the
+    -- ordering, and in particular, disconnect_terminate_ok will be
+    -- unhappy if the second address in its MX plan is unroutable IPv6.
+    skip_hosts = { '::/0' },
   }
 
   if os.getenv 'KUMOD_CLIENT_CERTIFICATE' then
--- a/crates/integration-tests/src/test/disconnect_in_data.rs
+++ b/crates/integration-tests/src/test/disconnect_in_data.rs
@@ -136,14 +136,14 @@ Response {
     assert_eq!(
         tf.peer_address,
         Some(ResolvedAddress {
-            name: "localhost-2".to_string(),
+            name: "localhost-1".to_string(),
             addr: daemon.sink.listener("smtp").into(),
         })
     );
     assert_eq!(
         delivery.peer_address,
         Some(ResolvedAddress {
-            name: "localhost-1".to_string(),
+            name: "localhost-2".to_string(),
             addr: daemon.sink.listener("smtp").into(),
         })
     );
--- a/crates/integration-tests/src/test/disconnect_terminate_ok.rs
+++ b/crates/integration-tests/src/test/disconnect_terminate_ok.rs
@@ -4,7 +4,16 @@ use std::time::Duration;
 
 /// This test verifies that a 421 disconnect doesn't cause collateral
 /// damage for the next message in the ready queue by making it
-/// experience a TransientFailure due to a connection error
+/// experience a TransientFailure due to a connection error.
+/// Please note that this test is sensitive to the composition
+/// of the connection plan: it assumes that the disconnect happens
+/// for the last viable candidate address.
+/// If IPv6 address(es) are returned for `localhost` and are present
+/// after the IPv4 address, and IPv6 is non-functional in the test
+/// environment, then this test will fail because that IPv6 address
+/// will (correctly!) produce a TransientFailure.
+/// We assume that skip_hosts is configured to skip IPv6 addresses
+/// in order for this test to pass in spite of that environment.
 #[tokio::test]
 async fn disconnect_terminate_ok() -> anyhow::Result<()> {
     let mut daemon = DaemonWithMaildir::start().await?;
--- a/crates/kumo-server-common/src/lib.rs
+++ b/crates/kumo-server-common/src/lib.rs
@@ -49,6 +49,7 @@ pub fn register(lua: &Lua) -> anyhow::Result<()> {
     }
 
     let kumo_mod = get_or_create_module(lua, "kumo")?;
+    kumo_mod.set("version", version_info::kumo_version())?;
 
     fn event_registrar_name(name: &str) -> String {
         format!("kumomta-event-registrars-{name}")
--- a/crates/kumod/src/smtp_dispatcher.rs
+++ b/crates/kumod/src/smtp_dispatcher.rs
@@ -170,6 +170,9 @@ impl SmtpDispatcher {
             for a in proto_config.mx_list.iter() {
                 a.resolve_into(&mut addresses).await?;
             }
+            // Note that ResolvedMxAddresses::Addresses is in LIFO
+            // order, and we have FIFO order.  Reverse it!
+            addresses.reverse();
             ResolvedMxAddresses::Addresses(addresses)
         };
 
--- a/crates/mod-smtp-response-normalize/Cargo.toml
+++ b/crates/mod-smtp-response-normalize/Cargo.toml
@@ -7,6 +7,7 @@ edition = "2024"
 anyhow.workspace = true
 chrono.workspace = true
 config = {path="../config"}
+memchr.workspace = true
 mlua = {workspace=true, features=["vendored", "lua54", "async", "send", "serialize"]}
 phf = "0.13"
 regex.workspace = true
@@ -14,6 +15,7 @@ uuid.workspace = true
 
 [dev-dependencies]
 criterion = {workspace=true}
+k9.workspace = true
 
 [[bench]]
 name = "normalizer"
--- a/crates/mod-smtp-response-normalize/benches/normalizer.rs
+++ b/crates/mod-smtp-response-normalize/benches/normalizer.rs
@@ -1,5 +1,5 @@
 use criterion::{Criterion, SamplingMode, Throughput, criterion_group, criterion_main};
-use mod_text_normalize::normalize;
+use mod_smtp_response_normalize::normalize;
 
 static GOOG: &str = "Your email has been rate limited because the From: header (RFC5322) in this message isn't aligned with either the authenticated SPF or DKIM organizational domain. To learn more about DMARC alignment, visit  https://support.google.com/a?p=dmarc-alignment  To learn more about Gmail requirements for bulk senders, visit  https://support.google.com/a?p=sender-guidelines. a640c23a62f3a-ab67626ed70si756442266b.465 - gsmtp";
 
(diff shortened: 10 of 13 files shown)


crates/kumod/src/smtp_server.rs
```diff
@@ -1510,6 +1519,13 @@ impl SmtpServerSession {
 
     #[instrument(skip(self))]
     async fn process(&mut self) -> anyhow::Result<()> {
+        if self.params.allow_proxy_protocol {
```
wez: I think this should be moved to happen after the `spool_started` check a little bit further below, so that we don't try to do material incoming connection processing until after we're ready.

I think we can do that without consequence; we don't need to signal back to the client that we're doing proxy stuff, so the various 421 responses should go through just fine?

crates/kumod/src/smtp_server.rs
```diff
@@ -2264,6 +2280,177 @@ impl SmtpServerSession {
...
+                        dest_addr.replace(addresses.destination_address.into());
+                        dest_port.replace(addresses.destination_port);
+                    }
+                    ppp::v1::Addresses::Tcp6(addresses) => {
+                        addr.replace(addresses.source_address.into());
+                        port.replace(addresses.source_port);
+                        dest_addr.replace(addresses.destination_address.into());
+                        dest_port.replace(addresses.destination_port);
+                    }
+                    _ => {}
```
wez: Let's explicitly spell out the `Unknown` variant here, rather than use the `_` wildcard.  It's just generally a bit more solid if an enum changes to add new variants in the future: I'd rather that the compiler tell us about them than to let us proceed in ignorance!

```suggestion
                    ppp::v1::Addresses::Unknown => {}
```

crates/kumod/src/smtp_server.rs
```diff
@@ -2264,6 +2280,177 @@ impl SmtpServerSession {
...
+                        dest_addr.replace(addresses.destination_address.into());
+                        dest_port.replace(addresses.destination_port);
+                    }
+                    ppp::v2::Addresses::IPv6(addresses) => {
+                        addr.replace(addresses.source_address.into());
+                        port.replace(addresses.source_port);
+                        dest_addr.replace(addresses.destination_address.into());
+                        dest_port.replace(addresses.destination_port);
+                    }
+                    _ => {}
```
wez:
```suggestion
                    ppp::v2::Addresses::Unspecified | ppp::v2::Addresses::Unix(_) => {}
```

crates/kumod/src/smtp_server.rs
```diff
@@ -2264,6 +2280,177 @@ impl SmtpServerSession {
...
+                }
+            }
+            HeaderResult::V1(Err(error)) => {
+                return Err(anyhow!("proxy protocol v1 parsing error: {error}"))
+            }
+            HeaderResult::V2(Err(error)) => {
+                return Err(anyhow!("proxy protocol v2 parsing error: {error}"))
+            }
+        }
+        self.read_buffer.clear();
```
wez: This may be a slightly moot comment, due to how SMTP is phrased, but this unilateral buffer clear has the potential to discard unread data.

Let's say that a slightly non-conforming client connects and immediate sends `EHLO domain` after the proxy protocol header, without waiting for the server to send the banner.  If the proxy header is relatively short, `read_buffer` now contains a fragment of that `EHLO` and when we reach this point the buffer clear will discard it.

Subsequently, when we start processing lines for real, we'll consider that first command to be nonsense.

What we should do here is consume only the bytes that were part of the recognized proxy header.  The ppp types make that look a bit awkward, but I think that you can read the `.len()` from either:

* https://docs.rs/ppp/latest/ppp/v1/struct.Header.html#structfield.header
* https://docs.rs/ppp/latest/ppp/v2/struct.Header.html#structfield.header

Armed with that consumed length, we can then `drain` out that many bytes from the read buffer, something like: `self.read_buffer.drain(0..consumed_len)`.

```suggestion
        self.read_buffer.drain(0..consumed_bytes);
```

crates/kumod/src/smtp_server.rs
```diff
@@ -2264,6 +2280,177 @@ impl SmtpServerSession {
...
+            let header = HeaderResult::parse(&self.read_buffer);
+            if header.is_complete() {
+                break header;
+            }
+        };
+
+        let mut addr: Option<IpAddr> = None;
+        let mut port: Option<u16> = None;
+        let mut dest_addr: Option<IpAddr> = None;
+        let mut dest_port: Option<u16> = None;
```
wez: See comment about clearing the buffer below

```suggestion
        let mut dest_port: Option<u16> = None;
        let mut consumed_bytes = 0;
```

crates/kumod/src/smtp_server.rs
```diff
@@ -2264,6 +2280,177 @@ impl SmtpServerSession {
...
+        match header {
+            HeaderResult::V1(Ok(header)) => {
+                let addresses = header.addresses;
```
wez, line 2349:
```suggestion
                consumed_bytes = header.header.len();
                let addresses = header.addresses;
```

crates/kumod/src/smtp_server.rs
```diff
@@ -2264,6 +2280,177 @@ impl SmtpServerSession {
...
+                        addr.replace(addresses.source_address.into());
+                        port.replace(addresses.source_port);
+                        dest_addr.replace(addresses.destination_address.into());
+                        dest_port.replace(addresses.destination_port);
+                    }
+                    _ => {}
+                }
+            }
+            HeaderResult::V2(Ok(header)) => {
+                let addresses = header.addresses;
```
wez, line 2368:
```suggestion
                consumed_bytes = header.header.len();
                let addresses = header.addresses;
```

crates/kumod/src/smtp_server.rs
wez: Let's roll these into `SocketAddr`s so that we have fewer pieces to reason about after we've read the header

```suggestion
        let mut addr: Option<SocketAddr> = None;
        let mut dest_addr: Option<SocketAddr> = None;
```

crates/kumod/src/smtp_server.rs
```diff
@@ -2264,6 +2280,177 @@ impl SmtpServerSession {
...
+        let mut port: Option<u16> = None;
+        let mut dest_addr: Option<IpAddr> = None;
+        let mut dest_port: Option<u16> = None;
+        match header {
+            HeaderResult::V1(Ok(header)) => {
+                let addresses = header.addresses;
+                match addresses {
+                    ppp::v1::Addresses::Tcp4(addresses) => {
+                        addr.replace(addresses.source_address.into());
+                        port.replace(addresses.source_port);
```
wez: I think you can do something like this to get a `SocketAddr`:

```suggestion
                        addr.replace((addresses.source_address, addresses.source_port).into());
```

https://doc.rust-lang.org/beta/std/net/enum.SocketAddr.html#impl-From%3C(I,+u16)%3E-for-SocketAddr

crates/kumod/src/smtp_server.rs
```diff
@@ -2264,6 +2280,177 @@ impl SmtpServerSession {
...
+            HeaderResult::V2(Err(error)) => {
+                return Err(anyhow!("proxy protocol v2 parsing error: {error}"))
+            }
+        }
+        self.read_buffer.clear();
+
+        let old_peer = self.peer_address.clone();
+        let old_via = self.my_address.clone();
+
+        self.peer_address = (addr.unwrap(), port.unwrap()).into();
```
wez: These unwraps are undesirable because an attacker could send a proxy protocol packet with an Unknown or Unspecified address to trigger a panic and break the service.

We need to handle this situation gracefully!  Using `SocketAddr` (as indicated above) to have fewer pieces to consider here makes it a bit less verbose to conditionally update the pieces here using `if let Some(..) = ..`

You can see how that is handled in the xclient logic; the same approach can be used here.

crates/kumod/src/smtp_server.rs
```diff
@@ -2264,6 +2280,177 @@ impl SmtpServerSession {
...
+            }
+            Err(rej) => {
+                tracing::warn!("{:?}", rej);
+                return Err(anyhow!(
+                    "proxy protocol caused dynamic parameter rejection: {} {}",
+                    rej.code,
+                    rej.message
+                ));
+            }
+        }
```
wez: I think we can extract this same logic from `process_xclient` into a `re_evaluate_listener_parameters` method and call the same logic from both and avoid the  copypasta.
I think the main difference here is how errors are reported, but I think that the way that is done in `process_xclient` can also be used here unmodified with the proxy protocol?

crates/kumod/src/smtp_server.rs
```diff
@@ -2264,6 +2280,177 @@ impl SmtpServerSession {
         }
     }
 
+    #[instrument(skip(self))]
+    async fn process_proxy_protocol(&mut self) -> anyhow::Result<()> {
+        if self.socket.is_none() {
+            return Err(anyhow!("socket is none"));
```
wez: You can do this more succinctly; please adjust this and other similar constructions below!

```suggestion
            anyhow::bail!("socket is none");
```

crates/kumod/src/smtp_server.rs
```diff
@@ -639,6 +645,9 @@ pub struct GenericEsmtpListenerParams {
 
     #[serde(default)]
     allow_xclient: Option<bool>,
+
+    #[serde(default)]
+    allow_proxy_protocol: Option<bool>,
```
wez: From a naming perspective, I think this is more of a `require_proxy_protocol` than an `allow_proxy_protocol`, because when it is enabled we require a proxy protocol header to be unilaterally sent by the client before we can send the banner.  That semantic difference alone means that we cannot automagically proceed for clients that are not sending such a header.

We'll need to indicate this in the docs for this option and recommend that the user deploy two different listener ports if they want both proxy enabled and regular listeners.

crates/kumod/src/smtp_server.rs
```diff
@@ -2264,6 +2280,177 @@ impl SmtpServerSession {
...
+                    }
+                    _ => {}
+                }
+            }
+            HeaderResult::V2(Ok(header)) => {
+                let addresses = header.addresses;
+                match addresses {
+                    ppp::v2::Addresses::IPv4(addresses) => {
+                        addr.replace(addresses.source_address.into());
+                        port.replace(addresses.source_port);
```
wez: BTW, if you're setting up the integration test using our existing `ha_proxy_source_address` stuff, we set the `source_port = 0` to indicate that we don't care what source port the outgoing proxy uses for us.  I don't think that matters too much here, it will just look weird at first.

wez (commented):
Thanks for this!
I think this broadly looks fine, my comments are mostly around the edge cases.
There's also a big question around testing!

Perhaps the easiest path through testing is:

1. Take `crates/integration-tests/src/test/xclient.rs` as a base
2. introduce a `KUMOD_TEST_REQUIRE_PROXY_PROTOCOL` env var to `crates/integration-tests/maildir-sink.lua` that is used to set the `allow_proxy_protocol` option (but renamed to `require_proxy_protocol` per comments below) in the listener.  That will allow the maildir sink to optionally be used in proxy mode for integration tests that enable it.
3. Set that env var when starting the sink:

```rust
let mut daemon = DaemonWithMaildirOptions::new()
    .env("KUMOD_TEST_REQUIRE_PROXY_PROTOCOL", "1")
    .start()
    .await
    .context("DaemonWithMaildir::start")?;
```

4. Either introduce some more env vars to `crates/integration-tests/source.lua` that allow turning on proxy support in the egress source (via `ha_proxy_source_address`), or otherwise make a new `proxy-source.lua` that is focused on just this use case, which is likely less confusing and complex overall.  You can use the `DaemonWithMaildirOptions::policy_file` method to use an alternative source policy.

With that setup, the source will use the proxy protocol to the sink and set its source address; you can then make assertions against what is observed in the *sink* session rather than the *source* session.  If you're copying from the xclient test note that it starts a tracer against the source, but for this setup, you'd need the tracer to be against the sink.

crates/integration-tests/src/test/proxy_protocol.rs
```diff
@@ -0,0 +1,57 @@
...
+    anyhow::ensure!(response.code == 250);
+
+    daemon
+        .wait_for_maildir_count(1, Duration::from_secs(10))
+        .await;
+
+    let trace_events = tracer.stop().await?;
+    eprintln!("{trace_events:#?}");
+
+    let re_evaluated_params = trace_events.iter().any(|event| {
```
wez, line 45: I recently updated the xclient tests to use `tracer.wait_for` to add an explicit wait for this event being reported. Without it, there is a potential race in the integration test that can cause test flakeyness.  Please update this one to do the same sort of thing!

crates/integration-tests/src/kumod.rs
```diff
@@ -268,12 +268,20 @@ impl DaemonWithMaildir {
...
+    }
+
+    pub async fn trace_sink(&self) -> anyhow::Result<ServerTracer> {
+        return Ok(self.run_trace(&self.sink).await?);
+    }
+
     /// Run a server trace and capture the records until the trace is stopped,
     /// at which point it will return the set of records.
-    pub async fn trace_server(&self) -> anyhow::Result<ServerTracer> {
+    pub async fn run_trace(&self, target: &KumoDaemon) -> anyhow::Result<ServerTracer> {
```
wez: Ah, should we perhaps migrate this method to live on `KumoDaemon`, then `trace_server` and `trace_sink` would then call through via `self.source.trace_server()` and `self.sink.trace_server()` respectively?

That would avoid the slightly awkward `target: &KumoDaemon` parameter here.

crates/integration-tests/proxy-source.lua
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  kumo.start_http_listener {
+    listen = '127.0.0.1:0',
+  }
+
+  kumo.configure_local_logs {
+    log_dir = TEST_DIR .. '/logs',
+    max_segment_duration = '1s',
+    headers = { 'X-*', 'Y-*', 'Subject' },
```
wez: Can probably eliminate this for this test?

```suggestion
```

crates/integration-tests/proxy-source.lua
```diff
@@ -0,0 +1,92 @@
...
+    relay_to = true,
+    log_oob = true,
+    log_arf = true,
+  }
+end)
+
+kumo.on('smtp_server_message_received', function(msg)
+  local result = msg:import_scheduling_header 'X-Schedule'
+  kumo.log_info('schedule result', kumo.serde.json_encode(result))
+end)
```
wez: Don't need this either?

```suggestion
```

wez (commented):
Almost there! Thanks!

Solvik:
thank you @cai-n !

PR #441, opened 2025-10-30
Allow temp/perm errors in dmarc

sophiajt:
This PR does a few things:

- We now bubble up different kinds of DNS errors as either temp or perm errors to align more closely with RFC 7001
- We also fill the AuthenticationResult differently. Now, the policy information is held in the `props` field of the AuthenticationResult, and the result is normalised.

### DIFF
--- a/crates/kumod/src/dmarc.rs
+++ b/crates/kumod/src/dmarc.rs
@@ -109,7 +109,7 @@ pub fn register<'lua>(lua: &'lua Lua) -> anyhow::Result<()> {
                     disp @ Disposition::Quarantine |
                     disp @ Disposition::Reject => {
                         let mut props = BTreeMap::default();
-                        props.insert("policy", disp.to_string().to_ascii_lowercase());
+                        props.insert("policy.published-domain-policy".to_string(), disp.to_string().to_ascii_lowercase());
                         Ok(lua.to_value_with(
                             &CheckHostOutput {
                                 disposition: result.result.clone(),


crates/kumo-dmarc/src/lib.rs
```diff
@@ -91,41 +114,50 @@ impl<'a> DmarcContext<'a> {
...
-                {
-                    for record in records {
-                        return record
-                            .evaluate(self, SenderDomainAlignment::OrganizationalDomain)
-                            .await;
+                        {
+                            DmarcRecordResolution::TempError => {
+                                return DispositionWithContext {
+                                    result: Disposition::TempError,
+                                    context: "DNS records could not be resolved".to_string(),
```
wez: please include the generated _dmarc.org_domain string in this context, as that would be super helpful to the end user operator!

crates/kumo-dmarc/src/lib.rs
```diff
@@ -91,41 +114,50 @@ impl<'a> DmarcContext<'a> {
...
+                                }
+                            }
+                            DmarcRecordResolution::PermError => {
+                                return DispositionWithContext {
+                                    result: Disposition::PermError,
+                                    context: "No DMARC found in DNS record".to_string(),
```
wez: include the domain here too.  Can this be made consistent with the "no DMARC records found for {}" message below? It feels like they are logically the same case, just with the from_domain vs. the org domin

crates/kumod/src/dmarc.rs
```diff
@@ -87,19 +87,44 @@ pub fn register<'lua>(lua: &'lua Lua) -> anyhow::Result<()> {
...
+                        props.insert("policy", disp.to_string().to_ascii_lowercase());
+                        Ok(lua.to_value_with(
+                            &CheckHostOutput {
+                                disposition: result.result.clone(),
+                                result: AuthenticationResult {
+                                    method: "dmarc".to_string(),
+                                    method_version: None,
+                                    result: "fail".to_string(),
+                                    reason: Some(result.context),
+                                    props: BTreeMap::default(),
```
wez:
```suggestion
                                    props,
```

crates/kumod/src/dmarc.rs
```diff
@@ -87,19 +87,44 @@ pub fn register<'lua>(lua: &'lua Lua) -> anyhow::Result<()> {
...
+                                },
+                            },
+                            serialize_options(),
+                        ))
+                    }
+                    disp @ Disposition::Quarantine |
+                    disp @ Disposition::Reject => {
+                        let mut props = BTreeMap::default();
+                        props.insert("policy", disp.to_string().to_ascii_lowercase());
```
wez: having looked through the RFC in more detail, the policy doesn't appear to be a thing that gets pulled through into `props` for dmarc; there's no defined property name for it.  I think that it is fine for it to be returned here via `CheckHostOutput::disposition` (although we should probably rename `CheckHostOutput` to something like `DmarcPolicyResult` or something, because `CheckHostOutput` is copypasta from the spf bindings!

sophiajt: per our chat, I've switched this to use the `policy.published-domain-policy` prop instead of the `policy` prop, as there didn't seem to be an allowance in the spec to extend the authentication result registry with quarantine and reject.

PR #445, opened 2025-10-31
Add Hornetsecurity functionality to policy-extras

tommairs:
This adds functionality to use Hornetsecurity's spam and malware protection as documented [here](https://docs.kumomta.com/userguide/integrations/hornetsecurity/).

### DIFF
--- a/assets/policy-extras/hornet.lua
+++ b/assets/policy-extras/hornet.lua
@@ -1,0 +1,175 @@
+--[[ hornet.lua ]]--
+-- Interface functions for Hornet Email Protection integration
+
+local name = "hornet"
+local mod = {}
+local kumo = require 'kumo'
+local utils = require 'policy-extras.policy_utils'
+
+
+---------------------------------------------------------------------------------------------
+--[[ hornet:connect ]]--
+-- inputs: 
+--   vhost - IP or hostname of the hornet server
+--   vparams - array of values incluing port, tls flag, maximum zise to scan, timeout.
+--            IE: {['port']=800,['timeout']=30,['use_tls']=true,['maxsize']=2097152}
+-- returns: array of hornet server connection values, AKA "hornet connection array"
+--
+function mod:connect(vhost,vparams)
+  return {
+    host = vhost or '127.0.0.1',
+    port = vparams.port or 8080,
+    timeout = vparams.timeout or 10,
+    use_tls = vparams.use_tls or false,
+  }
+end
+
+local function make_url(params, endpoint)
+  local vhost = string.format("%s:%d", params.host, params.port)
+  local scheme = 'http'
+  if params.use_tls then
+    scheme = 'https'
+  end
+		
+  return string.format('%s://%s/%s', scheme, vhost, endpoint)
+end
+
+
+
+
+---------------------------------------------------------------------------------------------
+--[[ hornet:scan ]]--
+-- inputs: 
+--   params" = hornet connection array, 
+--   extraparams = array of:
+--     addheaders = true/false   -- Add result headers directly to the message?
+--     mode = string            -- make this smtpout (default) to indicate outbound scanning, smtpin for inbound
+--     x_sanitize = true/false  -- make this true to have the engine remove dangerous attachments
+--   msg = message object
+-- returns:
+--   state and score (int) - The spam state and spam score returned for the message. For some states, the
+--      score may provide additional information about the category the message belongs to.
+--      For a complete list of states and scores, please refer to Evaluating State, Score and Verdict.
+
+--   verdict (str) - The message verdict (with no further detail), allowing you to trigger message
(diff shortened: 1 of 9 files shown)


assets/policy-extras/hornet.lua
```diff
@@ -0,0 +1,172 @@
...
+---------------------------------------------------------------------------------------------
+--[[ hornet:connect ]]
+--
+-- inputs:
+--   vhost - IP or hostname of the hornet server
+--   vparams - array of values incluing port, tls flag, maximum zise to scan, timeout.
+--            IE: {['port']=800,['timeout']=30,['use_tls']=true,['maxsize']=2097152}
+-- returns: array of hornet server connection values, AKA "hornet connection array"
+--
+function mod:connect(vhost, vparams)
```
wez, line 19: This module functions should be functions rather than methods, because they don't store any state in "self":

```suggestion
function mod.connect(vhost, vparams)
```

Having just done a pass through the docs, I can see that they have sections like this:

```lua
local hornethost =
    hornet:connect('172.31.17.26', { port = '8080', use_tls = 'false' })

  if not hornethost then
    print 'No connection to Hornetsecurity host'
  else
```

the `if not hornesthost` is impossible to trigger because the connect method unconditionally just returns an object, so we should update the docs to simplify things by removing the `if` bit.

Since this function isn't actually connecting to anything, I would advocate rolling it into the scan function below, making things simpler overall

assets/policy-extras/hornet.lua
```diff
@@ -0,0 +1,172 @@
...
+--  Events (array) - The list of current events that matched the message. Typically ["COVID-19", "EMOTET", …].
+
+--  elapsed (str) - Time it took for the server (in ms) to process the message and return the verdict
+
+function mod:scan(params, extraparams, msg)
+  if not params then
+    print 'No connection table provided'
+    return false
+  end
+
```
wez, line 93: I would collapse the `mod:connect` function here like this:

```suggestion
local params = {
   host = params.host or '127.0.0.1',
   port = params.port or 8080,
   timeout = params.timeout or 10,
   use_tls = params.use_tls or false,
}
```

assets/policy-extras/hornet.lua
```diff
@@ -0,0 +1,172 @@
...
+  if not msg then
+    return false
+  end
+
+  -- Flatten message data to a string
+  local msgdata = msg:get_data()
+
+  -- decompose msg here into vars to pass
+  local modifymsg = extraparams.addheaders or false -- Add result headers directly to the message?
```
wez, line 102: I'd also collapse `extraparams` into `params` just to keep things simpler with a single table style parameter to the scan function.

assets/policy-extras/hornet.lua
```diff
@@ -0,0 +1,172 @@
...
+    ['X-Helo'] = x_helo,
+    ['X-Mailfrom'] = x_mailfrom,
+    ['X-Rcptto'] = x_rcptto,
+    ['X-Sanitize'] = x_sanitize,
+    ['X-Params'] = 'mode=' .. direction_mode,
+  }
+
+  -- catch usecase if a raw number is provided for API timeout
+  if type(duration) == 'number' then
+    duration = '' .. tostring(duration) .. 's'
```
wez, line 132:
```suggestion
    duration = tostring(duration) .. 's'
```

assets/policy-extras/hornet.lua
```diff
@@ -0,0 +1,172 @@
...
+  -- If requested, add headers to message to describe the scan results
+  if modifymsg == true then
+    scanresult_json = kumo.serde.json_parse(scanresult)
+    for skey, svalue in pairs(scanresult_json) do
+      skey = 'X-Hornet-' .. skey
+      msg:append_header(skey, svalue)
+    end
+  end
+
+  return scanresult
```
wez, line 169: There's no need to form an SMTP style disposition just to crack it apart again; you can directly use the response information.

Unconditionally printing status in these helpers for per-message actions will generate a lot of noise in the diagnostic logs, so we should avoid that.  In addition, a print for an error with no error being propagated out is a silent error until someone looks at the diagnostic logs, and should be avoided as well.

I'd suggest phrasing the logic this way, which causes failures to talk to the API to bubble up as 421 technical difficulties errors in the incoming session.  There are other ways to handle this, but we should probably have a design session to talk about the higher level plans for this sort of integration.

```suggestion
  local scanresult = response:text()
  if response:status_code() ~= 200 then
        -- this will produce a 421 technical difficulties response
        error('failed to query hornetsecurity: ' .. scanresult)
  end 
  
  -- invalid json will produce a 421 technical difficulties response  
  scanresult_json = kumo.serde.json_parse(scanresult)

  if modifymsg then
    for skey, svalue in pairs(scanresult_json) do
      msg:append_header('X-Hornet-' .. skey, svalue)
    end
  end

  return scanresult_json
```

PR #448, opened 2025-11-06
Add a dmarc lua test, fix API

sophiajt:
This PR does a few different, related changes:

- it adds a dmarc test into the new mail_auth.lua framework
- it updates the dmarc API
  - to simplify the API by removing unnecessary parameters
  - to change the name of the call to align better with other auth systems
  - and finally, to bring it inline with the new resolver API


wez:
Looks great, thanks!

PR #449, opened 2025-11-07
Integrate NATS client

tbshrst:
Hello folks,

I'm proposing a NATS integration for kumo to publish messages to NATS servers with JetStream enabled. I think this is a good use case, since NATS is pretty popular, especially in the cloud-native environment. It supports publishing messages and authenticates itself via password or token. Unlike mod-kafka, a builder pattern is used in Rust, so the Lua input must be translated. The proposal employs the [async-nats](https://crates.io/crates/async-nats) crate.

The interface looks like this:

```lua
local nats = kumo.nats.connect {
	servers = {"127.0.0.1:4222"},
	name = "nats-client",
	auth = {
		username = "user",
		password = "somepassword",
	}
}

nats:publish {
	subject = "my.subject",
	payload = "hello from kumomta!",
	headers = {
		["Nats-Msg-Id"] = "unique-key",
	},
	await_ack = true, -- optional
}
nats:close()
```

I haven't written any documentation yet, as I wanted to wait for your response first.

### DIFF
--- a/crates/mod-nats/src/lib.rs
+++ b/crates/mod-nats/src/lib.rs
@@ -11,7 +11,7 @@ use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
 
-// https://docs.rs/async-nats/0.45.0/src/async_nats/options.rs.html#44
+// https://docs.rs/async-nats/0.46.0/src/async_nats/options.rs.html#43
 #[derive(Debug, Deserialize)]
 struct Config {
     servers: Vec<String>,
@@ -45,7 +45,7 @@ struct ConfigAuth {
     token: Option<KeySource>,
 }
 
-// https://docs.rs/async-nats/0.45.0/src/async_nats/auth.rs.html#4
+// https://docs.rs/async-nats/0.46.0/src/async_nats/auth.rs.html#4
 #[derive(Debug, Deserialize)]
 struct Auth {
     username: Option<String>,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -388,16 +388,14 @@ dependencies = [
 
 [[package]]
 name = "async-nats"
-version = "0.45.0"
+version = "0.46.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "86dde77d8a733a9dbaf865a9eb65c72e09c88f3d14d3dd0d2aecf511920ee4fe"
+checksum = "df5af9ebfb0a14481d3eaf6101e6391261e4f30d25b26a7635ade8a39482ded0"
 dependencies = [
  "base64 0.22.1",
  "bytes",
  "futures-util",
  "memchr",
- "nkeys",
- "nuid",
  "once_cell",
  "pin-project",
  "portable-atomic",
@@ -405,7 +403,7 @@ dependencies = [
  "regex",
  "ring",
  "rustls-native-certs 0.7.3",
- "rustls-pemfile",
+ "rustls-pki-types",
  "rustls-webpki 0.102.8",
  "serde",
  "serde_json",
@@ -2059,7 +2057,6 @@ dependencies = [
  "ed25519",
  "serde",
  "sha2",
- "signature",
  "subtle",
  "zeroize",
 ]
@@ -5080,21 +5077,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "nkeys"
-version = "0.4.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "879011babc47a1c7fdf5a935ae3cfe94f34645ca0cac1c7f6424b36fc743d1bf"
-dependencies = [
- "data-encoding",
- "ed25519",
- "ed25519-dalek",
- "getrandom 0.2.17",
- "log",
- "rand 0.8.5",
- "signatory",
-]
-
-[[package]]
 name = "nom"
 version = "5.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5175,15 +5157,6 @@ checksum = "bf33f538733d1a5a3494b836ba913207f14d9d4a1d3cd67030c5061bdd2cac85"
 dependencies = [
  "memchr",
  "unicode-segmentation",
-]
-
-[[package]]
-name = "nuid"
-version = "0.5.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fc895af95856f929163a0aa20c26a78d26bfdc839f51b9d5aa7a5b79e52b7e83"
-dependencies = [
- "rand 0.8.5",
 ]
 
 [[package]]
@@ -7310,24 +7283,11 @@ dependencies = [
 ]
 
 [[package]]
-name = "signatory"
-version = "0.27.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c1e303f8205714074f6068773f0e29527e0453937fe837c9717d066635b65f31"
-dependencies = [
- "pkcs8",
- "rand_core 0.6.4",
- "signature",
- "zeroize",
-]
-
-[[package]]
 name = "signature"
 version = "2.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "77549399552de45a898a580c1b41d445bf730df867cc44e6c0233bbc4b8329de"
 dependencies = [
- "digest",
  "rand_core 0.6.4",
 ]
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ ahash = "0.8.11"
 amqprs = {version="2.0", features=["tls", "traces"]}
 anyhow = "1.0"
 arc-swap = "1.6"
-async-nats = "0.45.0"
+async-nats = {version = "0.46", default-features = false, features = ["jetstream", "ring", "server_2_10", "server_2_11", "server_2_12"]}
 async-stream = "0.3"
 async-trait = "0.1"
 aws-lc-rs = "1.13.1"
--- a/crates/mod-nats/Cargo.toml
+++ b/crates/mod-nats/Cargo.toml
@@ -5,12 +5,7 @@ edition = "2021"
 
 [dependencies]
 anyhow.workspace = true
-async-nats = { workspace = true, features = [
-    "server_2_10",
-    "server_2_11",
-    "server_2_12",
-    "ring",
-] }
+async-nats = { workspace = true }
 config = { path = "../config" }
 data-loader = { path = "../data-loader", default-features = false }
 duration-serde = { path = "../duration-serde" }


crates/mod-nats/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+// https://docs.rs/async-nats/0.45.0/src/async_nats/auth.rs.html#4
+#[derive(Debug, Deserialize)]
+struct Auth {
+    // jwt: Option<String>,
+    // nkey: Option<String>,
+    // signature_callback: Option<CallbackArg1<String, Result<String, AuthError>>>,
+    // signature: Option<Vec<u8>>,
+    username: Option<String>,
+    password: Option<String>,
+    token: Option<String>,
```
wez, line 53: Where we're dealing with the credential material directly (eg: the password string, rather than passing down the path to a key file to an external crate), prefer to use `KeySource` to abstract away where the credentials come from.  This allows the user to employ a variety of different data stores as well as inline bytes if they wish.

the Rust side is documented here:
https://docs.kumomta.com/rustapi/data_loader/enum.KeySource.html

You will use https://docs.kumomta.com/rustapi/data_loader/enum.KeySource.html#method.get to load the bytes out of the KeySource.

The lua side of KeySource is documented here:
https://docs.kumomta.com/reference/keysource/

I'd suggest using KeySource for the username as well as the password and token as that allows completely separating the credential information out from the lua policy source code, if desired.

crates/mod-nats/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+// https://docs.rs/async-nats/0.45.0/src/async_nats/options.rs.html#44
+#[derive(Debug, Deserialize)]
+struct Config {
+    servers: Vec<String>,
+
+    name: Option<String>,
+    no_echo: Option<bool>,
+    max_reconnects: Option<usize>,
+    #[serde(default)]
+    #[serde(with = "duration_serde")]
```
wez:
```suggestion
    #[serde(default, with = "duration_serde")]
```

crates/mod-nats/src/lib.rs
```diff
@@ -0,0 +1,196 @@
...
+    client_cert: Option<PathBuf>,
+    client_key: Option<PathBuf>,
+    // tls_client_config: Option<rustls::ClientConfig>,
+    ping_interval: Option<Duration>,
+    // subscription_capacity: Option<usize>,
+    sender_capacity: Option<usize>,
+    // event_callback: Option<CallbackArg1<Event, ()>>,
+    inbox_prefix: Option<String>,
+    #[serde(default)]
+    #[serde(with = "duration_serde")]
```
wez:
```suggestion
    #[serde(default, with = "duration_serde")]
```

crates/mod-nats/src/lib.rs
tbshrst: I implemented credential transfer using Keysource. Since I was free to choose the format, I opted for the TOML format that was already in use, see Tests.

wez: Just a quick comment for the moment, as I don't have time for a full review right now, but: we typically have one KeySource per item, rather than creating a data format to bundle multiple together.  In other words, where you previously had a `String` for each credential-like item (one for username, one for password etc.), you'd replace that with a `KeySource` for each one, resolving each separately.

That gives the end user the most flexibility in how they configure and deploy their credential management.

tbshrst: Gotcha! Would you also prefer caching the credentials?  Since connection establishment this is usually a one-time operation during init, I left it out so far.

crates/integration-tests/src/nats.rs
```diff
@@ -0,0 +1,105 @@
...
+use testcontainers_modules::testcontainers::core::ContainerPort;
+use testcontainers_modules::testcontainers::runners::AsyncRunner;
+use testcontainers_modules::testcontainers::ImageExt;
+
+const SUBJECT: &str = "events";
+const USERNAME: &str = "user";
+const PASSWORD: &str = "userpassword";
+const PORT: u16 = 4222;
+
+lazy_static! {
```
wez: `lazy_static!` should be considered deprecated in favor of `std::sync::LazyLock`; please don't introduce new code using lazy_static.

crates/integration-tests/src/nats.rs
```diff
@@ -0,0 +1,105 @@
...
+    static ref ADDR: String = format!("localhost:{PORT}");
+    static ref NATS_CONFIG: Value = json!(
```
wez: FWIW, I don't think this needs to be static; I think you can make this a simple function that returns this json object, then call that when you are setting up the nats container.  Then you can replace the globals here with simple variables, which will make this code easier to maintain in the long run.

crates/mod-nats/src/lib.rs
wez: No need to cache inside this struct.  That should be left to something higher level; eg: the custom lua handler will keep the connection alive.  That should be good enough for most purposes, unless for some reason the queue parameters are tight (eg: max of 1 message per connection).

crates/mod-nats/src/lib.rs
```diff
@@ -0,0 +1,226 @@
+use async_nats::jetstream::{self, Context};
+use async_nats::rustls::lock::Mutex;
```
wez: Can we switch this wrapper to parking-lot's mutex? It is "faster" and doesn't have the `.unwrap()` wart:

```suggestion
use parking_lot::Mutex;
```

crates/mod-nats/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+#[derive(Clone)]
+struct Client {
+    context: Arc<Mutex<Option<Arc<Context>>>>,
+}
+
+impl Client {
+    fn get_context(&self) -> mlua::Result<Arc<Context>> {
+        self.context
+            .lock()
+            .unwrap()
```
wez: with parking lot's mutex, this goes away

```suggestion
```

crates/mod-nats/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+
+            if message.await_ack {
+                ack_fut.await.map_err(|err| any_err(err))?;
+            }
+
+            Ok(())
+        });
+
+        methods.add_async_method("close", |_lua, this, _: ()| async move {
+            this.context.lock().unwrap().take();
```
wez:
```suggestion
            this.context.lock().take();
```

crates/integration-tests/nats.lua
```diff
@@ -0,0 +1,62 @@
...
+    },
+    await_ack = true,
+  }
+
+  producer:close()
+
+  producer:publish {
+    subject = SUBJECT,
+    payload = 'not sent because already closed',
+  }
```
wez: I feel like this should be an error, and that we should have a test assertion for it, but we can't do that in this context

crates/integration-tests/nats.lua
```diff
@@ -0,0 +1,62 @@
...
+
+  producer:publish {
+    subject = SUBJECT,
+    payload = 'payload-3',
+    headers = {
+      ['Nats-Msg-Id'] = 'unique-key-3',
+    },
+    await_ack = true,
+  }
```
wez: Doing all this inside the init handler makes it very difficult to make test assertions, and also doesn't really validate that things work in the end context that you want to deploy, which is to route messages from the normal message flow.

Please re-structure this test so that it is doing the same sorts of things you describe in the docs you wrote out; set up a delivery handler, then inject messages via SMTP and have your integration test make assertions on the success/failure of the SMTP rejection, as well as on observing the appropriate messages come out the other end of the NATS instance

crates/mod-nats/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+    // event_callback: Option<CallbackArg1<Event, ()>>,
+    inbox_prefix: Option<String>,
+    #[serde(default, with = "duration_serde")]
+    request_timeout: Option<Duration>,
+    retry_on_initial_connect: Option<bool>,
+    ignore_discovered_servers: Option<bool>,
+    retain_servers_order: Option<bool>,
+    // read_buffer_capacity: Option<u16>,
+    // reconnect_delay_callback: Box<dyn Fn(usize) -> Duration + Send + Sync + 'static>,
+    // auth_callback: Option<CallbackArg1<Vec<u8>, Result<Auth, AuthError>>>,
```
wez: Please remove the commented/dead lines to keep the code looking a bit cleaner

```suggestion
    ping_interval: Option<Duration>,
    client_capacity: Option<usize>,
    inbox_prefix: Option<String>,
    #[serde(default, with = "duration_serde")]
    request_timeout: Option<Duration>,
    retry_on_initial_connect: Option<bool>,
    ignore_discovered_servers: Option<bool>,
    retain_servers_order: Option<bool>,
```

crates/mod-nats/src/lib.rs
```diff
@@ -0,0 +1,226 @@
...
+    token: Option<KeySource>,
+}
+
+// https://docs.rs/async-nats/0.45.0/src/async_nats/auth.rs.html#4
+#[derive(Debug, Deserialize)]
+struct Auth {
+    // jwt: Option<String>,
+    // nkey: Option<String>,
+    // signature_callback: Option<CallbackArg1<String, Result<String, AuthError>>>,
+    // signature: Option<Vec<u8>>,
```
wez:
```suggestion
```

docs/reference/kumo.nats/connect.md
```diff
@@ -0,0 +1,55 @@
+# kumo.nats.connect
+
```
wez, line 3: Let's put the version in here:

```suggestion
{{since('dev')}}
```

docs/reference/kumo.nats/publish.md
```diff
@@ -0,0 +1,19 @@
+### client:publish({PARAMS})
+
```
wez, line 3:
```suggestion
{{since('dev')}}
```

docs/reference/kumo.nats/close.md
```diff
@@ -0,0 +1,7 @@
+### client:close()
+
```
wez, line 3:
```suggestion
{{since('dev')}}
```

docs/userguide/policy/nats.md
```diff
@@ -0,0 +1,59 @@
+# Routing Messages via NATS
+
```
wez, line 3:
```suggestion
{{since('dev')}}
```

crates/integration-tests/nats.lua
tbshrst: done

crates/integration-tests/nats.lua
tbshrst: I integrated the tests into `sources.lua` and dropped this since, since it's not that worth testing from my perspective

wez:
Thanks for this; I tidied up a few things in the implementation and the docs and pushed it!

PR #450, opened 2025-11-07
add some bounce classifier rules and divide other rules into appropriate labels

danielsmithsd:
I have access to a lot of bounce history along with prior classifications outcomes.  After doing some analysis and reclassifying those bounce records, I made some tweaks to the labels and added a few of the more obvious bounce messages into the community file.

I realize some of the classifications can be subjective, so open to moving/removing/arguing about them as well.


wez:
From a technical perspective, this looks fine. Let's have @MHillyer review how things have been (re)classified!

MHillyer:
Thanks!

PR #457, opened 2025-11-22
Add AWS SigV4 signature module

AdityaAudi:
### Motivation
KumoMTA users often need to call AWS services (for example, Firehose, Lambda, S3, or SQS) directly from Lua. Without a shared implementation, each policy must re‑implement AWS SigV4 signing, leading to duplicated logic, inconsistent behavior, and unnecessary overhead in the MTA. Providing a single, reusable SigV4 module centralizes the signing logic, reduces code redundancy in policies, and makes it easier and safer to integrate KumoMTA with AWS services.

### Summary
This PR introduces complete support for AWS Signature Version 4 request signing and exposes a new Lua API that enables KumoMTA to integrate directly with AWS services.

### Key Features

- Implements the full AWS SigV4 algorithm
- Compatible with all major AWS services (S3, SNS, SQS, Kinesis Firehose, etc.)
- Provides SHA256 hashing and HMAC-SHA256 signing
- Secure credential loading via KeySource (file, Vault, inline)
- Supports session tokens for temporary AWS credentials
- Includes comprehensive documentation and example usage
- Adds 5 unit tests and 5 integration tests (all passing)

### Lua module
This PR adds the following function:

`kumo.aws.sign_v4()
`

This enables Lua scripts to generate properly signed AWS API requests, empowering the following use cases:
- Streaming logs to Kinesis Firehose
- Uploading content to S3
- Publishing messages to SNS
- Accessing SQS via signed query parameters


crates/mod-aws-sigv4/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+    .add(b'=')
+    .add(b'?')
+    .add(b'@')
+    .add(b'[')
+    .add(b']');
+
+#[derive(Deserialize, Debug)]
+struct SigV4Request {
+    /// AWS access key ID
+    access_key: String,
```
wez: While it is not required for the username portion of a credential pair, using `KeySource` for it allows eg: updating both the id and secret portion of a pair together if you happen to be using a credential store like a vault.

```suggestion
    access_key: KeySource,
```

crates/mod-aws-sigv4/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+    use aws_lc_rs::digest;
+    let hash = digest::digest(&digest::SHA256, data);
+    HEXLOWER.encode(hash.as_ref())
+}
+
+fn uri_encode(input: &str) -> String {
+    percent_encode(input.as_bytes(), URI_ENCODE_SET).to_string()
+}
+
+fn create_canonical_uri(path: &str) -> String {
```
wez, line 96: Should we pull in https://docs.rs/url/latest/url/struct.Url.html here and use its methods to parse and fix this?

crates/mod-aws-sigv4/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+        .iter()
+        .map(|(k, v)| (uri_encode(k), uri_encode(v)))
+        .collect();
+    encoded_params.sort();
+
+    encoded_params
+        .iter()
+        .map(|(k, v)| format!("{}={}", k, v))
+        .collect::<Vec<_>>()
+        .join("&")
```
wez, line 128: Since the map is already sorted, I think you can simplify something like this:

```suggestion
    params.iter().map(|k, v| format!("{}={}", uri_encode(k), uri_encode(v))).collect::<Vec<_>>().join("&")
```

crates/mod-aws-sigv4/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+        .cloned()
+        .collect::<Vec<_>>()
+        .join(";");
+
+    (header_string, signed_headers)
+}
+
+fn create_signing_key(secret_key: &str, date_stamp: &str, region: &str, service: &str) -> Vec<u8> {
+    let k_date = hmac_sha256(
+        format!("AWS4{}", secret_key).as_bytes(),
```
wez:
```suggestion
        format!("AWS4{secret_key}").as_bytes(),
```

crates/mod-aws-sigv4/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+async fn sign_request(req: SigV4Request) -> anyhow::Result<SigV4Response> {
+    // Get the secret key
+    let secret_key_bytes = req.secret_key.get().await?;
+    let secret_key = std::str::from_utf8(&secret_key_bytes)
+        .context("secret_key must be valid UTF-8")?
+        .to_string();
+
+    // Use provided timestamp or current time
+    let timestamp = req.timestamp.unwrap_or_else(Utc::now);
+    let amz_date = timestamp.format("%Y%m%dT%H%M%SZ").to_string();
```
wez, line 180: Can we use `timestamp.to_rfc3339()` here? (rfc3339 is mostly equivalent to ISO 8601)
The `chrono` crate is fairly high precision with the result of `Utc::now` (eg: you get sub-second results in the resulting timestamp string), and I'm not sure if that is acceptable for sigv4.

If it is, then using `to_rfc3339` here means that we don't have to have "special knowledge" of ISO 8601 for the timestamp format string.

crates/mod-aws-sigv4/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+
+    let canonical_request = format!(
+        "{}\n{}\n{}\n{}\n\n{}\n{}",
+        req.method,
+        canonical_uri,
+        canonical_query_string,
+        canonical_headers,
+        signed_headers,
+        payload_hash
+    );
```
wez, line 211: This form is a bit more readable, especially since there are so many parameters!

```suggestion
    let canonical_request = format!(
        "{method}\n{canonical_uri}\n{canonical_query_string}\n{canonical_headers}\n\n{signed_headers}\n{payload_hash}",
        method = req.method,
    );
```

crates/mod-aws-sigv4/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+
+    // Create string to sign
+    let algorithm = "AWS4-HMAC-SHA256";
+    let credential_scope = format!("{}/{}/{}/aws4_request", date_stamp, req.region, req.service);
```
wez:
```suggestion
    let credential_scope = format!("{date_stamp}/{region}/{service}/aws4_request", region=req.region, service=req.service);
```

crates/mod-aws-sigv4/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+    let canonical_request_hash = sha256_hex(canonical_request.as_bytes());
+
+    let string_to_sign = format!(
+        "{}\n{}\n{}\n{}",
+        algorithm, amz_date, credential_scope, canonical_request_hash
+    );
```
wez:
```suggestion
    let string_to_sign = format!(
        "{algorithm}\n{amz_date}\n{credential_scope}\n{canonical_request_hash}"
    );
```

crates/mod-aws-sigv4/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+    // Calculate signature
+    let signing_key = create_signing_key(&secret_key, &date_stamp, &req.region, &req.service);
+    let signature_bytes = hmac_sha256(&signing_key, string_to_sign.as_bytes());
+    let signature = HEXLOWER.encode(&signature_bytes);
+
+    // Create authorization header
+    let authorization = format!(
+        "{} Credential={}/{}, SignedHeaders={}, Signature={}",
+        algorithm, req.access_key, credential_scope, signed_headers, signature
+    );
```
wez, line 234:
```suggestion
    let authorization = format!(
        "{algorithm} Credential={access_key}/{credential_scope}, SignedHeaders={signed_headers}, Signature={signature}",
        access_key=req.access_key,
    );
```

crates/mod-aws-sigv4/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+    /// Request payload (body)
+    #[serde(default)]
+    payload: String,
+    /// Optional timestamp (defaults to current time)
+    timestamp: Option<DateTime<Utc>>,
+    /// Optional session token for temporary credentials
+    session_token: Option<String>,
+}
+
+#[derive(Deserialize, Debug)]
```
wez: Making this serializable makes it possible to serialize directly into a lua table when returning the final result

```suggestion
#[derive(Deserialize, Serialize, Debug)]
```

crates/mod-aws-sigv4/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+            let response = sign_request(req).await.map_err(any_err)?;
+
+            let result = lua.create_table()?;
+            result.set("authorization", response.authorization)?;
+            result.set("timestamp", response.timestamp)?;
+            result.set("signature", response.signature)?;
+            result.set("canonical_request", response.canonical_request)?;
+            result.set("string_to_sign", response.string_to_sign)?;
+
+            Ok(result)
```
wez: Can use the `Serialize` impl to take care of this boilerplate:

```suggestion
            lua.to_value(&result)
```

crates/mod-aws-sigv4/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+        headers.insert("x-amz-content-sha256".to_string(), payload_hash.clone());
+    }
+
+    // Create canonical request
+    let canonical_uri = create_canonical_uri(&req.uri);
+    let canonical_query_string = create_canonical_query_string(&req.query_params);
+    let (canonical_headers, signed_headers) = create_canonical_headers(&headers);
+
+    let canonical_request = format!(
+        "{}\n{}\n{}\n{}\n\n{}\n{}",
```
wez: Note: there are two `\n\n` in a row in here, is that intentional? It looks like it might be a typo!

https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv-create-signed-request.html#create-canonical-request doesn't seem to indicate that there should be two newlines here!

If it is intentional (eg: it is due to how things get split/joined or whatever the reason), please add a comment here to indicate why.

crates/mod-aws-sigv4/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+        authorization,
+        timestamp: amz_date,
+        canonical_request,
+        string_to_sign,
+        signature,
+    })
+}
+
+pub fn register(lua: &Lua) -> anyhow::Result<()> {
+    let aws_mod = get_or_create_sub_module(lua, "aws")?;
```
wez: Naming wise, how about we register this under `kumo.crypto` as `aws_sign_v4` instead?

```suggestion
    let aws_mod = get_or_create_sub_module(lua, "crypto")?;
```

This will make it slightly less work to put the docs for the function in the right place as part of this PR!

crates/mod-aws-sigv4/src/lib.rs
```diff
@@ -0,0 +1,315 @@
...
+
+    aws_mod.set(
+        "sign_v4",
```
wez: per comment above, let's make this be `kumo.crypto.aws_sign_v4`

```suggestion
        "aws_sign_v4",
```

crates/mod-aws-sigv4/README.md
```diff
@@ -0,0 +1,378 @@
+# AWS SigV4 Module for KumoMTA
```
wez, line 1: Let's move this file to `docs/reference/kumo.crypto/aws_sign_v4.md`.
That way it will show up in the reference docs!

You can use `./docs/build.sh serve` to build the docs and run the doc webserver to review how things look.  Note that the doc build will also format any embedded lua examples using our stylua formatting config to make it match the rest of the docs.

To make it fit in with the other doc pages, the title and preamble here needs to be tweaked a little bit to show the function name correctly in the TOC, the version that this was introduced in, and the function synopsis:

````suggestion
# aws_sign_v4

{{since('dev')}}

```lua
local result = kumo.crypt.aws_sign_v4(PARAMS)
```
````

File: crates/mod-aws-sigv4/test.lua [hunk 16]
```diff
@@ -0,0 +1,157 @@
...
+  headers = {
+    host = 'examplebucket.s3.amazonaws.com',
+  },
+  payload = '',
+}
+
+print('Authorization Header: ' .. result.authorization)
+print('Timestamp: ' .. result.timestamp)
+print('Signature: ' .. result.signature)
+print()
```
wez: This file doesn't actually make any test assertions, so we can't tell if things are working or not.

Please import the policy_utils module and make use of `utils.assert_eq` instead of printing the results; that will make this file into a functional unit test and also help to reduce the verbosity of the test output by having these things only print something when something is broken!

```lua
local utils = require 'policy-extras.policy_utils'

utils.assert_eq(result.authorization, 'SOMETHING')
```

In order to facilitate testing, since the signature is time-dependent, the tests in this file should pass in a fixed time value

crates/mod-aws-sigv4/src/lib.rs
AdityaAudi: I’ve updated `SigV4Request` so that `access_key` is also a `KeySource`, and in `sign_request` we now resolve both the access key id and secret via `KeySource::get()`. That allows rotating/coupling the id+secret together when using Vault or an event-backed source, while keeping the Lua calling shape consistent with other modules that use `KeySource`.

crates/mod-aws-sigv4/src/lib.rs
AdityaAudi: Right now `uri` is treated and documented as the **path** component (e.g. `"/my-bucket/object"`) rather than a full URL, which is what the SigV4 “canonical URI” expects. Given that, using `url::Url` wouldn’t add much here, so I’ve left `create_canonical_uri` as a simple path splitter/encoder and added a comment to make the “path only” expectation explicit.

crates/mod-aws-sigv4/src/lib.rs
AdityaAudi: I kept the explicit “encode then sort” behavior and added a comment explaining why. The SigV4 spec orders parameters by their *encoded* names/values, so we can’t rely solely on the `BTreeMap` order (which is on the unencoded key). Collecting `(encoded_key, encoded_value)` pairs into a `Vec` and then sorting ensures the order is correct even when encoding changes key ordering.

crates/mod-aws-sigv4/src/lib.rs
AdityaAudi: Done – `create_signing_key` now uses `format!("AWS4{secret_key}") `for the date key step.

crates/mod-aws-sigv4/src/lib.rs
AdityaAudi: I’ve kept the explicit `format("%Y%m%dT%H%M%SZ") / format("%Y%m%d")`. SigV4 requires the compact `YYYYMMDD'T'HHMMSS'Z'` form, whereas `to_rfc3339()` produces a full RFC 3339 string (with dashes, offsets, and optional subseconds), so we’d still need to strip/reformat it to match the AWS spec. The current format string also truncates the `Utc::now()` precision down to seconds, which aligns with SigV4.

crates/mod-aws-sigv4/src/lib.rs
AdityaAudi: Updated as suggested. The canonical request now uses a named‑argument `format!` with all of the components spelled out, and I added a comment (with a link to the AWS docs) explaining the required blank line between the canonical headers block and the signed‑headers line.

crates/mod-aws-sigv4/src/lib.rs
AdityaAudi: Applied – `credential_scope` now uses that named‑argument `format!` form, which reads more clearly than the positional version.

crates/mod-aws-sigv4/src/lib.rs
AdityaAudi: Also applied – `string_to_sign` now uses the named‑argument `format!` with `algorithm`, `amz_date`, `credential_scope`, and `canonical_request_hash`.

crates/mod-aws-sigv4/src/lib.rs
AdityaAudi: Done – the Authorization header now uses that named‑argument `format!`, with `access_key`, `credential_scope`, `signed_headers`, and `signature` named explicitly. This also now uses the access key id resolved from its `KeySource`.

crates/mod-aws-sigv4/src/lib.rs
AdityaAudi: Implemented – `SigV4Response` now derives `Serialize` alongside `Deserialize`, and we use that in the Lua bridge.

crates/mod-aws-sigv4/src/lib.rs
AdityaAudi: Done – inside `register`, after calling `sign_request`, we now just call `lua.to_value(&response)` instead of manually constructing a Lua table. The field names and Lua-visible structure remain the same.

crates/mod-aws-sigv4/src/lib.rs
AdityaAudi: Yes, that’s intentional. Per the SigV4 spec there’s a blank line between the canonical headers block and the signed‑headers line. I’ve left the `\n\n` in and added a comment with a link to the AWS canonical request documentation to clarify why the extra newline is there.

crates/mod-aws-sigv4/src/lib.rs
AdityaAudi: Updated as suggested. The function is now registered under the `kumo.crypto` submodule as `aws_sign_v4`, so the Lua API is `kumo.crypto.aws_sign_v4(...)`. All examples and tests were updated to use the new name.

crates/mod-aws-sigv4/test.lua
AdityaAudi: `test.lua` is now a functional test:

- Uses `local crypto = kumo.crypto` and `local utils = require 'policy-extras.policy_utils'`.
- Introduces a fixed timestamp (`FIXED_TIME = '2025-11-21T20:04:15Z`') so signatures are deterministic.
- Each call to `crypto.aws_sign_v4` passes `timestamp = FIXED_TIME`.
- Replaces the previous print‑only output with `utils.assert_eq` checks on the full Authorization header for the S3, SNS, SQS, and Firehose examples, and exercises KeySource + non‑GET in the fourth example.
- This makes the file behave as a proper unit/integration test and only emits output when something fails.

crates/mod-aws-sigv4/README.md
AdityaAudi: I’ve moved the detailed function documentation into `docs/reference/kumo.crypto/aws_sign_v4.md` with the `# aws_sign_v4` heading, `{{since('dev')}}`, a function synopsis, parameter/return tables, and examples. The crate `crates/mod-aws-sigv4/README.md` is now trimmed down to a brief overview plus a basic example and points readers at the reference page for full details.

PR #458, opened 2025-11-26
 Add aws_sign_v4 helper on kumo.http.Request

AdityaAudi:
### Motivation
The core kumo.crypto.aws_sign_v4 function provides a reusable SigV4 signer, but using it with the HTTP client requires manually passing method, URI, query parameters, headers, and payload from the request into the signer on every call. That leads to boilerplate in policies and increases the chance of subtle mismatches (e.g. forgetting to keep URI or query parameters in sync). A helper directly on kumo.http.Request makes it easier and safer to sign AWS requests built with the HTTP client.

### Summary
This PR adds a convenience method request:aws_sign_v4({PARAMS}) to kumo.http.Request that derives the HTTP method, URI path, query parameters, and host header from the request itself, delegates to kumo.crypto.aws_sign_v4, and applies the resulting Authorization and X-Amz-Date headers back onto the same request. The crypto module continues to own the core SigV4 implementation; the HTTP helper is just a thin integration layer. The reference documentation is updated to describe this helper and to cross‑link the HTTP and crypto pages.

### Implemented features in this PR
- **HTTP helper**: Adds `request:aws_sign_v4({PARAMS})` to `kumo.http.Request`, which:
  - Derives `method`, `URI path`, `query parameters`, and `host` from the request.
  - Merges those with caller‑supplied `access_key`, `secret_key`, `region`, `service`, optional `headers`/`payload`/`timestamp`/`session_token`.
  - Calls `kumo.crypto.aws_sign_v4` under the hood and applies `Authorization` and `X-Amz-Date` back onto the same request (returning the request for chaining).
- **Docs**: Updates `kumo.http.Request` docs with a new `request:aws_sign_v4({PARAMS})` section and Kinesis example, and adds a "HTTP client integration" note in `kumo.crypto.aws_sign_v4` that links to the new helper.

### DIFF
--- a/crates/mod-http/src/lib.rs
+++ b/crates/mod-http/src/lib.rs
@@ -395,8 +395,7 @@ impl LuaUserData for HeaderMapWrapper {
 
         methods.add_meta_method(MetaMethod::Pairs, |lua, this, ()| {
             let stateless_iter =
-                lua.create_function(|lua, (ud, key): (mlua::AnyUserData, Option<String>)| {
-                    let this = ud.borrow::<HeaderMapWrapper>()?;
+                lua.create_function(|lua, (this, key): (HeaderMapWrapper, Option<String>)| {
                     let iter = this.0.iter();
 
                     let mut this_is_key = false;


crates/mod-http/src/lib.rs
```diff
@@ -234,6 +234,97 @@ impl LuaUserData for RequestWrapper {
             },
         );
 
+        methods.add_async_method("aws_sign_v4", |lua, this, params: Value| async move {
```
wez: I'd feel happier if we can improve the strength of the typing in this implementation; in particular, I think that this module having hard coded in field names as strings is prone to problems if a later refactory change renames or otherwise changes the types of those fields.

What I think is that it is acceptable for `mod-http` to add `mod-aws-sigv4` as a dependency so that its types and its methods are more directly available.

Then I think the implementation here could either directly re-use those types, or at least use a subset of them.

I'd be inclined to do something like this here, which is overall a bit more succinct and should be more robust in the face of future changes:

```rust
methods.add_async_method("aws_sign_v4", |lua, this, params: SerdeWrappedValue<SigV4Request>| async move {

let mut signer_params = params.0;

let req = ... clone request builder here ...;

signer_params.query_params = req.url().query_pairs().collect();
signer_params.method = req.method().as_str().to_string();
signer_params.uri = req.url().path().to_string();

if !signer_params.headers.contains_key("host") {
   if let Some(host) = req.url().host_str() {
       signer_params.headers.insert("host", host.to_string());
    }
}

let sig = sign_request(signer_params).await.map_err(any_err)?;

this.apply(|b| {
   Ok(b.header("Authorization", sig.authorization)
            .header("X-Amz-Date", sig.timestamp))
})?;

Ok(this.clone())
});
```

`SerdeWrappedValue` is a helper type exported from the `config` crate which makes it a little more convenient to use serde to process parameters.

crates/mod-http/src/lib.rs
```diff
@@ -330,9 +421,22 @@ impl LuaUserData for ResponseWrapper {
 
 // Headermap ---
 
-#[derive(Clone, mlua::FromLua)]
+#[derive(Clone)]
```
wez: with the suggestion above, I think we can leave `HeaderMapWrapper`'s `FromLua` impl as it was prior to this PR

wez (commented):
Thanks for this!

crates/mod-http/src/lib.rs
AdityaAudi: The latest commit addresses the following:
 - Adds mod-aws-sigv4 as a dependency to mod-http
- Uses the shared SigV4Request type directly via serde deserialization
- Clones the request builder to extract method, uri, and query_params automatically
-  Ensures the host header is present
- Calls the existing sign_request() function from mod-aws-sigv4
- Applies the Authorization and X-Amz-Date headers

This eliminates hardcoded field name strings and should be robust against future type changes.

crates/mod-http/src/lib.rs
AdityaAudi: Reverted `HeaderMapWrapper` to just `#[derive(Clone)]` without the `FromLua` derive - back to its original state.

wez: hmm, prior to this pr, this was `#[derive(Clone, mlua::FromLua)]` and the code below to access `HeaderMapWrapper` could just reference the type directly though?

crates/mod-http/src/lib.rs
```diff
@@ -345,7 +395,8 @@ impl LuaUserData for HeaderMapWrapper {
 
         methods.add_meta_method(MetaMethod::Pairs, |lua, this, ()| {
             let stateless_iter =
-                lua.create_function(|lua, (this, key): (HeaderMapWrapper, Option<String>)| {
+                lua.create_function(|lua, (ud, key): (mlua::AnyUserData, Option<String>)| {
+                    let this = ud.borrow::<HeaderMapWrapper>()?;
```
wez: this block is also different from previously. Is that intentional/required? It doesn't appear to necessary?

AdityaAudi: The latest commit eliminates the manual `ud.borrow::<HeaderMapWrapper>()?` call and is cleaner. Tested and builds successfully. Thanks for catching that!

Issue #435, opened 2025-10-24
Add salted hashing functions


tommairs:
We already support digest functions (https://docs.kumomta.com/reference/kumo.digest/)
It could also be helpful to support salted hashes for more security in cases of passing passwords and keys.

For instance, being able to authenticate a user by passing a bcrypt (or PBKDF2, Argon, etc)  encrypted password in-line would eliminate the need for external authentication processes.

PR #459, opened 2025-11-26
feat: add TLS and RFC 1929 auth to KumoProxy

### LINKED ISSUE #451: Add authentication to KumoProxy
Currently KumoProxy relies on external firewall and router configuration to ensure authorized access.  It has been suggested that adding authentication to the proxy would be valuable.

One possible option is to add username+password (basic auth), but other options may be possible.

This ticket is waiting for sponsorship.

vietcgi:
## Summary

- Add TLS encryption support for KumoProxy SOCKS5 server connections
- Implement RFC 1929 username/password authentication for client access control
- Add client-side TLS support in kumod for connecting to TLS-enabled proxies
- Support mTLS (mutual TLS) for client certificate verification

Fixes #451

## Changes

**Proxy Server (crates/proxy-server):**
- New `--tls-certificate`, `--tls-private-key` options for server TLS
- New `--tls-client-ca` option for mTLS client verification
- New `--auth-file` option for TOML-based user credentials
- Generic async stream handling (works with both plain TCP and TLS)
- Constant-time password comparison to prevent timing attacks

**KumoMTA Client (crates/kumod):**
- New `socks5_proxy_username` and `socks5_proxy_password` fields
- New `socks5_proxy_tls`, `socks5_proxy_tls_ca`, `socks5_proxy_tls_insecure` fields
- TLS connector with optional CA verification and insecure mode

**Documentation:**
- Added docs for all new `make_egress_source` options

## Test plan

- [x] Unit tests for auth module (9 tests pass)
- [x] Manual testing on Linux (Ubuntu 24.04 via multipass):
  - Plain TCP + NOAUTH
  - Plain TCP + Auth (valid/invalid credentials)
  - TLS + NOAUTH
  - TLS + Auth
  - Auth rejection with wrong password
  - TLS mismatch (plain client -> TLS server)
- [x] cargo check, cargo clippy pass
- [x] Backward compatible (no auth/TLS required by default)

### DIFF
--- a/docs/changelog/main.md
+++ b/docs/changelog/main.md
@@ -27,6 +27,12 @@
    also supports AMP HTML parts.
  * Improved the context shown in error messages produced by the HTTP injection
    API
+ * New [kumo.start_proxy_listener](../reference/kumo/start_proxy_listener/_index.md)
+   function to configure a SOCKS5 proxy server with optional TLS encryption and
+   RFC 1929 username/password authentication. The proxy uses efficient `splice(2)`
+   zero-copy data transfer on Linux for plain TCP connections.
+   See [proxy_server_auth_rfc1929](../reference/events/proxy_server_auth_rfc1929.md)
+   for authentication callback details. Thanks to @vietcgi! #459
 
 ## Fixes
 
--- a/docs/reference/events/proxy_server_auth_rfc1929.md
+++ b/docs/reference/events/proxy_server_auth_rfc1929.md
@@ -1,0 +1,79 @@
+# proxy_server_auth_rfc1929
+
+```lua
+kumo.on(
+  'proxy_server_auth_rfc1929',
+  function(username, password, conn_meta) end
+)
+```
+
+Called by the proxy server when a client attempts RFC 1929 username/password
+authentication.
+
+This event is triggered when [require_auth](../kumo/start_proxy_listener/require_auth.md)
+is enabled on the proxy listener, or when a client offers username/password
+authentication even if not required.
+
+The event handler receives the following parameters:
+
+* *username* - the username provided by the client
+* *password* - the password provided by the client
+* *conn_meta* - a table containing connection metadata:
+    * `peer_address` - the client's socket address (IP and port)
+    * `local_address` - the server's socket address (IP and port)
+
+The proxy server expects the event handler to return a bool value; if it returns
+`true` then it considers the credentials to be valid and will allow the client
+to use the proxy. If it returns `false` then the authentication attempt is
+considered to have failed and the connection will be closed.
+
+This example shows how to implement a simple inline password database:
+
+```lua
+kumo.on('proxy_server_auth_rfc1929', function(username, password, conn_meta)
+  -- This is just an example of how to populate the return value,
+  -- not a recommended way to handle passwords in production!
+  local password_database = {
+    ['proxyuser'] = 'secretpassword',
+  }
+  if password == '' then
+    return false
+  end
+  return password_database[username] == password
+end)
+```
+
+## Returning Group and Identity Information
+
+Rather than just returning a boolean to indicate whether authentication was
+successful, you may return an [AuthInfo](../kumo.aaa/auth_info.md) object
+holding additional information. Here's an expanded version of the example
+above that shows how you can return group membership:
+
+```lua
+local password_database = {
+  ['proxyuser'] = {
+    password = 'secretpassword',
+    groups = { 'proxy-users', 'premium-tier' },
+  },
+}
+
+kumo.on('proxy_server_auth_rfc1929', function(username, password, conn_meta)
+  local entry = password_database[username]
+  if not entry then
+    return false
+  end
+  if entry.password ~= password then
+    return false
+  end
+
+  -- Return an AuthInfo that lists the identity and group membership
+  return {
+    identities = {
+      { identity = username, context = 'ProxyAuthRfc1929' },
+    },
+    groups = entry.groups,
+  }
+end)
+```
+
--- a/docs/reference/kumo/start_proxy_listener/_index.md
+++ b/docs/reference/kumo/start_proxy_listener/_index.md
@@ -1,0 +1,47 @@
+# kumo.start_proxy_listener
+
+```lua
+kumo.start_proxy_listener { PARAMS }
+```
+
+Configure and start a SOCKS5 proxy server.
+
+This function should be called only from inside your [init](../../events/init.md)
+event handler.
+
+The proxy server implements the SOCKS5 protocol and can be used by KumoMTA's
+egress sources to route outbound connections through the proxy. This is useful
+for scenarios where you need to control the egress IP address or route traffic
+through specific network paths.
+
+```lua
+kumo.on('init', function()
+  kumo.start_proxy_listener {
+    listen = '0.0.0.0:1080',
+  }
+end)
+```
+
+To enable TLS and authentication:
+
+```lua
+kumo.on('init', function()
+  kumo.start_proxy_listener {
+    listen = '0.0.0.0:1080',
+    use_tls = true,
+    tls_certificate = '/path/to/cert.pem',
+    tls_private_key = '/path/to/key.pem',
+    require_auth = true,
+  }
+end)
+
+kumo.on('proxy_server_auth_rfc1929', function(username, password, conn_meta)
+  -- Validate credentials here
+  return username == 'user' and password == 'secret'
+end)
+```
+
+`PARAMS` is a lua table that can accept the keys listed below:
+
+## Proxy Listener Parameters { data-search-exclude }
+
--- a/docs/reference/kumo/start_proxy_listener/hostname.md
+++ b/docs/reference/kumo/start_proxy_listener/hostname.md
@@ -1,0 +1,14 @@
+# hostname
+
+Specifies the hostname to use when generating a self-signed TLS certificate.
+
+The default is the system hostname.
+
+```lua
+kumo.start_proxy_listener {
+  listen = '0.0.0.0:1080',
+  use_tls = true,
+  hostname = 'proxy.example.com',
+}
+```
+
--- a/docs/reference/kumo/start_proxy_listener/listen.md
+++ b/docs/reference/kumo/start_proxy_listener/listen.md
@@ -1,0 +1,13 @@
+# listen
+
+Specifies the local IP and port number to which the proxy service
+should bind and listen.
+
+Use `0.0.0.0` to bind to all IPv4 addresses.
+
+```lua
+kumo.start_proxy_listener {
+  listen = '0.0.0.0:1080',
+}
+```
+
--- a/docs/reference/kumo/start_proxy_listener/no_splice.md
+++ b/docs/reference/kumo/start_proxy_listener/no_splice.md
@@ -1,0 +1,23 @@
+# no_splice
+
+On Linux, the proxy server uses `splice(2)` for efficient zero-copy data
+transfer between the client and remote connections. This can significantly
+reduce CPU usage for high-throughput proxy scenarios.
+
+Set this to `true` to disable splice and use regular userspace copying instead.
+This may be useful for debugging or if you encounter issues with splice on
+your system.
+
+The default is `false` (splice is enabled on Linux).
+
+Note: splice(2) is only used for plain TCP connections. TLS connections
+always use regular copying because the data must be decrypted/encrypted
+in userspace.
+
+```lua
+kumo.start_proxy_listener {
+  listen = '0.0.0.0:1080',
+  no_splice = true,
+}
+```
+
--- a/docs/reference/kumo/start_proxy_listener/require_auth.md
+++ b/docs/reference/kumo/start_proxy_listener/require_auth.md
@@ -1,0 +1,23 @@
+# require_auth
+
+If true, the proxy server will require RFC 1929 username/password authentication
+from clients before allowing them to use the proxy.
+
+When enabled, you must also register a handler for the
+[proxy_server_auth_rfc1929](../../events/proxy_server_auth_rfc1929.md) event
+to validate credentials.
+
+The default is `false`.
+
+```lua
+kumo.start_proxy_listener {
+  listen = '0.0.0.0:1080',
+  require_auth = true,
+}
+
+kumo.on('proxy_server_auth_rfc1929', function(username, password, conn_meta)
+  -- Validate credentials
+  return username == 'user' and password == 'secret'
+end)
+```
+
--- a/docs/reference/kumo/start_proxy_listener/timeout.md
+++ b/docs/reference/kumo/start_proxy_listener/timeout.md
@@ -1,0 +1,15 @@
+# timeout
+
+Specifies the connection timeout duration. If a client does not complete
+the SOCKS5 handshake or send data within this duration, the connection
+will be closed.
+
+The default is 60 seconds.
+
+```lua
+kumo.start_proxy_listener {
+  listen = '0.0.0.0:1080',
+  timeout = '30 seconds',
+}
+```
+
--- a/docs/reference/kumo/start_proxy_listener/tls_certificate.md
+++ b/docs/reference/kumo/start_proxy_listener/tls_certificate.md
@@ -1,0 +1,28 @@
+# tls_certificate
+
+Specify the path to a TLS certificate file to use for the server identity when
+*use_tls* is set to `true`.
+
+The default, if unspecified, is to dynamically allocate a self-signed certificate.
+
+```lua
+kumo.start_proxy_listener {
+  listen = '0.0.0.0:1080',
+  use_tls = true,
+  tls_certificate = '/path/to/cert.pem',
+}
+```
+
+You may specify that the certificate be loaded from a [HashiCorp Vault](https://www.hashicorp.com/products/vault):
+
+```lua
+kumo.start_proxy_listener {
+  listen = '0.0.0.0:1080',
+  use_tls = true,
+  tls_certificate = {
+    vault_mount = 'secret',
+    vault_path = 'tls/proxy.example.com.cert',
+  },
+}
+```
+
--- a/docs/reference/kumo/start_proxy_listener/tls_private_key.md
+++ b/docs/reference/kumo/start_proxy_listener/tls_private_key.md
@@ -1,0 +1,28 @@
+# tls_private_key
+
+Specify the path to a TLS private key file to use for the server identity when
+*use_tls* is set to `true`.
+
+The default, if unspecified, is to dynamically allocate a self-signed certificate.
+
+```lua
+kumo.start_proxy_listener {
+  listen = '0.0.0.0:1080',
+  use_tls = true,
+  tls_private_key = '/path/to/key.pem',
+}
+```
+
+You may specify that the key be loaded from a [HashiCorp Vault](https://www.hashicorp.com/products/vault):
+
+```lua
+kumo.start_proxy_listener {
+  listen = '0.0.0.0:1080',
+  use_tls = true,
+  tls_private_key = {
+    vault_mount = 'secret',
+    vault_path = 'tls/proxy.example.com.key',
+  },
+}
+```
+
--- a/docs/reference/kumo/start_proxy_listener/use_tls.md
+++ b/docs/reference/kumo/start_proxy_listener/use_tls.md
@@ -1,0 +1,16 @@
+# use_tls
+
+If true, the listener will start with TLS enabled and require clients to
+connect using TLS.
+
+When TLS is enabled, you can optionally specify [tls_certificate](tls_certificate.md)
+and [tls_private_key](tls_private_key.md). If not specified, a self-signed
+certificate will be generated automatically.
+
+```lua
+kumo.start_proxy_listener {
+  listen = '0.0.0.0:1080',
+  use_tls = true,
+}
+```
+


crates/kumod/src/egress_source.rs
```diff
@@ -973,17 +1030,152 @@ impl<'a> ProxyProto<'a> {
...
+                        protocol: None,
+                    },
+                ))
+            }
+        }
+    }
+}
+
+/// Build a TLS connector for SOCKS5 proxy connections.
+async fn build_socks5_tls_connector(
```
wez: I'd like to avoid copying and pasting the other existing and honestly rather *complex* tls code here, as it will lead to subtle divergence issues in the future.

What I propose is extracting `crates/rfc5321/src/tls.rs` into a separate crate, eg: `crates/kumo-tls-helper`.

I think that `crates/rfc5321/src/traits.rs` should probably also be moved there, as the two are coupled.

Then the logic here for the socks client should be re-phrased to work in terms of the `TlsOptions` struct exposed by that new `kumo-tls-helper` crate.

crates/proxy-server/src/auth.rs
```diff
@@ -0,0 +1,263 @@
+//! Authentication configuration and validation for the SOCKS5 proxy server.
```
wez: This is a decent effort to get from 0 -> 1, but I know that pretty much as soon as this lands, someone is going to ask for something more than plain text passwords stored in a text file.

What I had in mind for this was to drive it from lua, similar to how we do this in kumod.  That would give total customization in both the physical storage scheme and password verification options, and once something like https://github.com/KumoCorp/kumomta/issues/435 is merged, both places would be able to use it without significant rework.

Getting the proxy-server ready to run lua the same way is a reasonable amount of work on its own, but will enable deploying an actual configuration file for the service and its other options too, so I think that it would be a very nice change.

I propose that you do that as a separate PR from this one that focuses on enabling lua-based configuration of the proxy and we can get than landed/merged, then rebase this PR on top of that.

Take a look at `crates/tsa-daemon/src/main.rs` to see how it sets up `kumo-server-common`.

I think that the proxy server configuration policy file might end up looking something like:

```lua
local kumo = require 'kumo'

kumo.on("proxy_init", function()
   kumo.start_proxy_listener {
      listen = '0.0.0.0:5000',
      timeout = "1 m",
   }
end)
```

which would be equivalent to `proxy-server --listen 0.0.0.0:5000 --timeout-seconds=60`.

Note that this change would also enable spawning multiple proxy listeners from inside the same process, which is a nice upgrade on its own, but does mean that some of those parts of the proxy will probably need to be tweaked to support running in that way.

The other thing to note is that we'd need to make that change in a backwards compatible way, so that folks that are running eg: `proxy-server --listen 0.0.0.0:5000 --timeout-seconds=60` continue to work after upgrading.

I suggest that `proxy-server --proxy-config /path/to/proxy.lua` be the way to explicit load a lua based configuration, and that the other `listen` and `timeout` command line options be mutually exclusive with `--proxy-config`.

With that landed and in place, then this PR could provide authentication support similar to https://docs.kumomta.com/reference/events/smtp_server_auth_plain/

```lua
-- Use this to lookup and confirm a user/password credential
kumo.on('proxy_server_auth_1929', function(username, password, conn_meta)
   -- do something here to verify the username/password
   -- conn_meta could simply be a table exposing the peer address and
   -- local address
end)
```

The end-user is then free to deploy some kind of data lookup and/or memoize the result of the auth check using the various functions that our lua stuff supports.

crates/proxy-server/src/main.rs
```diff
@@ -1,22 +1,46 @@
...
     #[arg(long)]
     no_splice: bool,
 
+    /// Timeout in seconds for all I/O operations
     #[arg(long, default_value = "60")]
     timeout_seconds: u64,
+
+    /// Path to TLS certificate file (PEM format)
+    #[arg(long)]
+    tls_certificate: Option<PathBuf>,
```
wez: With the lua based config changes I suggested above, this options would not be command line options, but instead be added only inside the `start_proxy_listener` function.

crates/proxy-server/src/tls_helpers.rs
```diff
@@ -0,0 +1,104 @@
+//! TLS configuration helpers for the proxy server.
```
wez: by switching to using `kumo-server-common` as proposed above, we can then simply reference its `make_server_config` function directly rather than copying the code into this file.

crates/kumod/src/egress_source.rs
```diff
@@ -889,7 +946,7 @@ impl<'a> ProxyProto<'a> {
                         );
                         anyhow::ensure!(
                             pass.len() < 256,
-                            "username is too long for SOCKS5 protocol"
+                            "password is too long for SOCKS5 protocol"
```
wez: good catch!

crates/integration-tests/src/kumoproxy.rs
wez, line 1: The term "proxy" is a bit broad and potentially conflicts with eg: ha proxy which we might add to the integration test suite in the future.
Let's rename this file to something like `kumoproxy.rs` or `socks5proxy` to avoid needing to do that in the future.

crates/integration-tests/src/test/proxy_basic.rs
wez, line 1: The things that I think are important to test in here are:

1. Spawning in legacy mode still works without any visible change to existing users.
2. An end-to-end test with a source->proxy->sink that verifies that we can pass a message through that chain
3. A version of that end-to-end test that enables authentication on the client and the events in the proxy to validate auth.  This should test both the positive and negative cases: one with working credentials, and one without.
4. Another test that validates that `require_auth` functions correctly

I think you should be able to use `DaemonWithMaildirOptions` to build up the source/sink pair.  I'd suggest making a helper to spawn the test setup, something like this, but obviously you should be able to pass down options to support the different test cases above:

```rust
struct DaemonWithProxy {
   kumod: DaemonWithMaildir,
   proxy: ProxyDaemon,
}

impl DaemonWithProxy {
    async fn spawn(args: ProxyArgs) -> anyhow::Result<Self> {
       let proxy = ProxyDaemon::spawn(args).await?;
       let kumod = DaemonWithMaildirOptions::new()
              .policy_file("source_with_proxy.lua")
              .env("KUMO_PROXY_SERVER_ADDRESS", proxy.listener("proxy"))
              .start().await?;

       Ok(Self {
           proxy,
           kumod
        })
   }
}
```

Then in `crates/integration-tests/source_with_proxy.lua` you can use that env var to configure the egress path:

```lua
kumo.on('get_egress_path_config', function(domain, source_name, _site_name)
   return kumo.make_egress_path {
     socks5_proxy_server = os.getenv('KUMO_PROXY_SERVER_ADDRESS'),
     -- other parameters here
   }
end)
```

You can use env vars to pass down options to the policy so that you know whether you should be trying auth or not, and the auth credentials and so on.

crates/integration-tests/src/proxy.rs
```diff
@@ -0,0 +1,151 @@
...
+pub struct ProxyDaemon {
+    #[allow(unused)]
+    pub dir: TempDir,
+    pub listeners: BTreeMap<String, SocketAddr>,
+    child: Child,
+}
+
+#[derive(Default, Debug)]
+pub struct ProxyArgs {
+    pub policy_file: String,
```
wez: From a naming perspective, let's call this `proxy_config` so that it doesn't get confused with the main kumo policy file.

In order to be able to test the no-config-file case, I'd suggest making it an `Option`.

If `proxy_config` is `None`, I'd suggest that we use `--listen 127.0.0.1:0` instead of `--proxy-config PATH` to start the service.

crates/kumo-tls-helper/src/lib.rs
```diff
@@ -61,7 +88,7 @@ impl std::hash::Hash for RustlsCacheKey {
 
 lruttl::declare_cache! {
 /// Caches TLS connector information for the RFC5321 SMTP client
-static RUSTLS_CACHE: LruCacheWithTtl<RustlsCacheKey, Arc<ClientConfig>>::new("rfc5321_rustls_config", 32);
+static RUSTLS_CACHE: LruCacheWithTtl<RustlsCacheKey, Arc<ClientConfig>>::new("kumo_tls_helper_rustls_config", 32);
```
wez: Renaming this cache is effectively an API change, and we should mention it in the changelog under breaking changes in case someone has a policy that is tuning the cache capacity.

The list of caches is automatically extracted and displayed in https://docs.kumomta.com/reference/kumo/set_lruttl_cache_capacity/ when the docs are built.

I think renaming this makes sense, as its scope is potentially increased as part of moving it.

I'd suggest naming it `rustls_client_config`.

```suggestion
static RUSTLS_CACHE: LruCacheWithTtl<RustlsCacheKey, Arc<ClientConfig>>::new("rustls_client_config", 32);
```

crates/proxy-server/src/main.rs
```diff
@@ -1,81 +1,290 @@
...
+/// Full docs available at: <https://docs.kumomta.com>
 #[derive(Debug, Parser)]
-#[command(about)]
-pub struct Opt {
+#[command(about, version=version_info::kumo_version())]
+struct Opt {
+    /// Lua policy file to load.
+    /// Mutually exclusive with --listen and other legacy options.
+    #[arg(long)]
+    policy: Option<PathBuf>,
```
wez: I think `--proxy-config` has less opportunity to confuse folks in the long run!

```suggestion
    #[arg(long)]
    proxy_config: Option<PathBuf>,
```

crates/proxy-server/src/main.rs
```diff
@@ -1,81 +1,290 @@
...
+    /// for human consumption.
+    ///
+    /// json outputs machine readable records.
+    #[arg(long, default_value = "full")]
+    diag_format: DiagnosticFormat,
+
+    // Legacy CLI options for backwards compatibility
+    // These are mutually exclusive with --policy
+    /// [Legacy] Address(es) to listen on (e.g., "0.0.0.0:5000").
+    /// Mutually exclusive with --policy.
```
wez: You can do something like:

```rust
#[arg(long, conflicts_with="proxy-config")]
```

to have the clap argument parser enforce the mutually exclusive/conflicting arguments.

Be sure to check that this works as you expect: I don't know if it wants `proxy-config` or `proxy_config`, and you won't know which until you try to run it--it won't catch an issue at compile time.

The same can be applied to each of the conflicting arguments.

crates/proxy-server/src/main.rs
```diff
@@ -1,81 +1,290 @@
...
     #[arg(long, default_value = "60")]
     timeout_seconds: u64,
+
+    /// [Legacy] Path to TLS certificate file (PEM format).
+    #[arg(long)]
+    tls_certificate: Option<PathBuf>,
+
+    /// [Legacy] Path to TLS private key file (PEM format).
+    #[arg(long)]
+    tls_private_key: Option<PathBuf>,
```
wez: I think these command line arguments should be removed and made to work solely through the proxy config script; they aren't legacy options because they have never existed in the `main` branch!

```suggestion
```

crates/proxy-server/src/main.rs
```diff
@@ -1,81 +1,290 @@
...
+                    "--tls-certificate and --tls-private-key must be specified together, or not at all"
+                );
+            }
+        }
 
-    if opts.listen.is_empty() {
-        anyhow::bail!("No listeners defined! use the --listen option to specify at least one!");
+        Ok(())
     }
+}
```
wez, line 78: I think this whole block can be removed in favor of using claps `conflicts_with` attribute?

crates/proxy-server/src/main.rs
```diff
@@ -1,81 +1,290 @@
...
+            lua_funcs: &[kumo_server_common::register, mod_proxy::register],
+            policy: &policy,
+        }
+        .run(perform_init(), signal_shutdown())
+        .await
+    }
+}
+
+/// Run in legacy mode using CLI options (backwards compatibility)
+async fn run_legacy(opts: Opt) -> anyhow::Result<()> {
```
wez: I would love to avoid this code duplication!

My suggestion is that when the user doesn't provide a path to a config script, that we create one based on the command line arguments:

* Use `tempfile::NamedTempFile::with_prefix("kumo-proxy-config")` to create a temporary file
* We write out to it based on what we see in `opts`:

```lua
local kumo = require 'kumo'

kumo.on('proxy_init', function()
```

Then loop over the `opts.listen` and emit appropriate listener entries:

```
  kumo.start_proxy_listener {
    listen = LISTENER,
    -- no_splice and timeout here too   
  }
```

and then close it out:

```
end)
```

With that populated, we can then pass in that path via `StartConfig::policy`.

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -0,0 +1,221 @@
...
+) -> anyhow::Result<bool> {
+    let mut config = config::load_config().await?;
+
+    let conn_meta = ConnMeta {
+        peer_address: peer_address.to_string(),
+        local_address: local_address.to_string(),
+    };
+
+    // Use tuple (username, password, conn_meta) for the callback signature
+    let sig = CallbackSignature::<(String, String, ConnMeta), bool>::new("proxy_server_auth_1929");
```
wez: Please use the `declare_event!` macro to declare this event, as it gives us more metadata and clarity of purpose on the various parameters.  In the future, I want to add some runtime type checking around event handlers so having the parameter names to go with the types will be useful for that!

Also, I've renamed the event so that it is `rfc1929` rather than `1929`, which makes it a little clearer why the number is present in the name:

```rust
declare_event! {
static CHECK_AUTH: Single(
   "proxy_server_auth_rfc1929",
   username: String,
   password: String,
   conn_meta: ConnMeta,
) -> bool
};
```

this should be placed at the file scope rather than embedded inside this function.

You can then use `&CHECK_AUTH` instead of `&sig` to reference it below.

crates/proxy-server/src/proxy_handler.rs
```diff
@@ -24,22 +30,98 @@ pub async fn handle_proxy_client(
...
-    if !handshake
-        .methods
-        .into_iter()
-        .any(|m| m == socksv5::v5::SocksV5AuthMethod::Noauth)
-    {
-        return Err(anyhow::anyhow!(
-            "client {peer_address:?} requested authentication, but this proxy only supports NOAUTH"
-        ));
-    }
+    if require_auth {
```
wez: I think this block should be modified a bit.  Right now the logic is:

```
if require_auth:
   allow sending auth
else:
   only allow unauthenticated clients
```

I think it should be:

```
if UsernamePassword in handshake:
    check auth, rejecting session if auth event fails
    done with auth checks
if require_auth and UsernamePassword not in handshake:
    reject session
    done with auth checks
if Noauth in handshake:
    accept,
    done with auth checks
reject session because we don't know how to authenticate
```

this way we can have optionally authenticated sessions and makes things a bit more flexible.

I think this auth checking logic could (and probably should) be extracted into a function so that you can simply use either an explicit `return Ok(())` for the accept case, or `anyhow::bail` for the reject case.

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -0,0 +1,221 @@
...
+        )
+        .await
+    }
+}
+
+/// Connection metadata passed to the proxy_server_auth_1929 callback
+#[derive(Clone, Debug, serde::Serialize)]
+pub struct ConnMeta {
+    pub peer_address: String,
+    pub local_address: String,
```
wez: you can have serde take care of serializing these to strings

```suggestion
    pub peer_address: SockAddr,
    pub local_address: SockAddr,
```

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -0,0 +1,221 @@
...
+    username: String,
+    password: String,
+    peer_address: SocketAddr,
+    local_address: SocketAddr,
+) -> anyhow::Result<bool> {
+    let mut config = config::load_config().await?;
+
+    let conn_meta = ConnMeta {
+        peer_address: peer_address.to_string(),
+        local_address: local_address.to_string(),
```
wez:
```suggestion
        peer_address,
        local_address,
```

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -0,0 +1,227 @@
...
+        if self.use_tls {
+            tracing::info!("proxy listener (TLS) on {addr:?}");
+        } else {
+            tracing::info!("proxy listener on {addr:?}");
+        }
+
+        let params = Arc::new(self);
+
+        spawn(format!("proxy listener {addr:?}"), async move {
+            Self::accept_loop(listener, params, tls_acceptor).await
```
wez:
```suggestion
            if let Err(err) = Self::accept_loop(listener, params, tls_acceptor).await {
                tracing::error!("accept loop returned with error: {err:#}");
            }
```

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -0,0 +1,227 @@
...
+        tls_acceptor: Option<TlsAcceptor>,
+    ) {
+        // Get the local address from the listener
+        let local_address = match listener.local_addr() {
+            Ok(addr) => addr,
+            Err(err) => {
+                tracing::error!("failed to get local address: {err:#}");
+                return;
+            }
+        };
```
wez:
```suggestion
        let local_address = listener.local_addr().context("failed to get local address")?;
```

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -0,0 +1,227 @@
...
+
+        loop {
+            let (socket, peer_address) = match listener.accept().await {
+                Ok(tuple) => tuple,
+                Err(err) => {
+                    tracing::error!("accept failed: {err:#}");
+                    return;
+                }
+            };
```
wez:
```suggestion
            let (socket, peer_address) = listener.accept().await.context("accept failed")?;
```

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -0,0 +1,227 @@
...
+                            return;
+                        }
+                    }
+                } else {
+                    Self::handle_client(socket, peer_address, local_address, &params).await
+                };
+
+                if let Err(err) = result {
+                    tracing::error!("proxy session error from {peer_address:?}: {err:#}");
+                }
```
wez: Can make this a bit more succinct by putting the body in an `async move { }` block and centralizing capturing the error and logging it, like this:

```suggestion
                let result = async move {
                    if let Some(acceptor) = tls_acceptor {
                     let tls_stream = acceptor.accept(socket).await.with_context(||format!("failed TLS handshake from {peer_address:?}")?;
                      Self::handle_client(tls_stream, peer_address, local_address, &params)
                                .await
                } else {
                    Self::handle_client(socket, peer_address, local_address, &params).await
                }
                };

                if let Err(err) = result {
                    tracing::error!("proxy session error from {peer_address:?}: {err:#}");
                }
```

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -0,0 +1,227 @@
...
+    }
+}
+
+declare_event! {
+    static CHECK_AUTH: Single(
+        "proxy_server_auth_rfc1929",
+        username: String,
+        password: String,
+        conn_meta: ConnMeta,
+    ) -> bool;
```
wez: Brand new in `main` is an enhanced access control layer (you can get a high level overview in https://docs.kumomta.com/reference/access_control/) which allows returning supplemental information, such as group membership, from the underlying authentication data store.

The way that works is that the authentication function returns a new `AuthInfo` type holding that information.

You can see examples of those events returning `AuthInfo` here in the docs:
https://docs.kumomta.com/reference/kumo.aaa/auth_info/#constructing-an-authinfo

To make the proxy work with this, you'll need to change the return value of the event like this:

```suggestion
    ) -> SerdeWrappedValue<AuthKindResult>;
```

You can see how we process the result of auth in the smtp server over here:

https://github.com/KumoCorp/kumomta/blob/1598adb8ccf84359d7075bb9e50844506e16a38b/crates/kumod/src/smtp_server.rs#L2640-L2719

Smtp is slightly more complex because there are two identities that are part of the authentication request. We just have a single identity here in the proxy auth.

I've predefined `IdentityContext::ProxyAuthRfc1929` to be used here when populating the `AuthInfo`.

crates/proxy-server/src/proxy_handler.rs
```diff
@@ -239,3 +239,198 @@ async fn request_addr(request: &SocksV5Request) -> std::io::Result<SocketAddr> {
...
+/// 3. If client offers Noauth and we don't require auth, accept
+/// 4. Otherwise reject with no acceptable method
+async fn perform_auth<S>(
+    stream: &mut S,
+    handshake: &socksv5::v5::SocksV5Handshake,
+    peer_address: SocketAddr,
+    local_address: SocketAddr,
+    timeout_duration: std::time::Duration,
+    require_auth: bool,
+) -> anyhow::Result<()>
```
wez: I think this function should return the `AuthInfo` that we produce as part of performing authentication.

I'd suggest that we first create an AuthInfo with the peer address; this represents the unauthenticated session corresponding to the peer address:

```rust
let mut auth_info = AuthInfo::default();
auth_info.set_peer_address(Some(peer_address.ip()));
```

we can then merge in any additional data returned from the event callback to produce an authenticated state.

We can return that auth_info; for the Noauth case we return it without any additional identities.

While we don't do anything with that information here in this PR, having it available will enable a subsequent PR to use the new ACL system to create policies that can grant/restrict the ability to bind to specific local addresses/CIDRs or connect to specific remote addresses/CIDRs.

```suggestion
) -> anyhow::Result<AuthInfo>
```

crates/integration-tests/src/test/proxy_basic.rs
```diff
@@ -0,0 +1,335 @@
...
+    eprintln!("sending message through proxy");
+    let mut client = daemon
+        .kumod
+        .smtp_client()
+        .await
+        .context("make smtp_client")?;
+
+    let response = MailGenParams {
+        ..Default::default()
+    }
```
wez:
```suggestion
    let response = MailGenParams::default()
```

crates/integration-tests/src/test/proxy_basic.rs
```diff
@@ -0,0 +1,335 @@
...
+    .send(&mut client)
+    .await
+    .context("send message")?;
+
+    eprintln!("response: {response:?}");
+    // Message should be accepted by the source kumod, but delivery will fail
+    // because proxy auth fails. The message will be queued but not delivered.
+
+    // Give it a moment to attempt delivery
+    tokio::time::sleep(Duration::from_secs(2)).await;
```
wez: This is prone to flakeyness in CI if something happens to take longer than a couple of seconds due to high load.

You can avoid that by using the `wait_for_source_summary` method.  Here's an example:

https://github.com/KumoCorp/kumomta/blob/1598adb8ccf84359d7075bb9e50844506e16a38b/crates/integration-tests/src/test/xfer.rs#L45-L53

That method works by looking at the various log records and summing them up. That particular example waits for the Reception count and TransientFailure count to both be larger than 0, or for 50 seconds to elapse, whichever is soonest.

For this particular test here you probably just want to wait for `TransientFailure > 0`, assuming that the failed auth shows up as a transient failure.

crates/integration-tests/src/test/proxy_basic.rs
```diff
@@ -0,0 +1,335 @@
...
+
+    // Verify NO messages were delivered (auth should have failed)
+    let messages = daemon.kumod.extract_maildir_messages()?;
+    anyhow::ensure!(
+        messages.is_empty(),
+        "expected 0 messages delivered with wrong auth, got {}",
+        messages.len()
+    );
```
wez: I'd suggest making an assertion based on the delivery summary, similar to this:

https://github.com/KumoCorp/kumomta/blob/1598adb8ccf84359d7075bb9e50844506e16a38b/crates/integration-tests/src/test/xfer.rs#L88-L105

this way to you can assert that you saw the appropriate number of failures and deliveries rather than inferring that the behavior was correct based on the absence of messages in the maildir.

crates/integration-tests/src/test/proxy_basic.rs
```diff
@@ -0,0 +1,335 @@
...
+    // Give it a moment to attempt delivery
+    tokio::time::sleep(Duration::from_secs(2)).await;
+
+    // Verify NO messages were delivered
+    let messages = daemon.kumod.extract_maildir_messages()?;
+    anyhow::ensure!(
+        messages.is_empty(),
+        "expected 0 messages when require_auth rejects client, got {}",
+        messages.len()
+    );
```
wez: please also adapt to follow the suggestions made for the test above!

crates/integration-tests/maildir-sink.lua
```diff
@@ -38,6 +38,8 @@ kumo.on('init', function()
   kumo.configure_local_logs {
     log_dir = TEST_DIR .. '/logs',
     max_segment_duration = '1s',
+    min_free_space = 0,
+    min_free_inodes = 0,
```
wez: Let's revert this, as it isn't part of the main purpose of this PR, and the default behavior of min_free_xxx is desirable in the main CI system!

```suggestion
```

crates/integration-tests/proxy_init.lua
wez, line 1: I think we can eliminate this file and just use `proxy_with_auth.lua` and it should behave the same way?

crates/integration-tests/proxy_auth.lua
wez, line 1: I think we can eliminate this file and just use `proxy_with_auth.lua` and it should behave the same way?

crates/integration-tests/source_with_proxy.lua
```diff
@@ -0,0 +1,90 @@
...
+
+  kumo.start_http_listener {
+    listen = '127.0.0.1:0',
+  }
+
+  kumo.configure_local_logs {
+    log_dir = TEST_DIR .. '/logs',
+    max_segment_duration = '1s',
+    min_free_space = 0,
+    min_free_inodes = 0,
```
wez:
```suggestion
```

wez (commented):
Thank you for your work on this, I'm excited to see this get merged!

It took me a little while to get back to you on this next round of review because I was building out the new ACL system with this in mind and wanted to have that in a better state before suggesting here how to integrate those changes here.

vietcgi:
Hey @wez,
  
I've addressed all your feedback from today's review:

Switched to AuthInfo return type with the new ACL system integration
Using wait_for_source_summary instead of sleep in tests + proper delivery assertions
Consolidated the lua configs into just proxy_with_auth.lua
Reverted the maildir-sink.lua changes
All the small suggestions (MailGenParams::default, error handling cleanup, etc)

I squashed it down to 2 clean commits. Ready for another look when you get a chance!

crates/integration-tests/proxy_auth.lua
vietcgi: sure we can

crates/integration-tests/src/test/proxy_basic.rs
```diff
@@ -0,0 +1,365 @@
...
+        .await
+        .context("send message")?;
+
+    eprintln!("response: {response:?}");
+    anyhow::ensure!(response.code == 250, "expected 250, got {}", response.code);
+
+    // Wait for delivery
+    daemon
+        .kumod
+        .wait_for_maildir_count(1, Duration::from_secs(10))
```
wez: let's raise these timeouts; we've definitely seen flakes in CI when the timeouts are this short, and things will typically complete in a couple of seconds when the system is operating correctly, so it shouldn't hurt to make these closer to 1 minute:

```suggestion
        .wait_for_maildir_count(1, Duration::from_secs(50))
```

crates/integration-tests/src/test/proxy_basic.rs
```diff
@@ -0,0 +1,365 @@
...
```
wez:
```suggestion
        .wait_for_maildir_count(1, Duration::from_secs(50))
```

crates/integration-tests/src/test/proxy_basic.rs
```diff
@@ -0,0 +1,365 @@
...
+    eprintln!("response: {response:?}");
+    // Message should be accepted by the source kumod, but delivery will fail
+    // because proxy auth fails. The message will be queued but not delivered.
+
+    // Wait for transient failure to be logged (auth failure shows as transient failure)
+    daemon
+        .kumod
+        .wait_for_source_summary(
+            |summary| summary.get(&TransientFailure).copied().unwrap_or(0) > 0,
+            Duration::from_secs(10),
```
wez:
```suggestion
            Duration::from_secs(50),
```

crates/integration-tests/src/test/proxy_basic.rs
```diff
@@ -0,0 +1,365 @@
...
+        )
+        .await;
+
+    // Verify NO messages were delivered (auth should have failed)
```
wez, line 239: Please also add an assertion that the delivery count in source_counts is 0

crates/integration-tests/src/test/proxy_basic.rs
```diff
@@ -0,0 +1,365 @@
...
+        .context("send message")?;
+
+    eprintln!("response: {response:?}");
+
+    // Wait for transient failure to be logged (require_auth rejection shows as transient failure)
+    daemon
+        .kumod
+        .wait_for_source_summary(
+            |summary| summary.get(&TransientFailure).copied().unwrap_or(0) > 0,
+            Duration::from_secs(10),
```
wez:
```suggestion
            Duration::from_secs(50),
```

crates/integration-tests/src/test/proxy_basic.rs
```diff
@@ -0,0 +1,365 @@
...
+        )
+        .await;
+
+    // Verify transient failures occurred (require_auth rejected the client)
+    let delivery_summary = daemon.kumod.dump_logs().await?;
+    anyhow::ensure!(
```
wez, line 306: please also assert that the delivery count is 0 here too

crates/proxy-server/src/main.rs
```diff
@@ -1,81 +1,157 @@
...
+        let temp_file = opts.generate_legacy_config()?;
+        let path = temp_file.path().to_path_buf();
+        (path, Some(temp_file))
+    } else {
+        // Modern mode: use the provided config file
+        let path = opts
+            .proxy_config
+            .unwrap_or_else(|| PathBuf::from("/opt/kumomta/etc/proxy/init.lua"));
+        (path, None)
+    };
```
wez: It's slightly more idiomatic and cleaner to flip this around this way; we can remove the impossible `unwrap_or_else` case and its fallback to a default path that we'll never use:

```suggestion
    let legacy_file;
    
    let policy_path = match opts.proxy_config {
        Some(path) => path,
        None => {
           // Legacy mode: generate a temporary Lua config from CLI options
           legacy_file = opts.generate_legacy_config()?;
           temp_file.path().to_path_buf()
        }
    };
```

That `let legacy_file;` line looks weird on its own up there at first, but it is the idiomatic way to hold ownership of the temporary file object and avoid adding boilerplate/noise to thread an optional extra thing around through the match.

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -0,0 +1,244 @@
...
+        conn_meta: ConnMeta,
+    ) -> SerdeWrappedValue<AuthKindResult>;
+}
+
+/// Validates credentials via the proxy_server_auth_rfc1929 Lua callback.
+/// The callback receives (username, password, conn_meta) where conn_meta is a table
+/// containing peer_address and local_address.
+/// Returns an AuthInfo on success (with the authenticated identity populated),
+/// or None if authentication failed.
+pub async fn authenticate_user(
```
wez: I'm normally in favor of breaking things out into separate functions like this, but I think that in this case it leads to more code than is necessary; I think that we'll have clearer and more concise code if this logic was inlined into the `perform_auth` function so that it could then directly consult the `AuthKindResult` and either communicate the authentication failure directly, or directly update its `auth_kind` state.

crates/proxy-server/src/proxy_handler.rs
```diff
@@ -239,3 +244,206 @@ async fn request_addr(request: &SocksV5Request) -> std::io::Result<SocketAddr> {
...
+                    })?
+                    .with_context(|| {
+                        format!("failed to send auth success response to {peer_address:?}")
+                    })?;
+
+                // Merge returned auth info (groups, identities) into our auth_info
+                auth_info.merge_from(returned_auth_info);
+                return Ok(auth_info);
+            }
+        }
```
wez, line 466: as mentioned above, I think that if we inline `authenticate_user` here the amount of code here will decrease and be easier to follow and maintain!

crates/proxy-server/src/proxy_handler.rs
```diff
@@ -239,3 +245,251 @@ async fn request_addr(request: &SocksV5Request) -> std::io::Result<SocketAddr> {
...
+        let callback_result = config
+            .async_call_callback(
+                &crate::mod_proxy::CHECK_AUTH,
+                (username.clone(), password, conn_meta),
+            )
+            .await;
+        config.put();
+
+        let auth_kind_result = match callback_result {
+            Ok(wrapped) => wrapped.0,
```
wez: `config.put()` should not be called in the error case

```suggestion

        let auth_kind_result = match callback_result {
            Ok(wrapped) => {
              config.put();
              wrapped.0
            }
```

wez (commented):
Thank you!

I noticed when checking it out and building locally that `splice(2)` wasn't hooked up, so I added some code that should work to enable that again.  It compiles, but I haven't explicitly verified that it it is operating correctly.  I force-pushed my updates to your branch.

Could you take a look to verify that splice is operating correctly? This doesn't strictly require an integration test (although I would appreciate it if you feel like making one!).  To verify I suppose that you could inspect the log output to see that splice is being used, and then also check to make sure that we're not leaking any additional file descriptors while the service is running and handling connections.  I would suggest looking at `lsof -np $(pgrep proxy-server)` and comparing splice to no-splice mode.

With that verified, the only thing remaining is to write up a doc page for the new `kumo.start_proxy_listener` function, then add an entry to the changelog for it!

Thank you again for your work and attention here!

wez:
Thank you! I've merged this. I made a couple of fixups to the docs, and I also renamed the `no_splice` option in lua to `use_splice` to avoid negative logic.  The command line argument remains `--no-splice`.

I think the only thing left to do now before we can consider #451 complete is to enable using TLS on the client side; then we can document the proxy user/pass and TLS options.

Would you mind opening a PR for that?

Thank you again!

PR #463, opened 2025-12-16
Add dmarc reporting

sophiajt:
The start of reporting support for DMARC. We create the report from the errors we encounter, though we do not yet have this wired into the email reporting system.

Our main task is collecting the errors, though where we put them still needs to be decided. Currently, a temporary file is used.

Once collected, the errors can be aggregated and built into the RFC 7489-compliant report.


assets/policy-extras/mail_auth.lua
```diff
@@ -97,8 +97,18 @@ function mod.check(msg, config)
   end
 
   local dmarc_auth_result = nil
-  if config.dmarc then
-    local dmarc_disp = kumo.dmarc.check_msg(msg, {}, config.resolver)
+  if config.dmarc and config.dkim and config.spf then
+    local dkim_auth_results = msg:dkim_verify(config.resolver)
+    local spf_disp = kumo.spf.check_msg(msg, config.resolver)
+    spf_auth_result = spf_disp.result
+
```
wez: these are all computed above already, don't need to repeat that work here?
Except perhaps to move the `local spf_disp` outside of the conditional on line 71 alongside the spf_auth_result initializer?

```suggestion
```

crates/kumo-dmarc/src/types/results.rs
```diff
@@ -49,15 +85,58 @@ pub enum DkimResult {
...
+            "policy" => DkimResult::Policy,
+            "neutral" => DkimResult::Neutral,
+            "temperror" => DkimResult::TempError,
+            "permerror" => DkimResult::PermError,
+            _ => DkimResult::None,
+        }
+    }
+}
+
+impl From<BString> for DkimResult {
```
wez: I'm a little surprised to see BString here; is that due to my recent parser refactoring? I suppose it comes from AuthenticationResults props now returning BStrings.

It's not especially important, but I would like to reduce the repetition here between this and the above impl.

Suggestions are:

1. Can we `impl From<AsRef<[u8]>> for DkimResult` or whatever the precise syntax works out to be, and just have one that works on anything convertible to a byte slice? 
2. For the BString variant, the strings are small enough that I'd be ok with simply doing `value.to_string().into()` for the BString impl, so that we at least avoid duplicating the string matching.

Bonus points for avoiding making a new string via `to_lowercase` and using `eq_ascii_ignore_case`.  But the strings are short enough that I don't find the lowercasing critical, so not something to loose sleep over.

crates/kumo-dmarc/src/lib.rs
```diff
@@ -1,20 +1,35 @@
...
-
-pub use types::results::DmarcResult;
+use uuid::Uuid;
 
 mod types;
 
 #[cfg(test)]
 mod tests;
 
+const DMARC_REPORT_LOG_FILEPATH: &'static str = "/tmp/example.log";
```
wez: `/var/log/kumomta` is prepared by the deb/rpm installer to have appropriate permissions for the kumod user, so it is a better hardcoded default than `/tmp` for the timebeing.

```suggestion
const DMARC_REPORT_LOG_FILEPATH: &'static str = "/var/log/kumomta/dmarc.log";
```

crates/kumo-dmarc/src/lib.rs
```diff
@@ -1,20 +1,35 @@
...
+
 pub struct CheckHostParams {
```
wez: I think the name of this struct is a carryover from the SPF crate.  I think it could probably do with renaming to better match its current purpose (because we're not checking a host), but that's a super minor detail vs. everything else going on in here at the moment!

crates/kumo-dmarc/src/lib.rs
```diff
@@ -95,21 +166,234 @@ impl<'a> DmarcContext<'a> {
...
+    pub async fn report_error(
+        &self,
+        record: &Record,
+        dmarc_domain: &str,
+        sender_domain_alignment: SenderDomainAlignment,
+        error: &str,
+    ) -> std::io::Result<()> {
+        let source_ip = match self.received_from.parse() {
+            Ok(source_ip) => source_ip,
+            Err(_) => Ipv4Addr::new(127, 0, 0, 1).into(),
```
wez: I'm not sure about this.  The source IP is pretty important for this sort of report, so if we're implicitly producing the loopback we're silently masking this error and producing a wonky report for someone. Can we perhaps push the conversion up to a higher level? Whether we can move the conversion or not, we should raise and propogage this error and have it bubble up and prevent us from trying to operate on this record if the source ip is bogus.

crates/kumo-dmarc/src/lib.rs
```diff
@@ -95,21 +166,234 @@ impl<'a> DmarcContext<'a> {
...
+                    envelope_to: self.recipient_list.into(),
+                    envelope_from: if let Some(mail_from_domain) = self.mail_from_domain {
+                        vec![mail_from_domain.into()]
+                    } else {
+                        vec![]
+                    },
+                    header_from: self.from_domain.into(),
+                },
+                auth_results: AuthResults {
+                    dkim: self.dkim_results.iter().map(|x| x.clone().into()).collect(),
```
wez, line 235: Could we `self.dkim_results.to_vec()` here?  Just looks slightly clunky; I suppose the critical part is the `.into()`, so perhaps no.

crates/kumo-dmarc/src/lib.rs
```diff
@@ -95,21 +166,234 @@ impl<'a> DmarcContext<'a> {
...
+                },
+            };
+
+            let result = serde_json::to_string(&error_record)?;
+
+            let mut f = File::options()
+                .append(true)
+                .open(DMARC_REPORT_LOG_FILEPATH)?;
+
+            writeln!(f, "{result}")?;
```
wez, line 246: The fmt impl can produce a number of discrete `write(2)` calls down to the underlying file which is both potentially expensive and, if there are concurrent writers, can result in interleaved writes at the end of the file.

You can avoid the fragmentation issue by buffering up the fmt operation; you can either use a temporary string to hold this (eg: `format!` it and then `f.write_all()` the result), or explicitly use a BufWriter.

Both of these techniques help to reduce fragmentation but don't guarantee a single atomic append write.

If there is the potential for other threads to concurrently be trying to write to this file, then we have a race condition that we'll need to solve with a Mutex of some kind.  We cannot reliably use OS file locking for this, as unix kernel locks are process-scoped rather than thread-scoped, and if the file is on a remote filesystem, then they may not work correctly at all.

crates/kumo-dmarc/src/lib.rs
```diff
@@ -95,21 +166,234 @@ impl<'a> DmarcContext<'a> {
...
+
+            // self.aggregate().await;
```
wez: Can probably remove this? Feels like a stray bit of debugging logic!

```suggestion
```

crates/kumo-dmarc/src/lib.rs
```diff
@@ -95,21 +166,234 @@ impl<'a> DmarcContext<'a> {
...
+        Ok(())
+    }
+
+    pub async fn aggregate(&self) -> std::io::Result<()> {
+        let mut output = vec![];
+        let file = File::open(DMARC_REPORT_LOG_FILEPATH)?;
+        let lines = BufReader::new(file).lines();
+
+        for line in lines.map_while(Result::ok) {
+            output.push(serde_json::from_str::<ErrorRecord>(&line)?);
```
wez: from a robustness perspective, this logic will break out at the first sign of error.
That's probably reasonable wrt. to processing the report, but it will leave the things as-is.  Since the most likely cause of an error here is a corrupt file of some kind, a human will need to do something about this.

I think we should produce some helpful context here; I'd recommend that we switch to `anyhow::Result` for the result type so that we can do something like:

```rust
output.push(
  serde_json::from_str::<ErrorRecord>(&line)
    .with_context(|| {
      format!("Failed to decode a line from the DMARC report file \
           {DMARC_REPORT_LOG_FILEPATH}. \
           The line was: {line}. \
           Is the file corrupt?")
    })?
);
```

It would be good to self-heal/recover from this sort of thing.  Probably the most reasonable outcome would be to rename the file away so that we can't accumulate more things into it, and don't keep trying to read from it on a subsequent aggregation run.  We don't have to do that in this PR; I know that we talked about the next PR in this area being focused on providing control over where the log file lives and so on, so we can come back to tackle this on the other side of it.

crates/kumo-dmarc/src/lib.rs
```diff
@@ -95,21 +166,234 @@ impl<'a> DmarcContext<'a> {
...
+            let version = first_records[0].version.clone();
+            let org_name = first_records[0].org_name.clone();
+            let email = email.clone();
+            let extra_contact_info = first_records[0].extra_contact_info.clone();
+
+            let mut date_range = DateRange::new(first_records[0].when, first_records[0].when);
+
+            let report_id = Uuid::new_v4().to_string();
+
+            let domain = first_records[0].domain.clone();
```
wez: super minor nit, but can we do something like `let first_record = &first_records[0];` to factor out the repeated `first_records[0]` referencing?

crates/kumo-dmarc/src/lib.rs
```diff
@@ -95,21 +166,234 @@ impl<'a> DmarcContext<'a> {
...
+            writeln!(f, "{result}")?;
+
+            // self.aggregate().await;
+        }
+
+        Ok(())
+    }
+
+    pub async fn aggregate(&self) -> std::io::Result<()> {
+        let mut output = vec![];
```
wez: naming nit: `output` implies something that we're producing to be consumed by someone else, but in the context of this function, this is more of an input!  Does it make sense to name it something like `input_records`?  Or substitute your own choice of name :)

crates/kumo-dmarc/src/lib.rs
```diff
@@ -95,21 +166,234 @@ impl<'a> DmarcContext<'a> {
...
+
+        for line in lines.map_while(Result::ok) {
+            output.push(serde_json::from_str::<ErrorRecord>(&line)?);
+        }
+
+        let mut errors_grouped_by_email: HashMap<String, HashMap<IpAddr, Vec<ErrorRecord>>> =
+            HashMap::new();
+
+        'outer: for record in output {
+            for (email, group) in errors_grouped_by_email.iter_mut() {
```
wez: Since `errors_grouped_by_email` is a `HashMap`, can we do `errors_grouped_by_email.entry(record.email.clone())` here to get a mutable reference to the group?

crates/kumo-dmarc/src/lib.rs
```diff
@@ -95,21 +166,234 @@ impl<'a> DmarcContext<'a> {
...
+            new_group.insert(record.source_ip.clone(), vec![record]);
+            errors_grouped_by_email.insert(record_email, new_group);
+        }
+
+        for (email, errors_grouped_by_ip) in errors_grouped_by_email.iter_mut() {
+            let mut errors = vec![];
+            let mut record = vec![];
+
+            //we know this is safe to do because for this list to be present, we will have found it earlier
+            let (_, first_records) = errors_grouped_by_ip.iter().next().unwrap();
```
wez: I like to avoid bare unwraps and put the explaining comment into an `.expect`, something like:

```suggestion
            let (_, first_records) = errors_grouped_by_ip.iter().next().expect("guaranteed to not be empty by the logic above");
```

wez: From a determinism perspective, this first record that we pick with this logic could be different on different runs due to the randomness in the hash map initializer.

Is that a practical problem?  Should we use a BTreeMap if so?

crates/kumo-dmarc/src/lib.rs
```diff
@@ -95,21 +166,234 @@ impl<'a> DmarcContext<'a> {
...
+                for group_error in error_group_for_ip.iter() {
+                    errors.push(group_error.error.clone());
+
+                    if date_range.begin > group_error.when {
+                        date_range.begin = group_error.when;
+                    }
+
+                    if date_range.end < group_error.when {
+                        date_range.end = group_error.when;
+                    }
```
wez: Can simplify like this?

```suggestion
                    date_range.begin = date_range.begin.min(group_error.when);
                    date_range.end = date_range.end.max(group_error.when);
```

crates/kumo-dmarc/src/lib.rs
```diff
@@ -95,21 +166,234 @@ impl<'a> DmarcContext<'a> {
...
+                        .auth_results
+                        .dkim
+                        .extend_from_slice(&group_error.auth_results.dkim);
+                    results
+                        .auth_results
+                        .spf
+                        .extend_from_slice(&group_error.auth_results.spf);
+                }
+
+                record.push(results);
```
wez, line 369: wondering if we should have a step prior to this push where we sort and dedup `results.identifiers.envelope_from` and `envelope_to` to avoid it being potentially filled with redundant entries.

I suppose if we did that, we probably should do the same for the dkim and spf results?

I confess to not having looked at the RFC in detail for this bit!

crates/kumo-dmarc/src/lib.rs
```diff
@@ -95,21 +166,234 @@ impl<'a> DmarcContext<'a> {
...
+                    align_spf,
+                    policy,
+                    subdomain_policy,
+                    rate,
+                    report_failure,
+                ),
+                record,
+            };
+
+            // if let Ok(result) = instant_xml::to_string(&feedback) {
```
wez, line 394: I'm assuming that this is the point at which we'd like to emit a lua event to let the policy glue generate an email for the xml version of the report, when we write that in the next PR?

crates/kumo-dmarc/src/tests.rs
```diff
@@ -337,21 +339,52 @@ async fn evaluate_ip<'a>(
 ) -> DispositionWithContext {
     let mut dkim_vec = vec![];
 
+    let spf_result = AuthenticationResult {
+        method: BString::new("spf".as_bytes().to_vec()),
```
wez: Can hopefully simplify the `BString` construction with a simple `.into()`:

```suggestion
        method: "spf".into(),
```

crates/kumod/src/dmarc.rs
```diff
@@ -32,6 +45,16 @@ pub fn register<'lua>(lua: &'lua Lua) -> anyhow::Result<()> {
 
                 let mail_from_domain = msg_sender.ok().map(|x| x.domain().to_string());
 
+                let recipient_list = msg
```
wez: naming: this is more of a `recipient_domain_list`.

What are the intended semantics if the underlying recipient list has `["foo@gmail.com", "bar@gmail.com"]`?

Right now this will produce here a list: `["gmail.com", "gmail.com"]`.  Should that be deduplicated before we continue with other logic?

crates/kumod/src/dmarc.rs
```diff
@@ -72,10 +95,38 @@ pub fn register<'lua>(lua: &'lua Lua) -> anyhow::Result<()> {
                 let dkim_results: Vec<AuthenticationResult> =
                     config::from_lua_value(&lua, dkim_results)?;
 
+                let spf_result: AuthenticationResult = config::from_lua_value(&lua, spf_result)?;
+
+                let received_from = spf_result
```
wez, line 100: re: earlier comment about parsing IP address, could we do that on the next line?

crates/kumod/src/dmarc.rs
```diff
@@ -72,10 +95,38 @@ pub fn register<'lua>(lua: &'lua Lua) -> anyhow::Result<()> {
...
+                    .cloned()
+                    .unwrap_or_default();
+
+                let reporting_info = if use_reporting {
+                    if let Ok(reporting_info) =
+                        config::from_lua_value::<ReportingInfo>(&lua, opt_reporting_info)
+                    {
+                        Some(reporting_info)
+                    } else {
+                        return Err(mlua::Error::external(RejectError {
```
wez, line 110: For the sake of being able to debug this, we definitely will want to log the error message from the `Err()` variant, otherwise this is a frustrating error message with no context!

TBH, I would just let that error get raised by mlua; I'll suggest a modification to the function definition above.

crates/kumod/src/dmarc.rs
```diff
@@ -20,10 +23,20 @@ pub fn register<'lua>(lua: &'lua Lua) -> anyhow::Result<()> {
...
+                spf_result,
+                opt_resolver_name,
+                opt_reporting_info,
+            ): (
                 UserDataRef<Message>,
+                bool,
+                mlua::Value,
                 mlua::Value,
                 Option<String>,
+                mlua::Value,
```
wez: Can have mlua process the type at function call dispatch time, and raise the serde error for you; `SerdeWrappedValue` is a bit of glue that does the equivalent of `config::from_lua_value`.

Can (and probably should!) use the same approach for the other parameters here!

```suggestion
                Option<SerdeWrappedValue<ReportingInfo>>,
```

crates/kumo-dmarc/src/lib.rs
sophiajt: Yeah, we need to do the type conversion step here

crates/kumo-dmarc/src/lib.rs
sophiajt: In theory, this should write one flushed output per error. But you're right that there are conditions under which this isn't thread-safe. Ultimately, I was thinking this would not live here, but it would live in a global error handler which itself would do the locking, so that we could be sure at a thread-level that each line is being done one at a time.

crates/kumo-dmarc/src/lib.rs
sophiajt: I honestly haven't a clue what the intent here is in the spec (as most of the xml reporting part of the spec doesn't go into much detail), but I'm happy to implement either way

crates/kumo-dmarc/src/lib.rs
sophiajt: Yes. I just wanted a placeholder for where it would go

crates/kumod/src/dmarc.rs
sophiajt: Not sure what you mean here. By earlier comment, do you mean the one about anyhow results?

sophiajt: Oh by earlier, does this connect to the suggestion made below?

crates/kumo-dmarc/src/types/results.rs
sophiajt: I've simplified it a bit. The previous String impl was there for the old code, and is no longer needed.

PR #470, opened 2026-01-16
adding audit-series module

kayozaki:
Adding module which keep track of an in-memory counter, which'll expire over time.

The main usage I'm looking at is to be able to track failures against an external resource.
For example, if RBL calls are timing out i'd want to be able to count the failures and fail-open for a set duration.
Later try again once the failure state expires.
I can call redis to maintain the counter, but redis itself can also go down, thus prefer something quick, local and light.


crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
+use config::{from_lua_value, get_or_create_sub_module};
+use lru_cache::LruCache;
+use lruttl::LruCacheWithTtl;
+use mlua::{Lua, Value};
+use serde::Deserialize;
+use std::sync::atomic::{AtomicIsize, Ordering};
+use std::sync::{Arc, LazyLock};
+use std::time::{SystemTime, UNIX_EPOCH};
+use tokio::sync::Mutex;
```
wez: in general, tokio's async Mutex should be avoided in favor of a regular Mutex, as it is more efficient and lower overhead to use it.  (although: prefer `parking_lot::Mutex` over the std Mutex).

The main reason to consider the async Mutex is where you need to do async operations while holding the lock, and that isn't needed here.

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+    LazyLock::new(|| Mutex::new(LruCacheWithTtl::new("audit_counters", 256)));
+
+static AUDIT_DEFINITION: LazyLock<Mutex<LruCache<String, AuditConfig>>> =
+    LazyLock::new(|| Mutex::new(LruCache::new(64)));
+
+#[derive(Deserialize, Clone)]
+pub struct AuditConfig {
+    pub bucket_count: usize,
+    #[serde(with = "duration_serde")]
+    pub window: Option<Duration>,
```
wez: I noticed that there's some error raising when the window is not configured or zero.  Can we remove the `Option` wrapper from here, so that it is impossible to omit it?  What's the use case for not specifying the window, and should it have a default value?

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+
+    Ok(())
+}
+
+pub fn register(lua: &Lua) -> anyhow::Result<()> {
+    let audit = get_or_create_sub_module(lua, "audit_series")?;
+
+    audit.set(
+        "define",
+        lua.create_async_function(|lua, (name, config): (String, Value)| async move {
```
wez: FWIW, you can use the `config::SerdeWrappedValue` wrapper to make function parameter types a little more declarative; you won't then need to do the explicit `from_lua_value` conversion

```suggestion
        lua.create_async_function(|lua, (name, config): (String, SerdeWrappedValue<AuditConfig>)| async move {
```

wez: switching away from the tokio mutex means that these lua functions no longer need to be async, which is good: non-async functions are a bit lower overhead and that is generally desirable for something like counter/metric manipulation!

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+fn duplicate_series_error(name: &str) -> mlua::Error {
+    mlua::Error::external(format!("Audit series '{name}' is already defined"))
+}
+
+fn missing_series_error(name: &str) -> mlua::Error {
+    mlua::Error::external(format!("Audit series '{name}' is not defined"))
+}
+
+fn current_epoch_secs() -> Result<u64, mlua::Error> {
+    SystemTime::now()
```
wez: `SystemTime` is relatively expensive to use in this use case; since we don't need to export the timestamp anywhere, my proposal is to do something like:

```rust
static START: LazyLock<Instant> = LazyLock::new(Instant::now);

fn elapsed_since_start_secs() -> u64 {
   START.elapsed().as_secs()
}
```

This should use the slightly cheaper monotonic clock under the covers.  This should work fine with the start time as the epoch, rather than the unix epoch, as we only care about understanding the delta from *an* epoch, rather than the unix epoch, in the math in this module.

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+        .map(|duration| duration.as_secs())
+}
+
+fn window_secs(window: Duration) -> Result<u64, mlua::Error> {
+    let secs = window.as_secs();
+    if secs == 0 {
+        return Err(mlua::Error::external(
+            "Audit series window must be at least one second",
+        ));
+    }
```
wez: I feel like this should be checked only at the time that the series is defined, in order to keep the overhead lower.

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+use lruttl::LruCacheWithTtl;
+use mlua::{Lua, Value};
+use serde::Deserialize;
+use std::sync::atomic::{AtomicIsize, Ordering};
+use std::sync::{Arc, LazyLock};
+use std::time::{SystemTime, UNIX_EPOCH};
+use tokio::sync::Mutex;
+use tokio::time::{Duration, Instant};
+
+static AUDIT_COUNTERS: LazyLock<Mutex<LruCacheWithTtl<String, Arc<AtomicIsize>>>> =
```
wez: The key to the cache map can be structured, and I'd recommend it, as it makes it easier to understand the shape of the data; it is an improvement over the current slightly adhoc string formatting approach, which has a collision issue (explained elsewhere in the review):

```rust
#[derive(Hash)]
struct AuditCounterKey {
   /// series name, matches AUDIT_DEFINITION key
   series_name: String,
   /// The key to use within that series for the counter
   key: String,
   bucket: u64,
}
```

```suggestion
static AUDIT_COUNTERS: LazyLock<Mutex<LruCacheWithTtl<AuditCounterKey, Arc<AtomicIsize>>>> =
```

With the key in that structure, I invite you look over https://github.com/sunshowers-code/borrow-complex-key-example to see how you can avoid making copies of the name via a special borrowed version of the counter key.  That's something to strongly consider if the number of buckets is large, as it can avoid memory allocation (and thus fragmentation) when performing the aggregate counter operations.

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+    LazyLock::new(|| Mutex::new(LruCacheWithTtl::new("audit_counters", 256)));
+
+static AUDIT_DEFINITION: LazyLock<Mutex<LruCache<String, AuditConfig>>> =
+    LazyLock::new(|| Mutex::new(LruCache::new(64)));
+
+#[derive(Deserialize, Clone)]
```
wez: Let's make this `Copy`; the rationale is that the current shape of `AuditConfig` can be simply a bit-copy when extracted from the cache map.  If adding eg: a `String` field occurs at a later date then we should change the cache map to hold `Arc<AuditConfig>` instead, so that we're not doing additional heap allocations for those strings.  Making the struct `Copy` will prevent unintentionally introducing those heap allocation overheads by preventing compilation, as `String` is not `Copy`.

```suggestion
#[derive(Deserialize, Clone, Copy)]
```

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+    let secs = window.as_secs();
+    if secs == 0 {
+        return Err(mlua::Error::external(
+            "Audit series window must be at least one second",
+        ));
+    }
+    Ok(secs)
+}
+
+fn current_bucket(window: Duration) -> Result<(u64, u64), mlua::Error> {
```
wez: Please define a struct for the return value; it's not easy to see the semantics of the returned tuple at a glance.  Something like this, but use your judgement for naming:

```rust
struct BucketInfo {
   /// The width of the window, measured in seconds
   window_secs: u64,
   /// The ordinal of the bucket
   bucket: u64,
}
```

```suggestion
fn current_bucket(window: Duration) -> Result<BucketInfo, mlua::Error> {
```

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+    pub bucket_count: usize,
+    #[serde(with = "duration_serde")]
+    pub window: Option<Duration>,
+}
+
+#[derive(Deserialize)]
+pub struct AuditInput {
+    pub key: String,
+    #[serde(default)]
+    pub count: isize,
```
wez: Should probably be unsigned? I don't think it makes sense to have negative counts?

```suggestion
    pub count: usize,
```

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+
+#[derive(Deserialize, Clone)]
+pub struct AuditConfig {
+    pub bucket_count: usize,
+    #[serde(with = "duration_serde")]
+    pub window: Option<Duration>,
+}
+
+#[derive(Deserialize)]
+pub struct AuditInput {
```
wez: Please add doc comments to the fields so that it is easier to intuit what is happening in the code without reading the corresponding lua docs!

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+    let window = config.window.ok_or_else(|| {
+        mlua::Error::external(format!("Audit series '{name}' window is not configured"))
+    })?;
+    let (bucket_epoch, window_secs) = current_bucket(window)?;
+
+    let counters = AUDIT_COUNTERS.lock().await;
+    let mut total = 0isize;
+    for offset in 0..config.bucket_count {
+        let bucket_start = bucket_epoch.saturating_sub(offset as u64 * window_secs);
+        let key = format!("{}:{}", input.key, bucket_start);
```
wez: with a structured key, this becomes a little simpler and less mathy.  Also, there is potential for collision with the current approach, as the `name` field is ignored. That means that two different series names with the same key will collide and have unpredictable values.  Including the name in the key will resolve that collision:

```suggestion
        let key = AuditCounterKey {
            series_name: name.clone(),
            key: input.key.clone(),
            bucket: bucket_epoch.saturating_sub(offset),
        };
```

docs/reference/kumo.audit_series/add.md
```diff
@@ -0,0 +1,31 @@
...
+
+`key` : key name to track counters for
+
+`count` : value to increment or use negative number to decrement
+
+Returned value represents the value in the current bucket. Use the get function to obtain the total of all buckets.
+
+Error is generated if audit series name is not yet defined.
+
+```
```
wez: Use `lua` for the fenced code block examples

````suggestion
```lua
````

File: crates/mod-audit-series/src/lib.rs [hunk 13]
```diff
@@ -0,0 +1,328 @@
...
+            key,
+            counter.clone(),
+            Instant::now() + Duration::from_secs(ttl_secs),
+        )
+        .await;
+
+    Ok(counter.load(Ordering::Relaxed))
+}
+
+pub async fn reset_audit_series(name: &str, input: AuditInput) -> Result<(), mlua::Error> {
```
wez: We don't need an `AuditInput` to be passed in here, let's just pass the key that we want to zero out instead.

```suggestion
pub async fn reset_audit_series(name: &str, key: &str) -> Result<(), mlua::Error> {
```

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+        let key = format!("{}:{}", input.key, bucket_start);
+        if let Some(counter) = counters.get(&key) {
+            total += counter.load(Ordering::Relaxed);
+        }
+    }
+
+    Ok(total)
+}
+
+pub async fn add_audit_series_count(name: &str, input: AuditInput) -> Result<isize, mlua::Error> {
```
wez: Let's pass the key and delta more directly and explicitly; `AuditInput` feels too generic for this and the potential for future scope growth of that type will make it harder to understand and maintain the code over time.

```suggestion
pub async fn add_audit_series_count(name: &str, key: &str, count: isize) -> Result<isize, mlua::Error> {
```

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+pub async fn define_audit_series(name: &str, config: AuditConfig) -> Result<(), mlua::Error> {
+    let mut cache = AUDIT_DEFINITION.lock().await;
+    if cache.contains_key(name) {
+        return Err(duplicate_series_error(name));
+    }
+    cache.insert(name.to_string(), config);
+    Ok(())
+}
+
+pub async fn get_audit_series_total(name: &str, input: AuditInput) -> Result<isize, mlua::Error> {
```
wez: Let's just pass the key name, rather than `AuditInput` here:

```suggestion
pub async fn get_audit_series_total(name: &str, key: &str) -> Result<isize, mlua::Error> {
```

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+        })?,
+    )?;
+
+    // Accumulate all buckets and return the total
+    audit.set(
+        "get",
+        lua.create_async_function(|lua, (name, input): (String, Value)| async move {
+            let input: AuditInput = from_lua_value(&lua, input)?;
+            get_audit_series_total(&name, input).await
+        })?,
```
wez: Feels simpler and cleaner to just pass the key name directly, rather than wrap it up in an options table.
Are there other parameters that you're likely to want to add here in the future?

```suggestion
        lua.create_function(move |_lua, (name, key): (String, String)|  {
            get_audit_series_total(&name, &key)
        })?,
```

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+    )?;
+
+    // Add to the current bucket and return the existing value
+    audit.set(
+        "add",
+        lua.create_async_function(|lua, (name, input): (String, Value)| async move {
```
wez: Similar comments to above; can we just directly pass the key and counter value here?

crates/mod-audit-series/src/lib.rs
```diff
@@ -0,0 +1,328 @@
...
+            let input: AuditInput = from_lua_value(&lua, input)?;
+            add_audit_series_count(&name, input).await
+        })?,
+    )?;
+
+    // Reset the values of given audit name to 0 for all buckets
+    audit.set(
+        "reset",
+        lua.create_async_function(|lua, (name, input): (String, Value)| async move {
```
wez: Can just pass the key directly here also, to simplify this interface

docs/reference/kumo.audit_series/_index.md
```diff
@@ -0,0 +1,12 @@
+---
+tags:
+ - audit_series
+---
+
+{{since('dev')}}
+
+# Audit_series
+
+Audit_series maintains an in-memory chronological set of metrics of events. Metrics are not persisted across process restarts and are only meant for short-term bookkeeping of events. If persistence or data sharing across multiple kumomta instances is important, try using external database such as redis for storage.
```
wez:
```suggestion
Audit_series maintains an in-memory chronological set of metrics of events. Metrics are not persisted across process restarts, nor shared between nodes, and are only meant for short-term bookkeeping of events in the current process instance. If persistence or data sharing across multiple kumomta instances is important, try using external database such as redis for storage.
```

wez (commented):
I can appreciate the feature.  I have a number of comments here!

One thing not addressed inline is that the use of a lru cache to hold bucket counts makes this a lossy mechanism, which has the potential to feel non-deterministic depending on the traffic patterns.

For example, if a bucket is evicted to make room for a newly bumped bucket elsewhere, than can cause a value to effectively zero out for the evicted bucket.
If that bucket was outside of the defined window then it is fine, but if not, suddenly your counter totals are impacted in surprising ways.

What are you thoughts on this?

At the very least, the lossy nature and the cache size tuning should be discussed in the docs for the `define` function, as I think that it would be surprising and a source of support issues not to do so.

crates/mod-audit-series/src/lib.rs
kayozaki: I could probably workaround by keeping track of success and failure with two series name.
I want to support negative so we can decrement the failure count when we have successful transaction.
Then increment when we have failed transaction. If overall count exceeds certain threshold of failure (as in repeated failures) then only do fail open.

wez:
I've been sitting on this for a bit because a couple of things were on my mind about it:

 * I felt that losing an internal bucket could make reasoning about the current value very unreliable and be a source of very complicated support questions from folks that were taken surprise by it in the future
 * Spreading the buckets out to individual cache entries is relatively expensive because each bucket access turns into a fairly complex (and potentially contended) access operation.  While it is technically an `O(1)` access, there is a decent amount going on to manage that cache.
 * There can be a large burden on the cache from the older, aged out buckets; while they will eventually be evicted, it's  wasteful of RAM while they await eviction, and that increases heap pressure and may lead to excess fragmentation
 * The tests in this PR do a lot of the heavy lifting in lua, which is OK for testing the lua interface, but it's a lot of effort to write and maintain for the lower level stuff, which I really want to be done in Rust.

What I propose is this:

* Over in https://github.com/KumoCorp/kumomta/pull/475 I have sketched out a `CounterSeries` type that implements the underlying audit series as a native and self-contained Rust type with unit tests that can deterministically exercise the time sensitive code without writing a long running or potentially flaky test.  The memory overhead of this type is small and bounded, regardless of the time intervals between operating on the series
* Assuming that the `CounterSeries` can be employed in your intended use case, I propose that I land it and that you adjust this PR so that the audio-series module creates a cache mapping series names to `Arc<Mutex<CounterSeries>>` and uses that to implement the interface.
* The lua tests here then just need to exercise the lua bindings and not need to do any of the time dependent stuff as that is covered by the unit tests for that module

What do you think about that?

Issue #462, opened 2025-12-03
Add more detailed logging to KumoProxy


tommairs:
Currently, the logging output of KumoProxy is sparse.  While you can configure it as a Systemd service and capture the journal, this is more system-level error and debug logging as opposed to process informational logging.  As an operator, it would be useful to see the logging of which IP was selected and routed from and to (the path).  It would also be helpful to see the number of open connections at any time.  

A useful log might include information like:
<timestamp>, <origin_ip>, <proxy_ip>, <destination_ip>, <open_connection_count>, <message_id> (or other identifiers)

PR #472, opened 2026-01-30
Add Prometheus metrics to proxy-server

AdityaAudi:
## Summary

This PR enhances KumoProxy's observability by adding Prometheus-compatible metrics for monitoring connections and traffic, exposed via an HTTP endpoint.

## Motivation (https://github.com/KumoCorp/kumomta/issues/462)

Previously, KumoProxy had limited observability, making it difficult for operators to:
- Monitor the number of open connections in real-time
- Understand connection failures and their causes
- Measure traffic volume through the proxy
- Track outbound connection patterns

## Prometheus Metrics

Added 8 metrics labeled by listener address:

| Metric | Type | Description |
|--------|------|-------------|
| proxy_connections_accepted_total | counter | Total incoming connections accepted |
| proxy_connections_failed_total | counter | Connections that failed during handshake or proxying |
| proxy_connections_completed_total | counter | Proxy sessions that completed successfully |
| proxy_active_connections | gauge | Current number of active connections |
| proxy_bytes_received_total | counter | Total bytes received from clients |
| proxy_bytes_sent_total | counter | Total bytes sent to clients |
| proxy_outbound_connections_total | counter | Outbound connections (labeled by destination) |
| proxy_tls_handshake_failures_total | counter | TLS handshake failures |

*Note:* proxy_outbound_connections_total uses PruningCounterRegistry to mitigate high-cardinality issues from the destination label.

## HTTP Metrics Endpoint

Added new Lua function proxy.start_http_listener() exposing:

- /metrics - Prometheus text format
- /metrics.json - JSON format  
- /proxy/status - Health check endpoint
- /api/admin/set_diagnostic_log_filter/v1 - Runtime log filter adjustment
- /rapidoc - Interactive API documentation
- /api-docs/openapi.json - OpenAPI specification

OpenAPI spec can also be generated via CLI: proxy-server --dump-openapi-spec

## Configuration Example

```
local kumo = require 'kumo'
local proxy = require 'proxy'

kumo.on('proxy_init', function()
  -- Start SOCKS5 proxy listener
  kumo.start_proxy_listener {
    listen = '0.0.0.0:1080',
    timeout = '60 seconds',
  }
  
  -- Start HTTP listener for metrics
  proxy.start_http_listener {
    listen = '0.0.0.0:8080',
    trusted_hosts = { '127.0.0.1', '::1' },
  }
```


crates/proxy-server/src/proxy_handler.rs
```diff
@@ -9,6 +13,89 @@ use std::net::{IpAddr, SocketAddr};
 use tokio::io::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
 use tokio::net::{TcpSocket, TcpStream};
 use tokio::time::timeout;
+use uuid::Uuid;
+
+/// Connection info for logging purposes
+#[derive(Clone, Debug)]
+pub struct ConnectionInfo {
+    /// Unique identifier for this connection
+    pub connection_id: String,
```
wez: Using a uuid directly uses less storage than the string representation of a uuid, and doesn't require a heap allocation like the underlying Vec storage for the String.

```suggestion
    pub connection_id: Uuid,
```

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -181,14 +196,51 @@ declare_event! {
...
-            params.start().await.map_err(any_err)?;
+            if !config::is_validating() {
+                params.start().await.map_err(any_err)?;
+            }
+            Ok(())
+        })?,
+    )?;
+
+    kumo_mod.set(
+        "start_proxy_http_listener",
```
wez: Probably this should be `proxy.start_http_listener` to mirror what we're doing with https://docs.kumomta.com/reference/tsa/start_http_listener/

We can achieve that here by changing line 221 above to use `proxy` instead of `kumo`.

Similarly, the `start_proxy_listener` function above should be moved to that namespace.

Since we haven't released a stable version of kumo with the proxy stuff exposed to lua, it is OK to move it around as part of this PR without going to great lengths to preserve backwards compatibility.

@vietcgi: heads up that we're probably going to move this around.

Moving `start_proxy_listener` should probably be its own PR to keep things generally a bit smaller and easier to reason about.  We can then layer this one on top of that.

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -181,14 +196,51 @@ declare_event! {
     ) -> SerdeWrappedValue<AuthKindResult>;
 }
 
+/// Create a simple router for the proxy metrics HTTP endpoint
+fn make_proxy_router() -> RouterAndDocs {
```
wez: HTTP registration was changed quite a lot recently, to make it a bit better factored and ensure that we don't miss any docs.

https://github.com/KumoCorp/kumomta/commit/a8163b6e48199e2d95a8c19b6b2afd73810b5ee8

Here's how tsa registers its handlers now:

https://github.com/KumoCorp/kumomta/blob/f53ed1aeae6f270ec6fff50889af85b30536163a/crates/tsa-daemon/src/http_server.rs#L43-L55

Each handler is required to use the `utoipa::path` annotation which defines the path and ensures that we get docs in emitted to the jsonschema.

Adding something like this:

https://github.com/KumoCorp/kumomta/blob/f53ed1aeae6f270ec6fff50889af85b30536163a/crates/tsa-daemon/src/main.rs#L50-L54

will allow extracting the json schema.  We could then add those to the docs.

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -181,14 +196,51 @@ declare_event! {
...
+    RouterAndDocs {
+        router,
+        docs: ApiDoc::openapi(),
+    }
+}
+
+/// Simple status endpoint for the proxy
+async fn proxy_status() -> &'static str {
+    "KumoProxy OK"
+}
```
wez, line 248: I think its worth discussing what we think that a good status response might be; is there a lot of value to always returning a static string?

crates/proxy-server/src/proxy_handler.rs
```diff
@@ -9,6 +13,89 @@ use std::net::{IpAddr, SocketAddr};
...
+    /// Client/origin IP address
+    pub origin_ip: SocketAddr,
+    /// Proxy listener address
+    pub proxy_ip: SocketAddr,
+    /// Destination address (populated after CONNECT)
+    pub destination_ip: Option<SocketAddr>,
+    /// Bound source address for outbound connection
+    pub source_ip: Option<SocketAddr>,
+    /// Authenticated username, if any
+    pub username: Option<String>,
```
wez: Identity can be a lot more nuanced than a simple username; `AuthInfo` is the definitive struct we use for this concept and that is probably what we should use in logs.

crates/proxy-server/src/proxy_handler.rs
```diff
@@ -9,6 +13,89 @@ use std::net::{IpAddr, SocketAddr};
...
+            origin_ip: peer_address,
+            proxy_ip: local_address,
+            destination_ip: None,
+            source_ip: None,
+            username: None,
+        }
+    }
+
+    fn log_connect(&self, result: &str) {
+        let timestamp = Utc::now().to_rfc3339();
```
wez: This timestamp is redundant with the timestamp that is emitted by the `tracing` crate

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -111,19 +114,31 @@ impl ProxyListenerParams {
...
             .local_addr()
             .context("failed to get local address")?;
+        let listener_label = local_address.to_string();
 
         loop {
             let (socket, peer_address) = listener.accept().await.context("accept failed")?;
 
             let params = params.clone();
             let tls_acceptor = tls_acceptor.clone();
+            let listener_label_clone = listener_label.clone();
```
wez: this will keep a string copy of the local_address alive for the duration of the proxy session, which is wasteful.  We already capture the local address; we can turn this into a string only when needed in the error case.

docs/reference/kumo/start_proxy_http_listener/_index.md
```diff
@@ -0,0 +1,89 @@
...
+
+The HTTP listener provides the following endpoints:
+
+| Endpoint | Description |
+|----------|-------------|
+| `/metrics` | Prometheus text format metrics |
+| `/metrics.json` | JSON format metrics |
+| `/proxy/status` | Simple health check endpoint |
+| `/api/admin/set_diagnostic_log_filter/v1` | Change diagnostic log filter |
+| `/rapidoc` | Interactive API documentation |
```
wez: This can be replaced by hooking up the json schema to the doc build.

See:

https://github.com/KumoCorp/kumomta/blob/f53ed1aeae6f270ec6fff50889af85b30536163a/docs/update-openapi.sh#L37

https://github.com/KumoCorp/kumomta/blob/f53ed1aeae6f270ec6fff50889af85b30536163a/crates/jsonschematodocs/src/main.rs#L828

https://github.com/KumoCorp/kumomta/blob/f53ed1aeae6f270ec6fff50889af85b30536163a/docs/generate-toc.py#L573

docs/reference/kumo/start_proxy_http_listener/_index.md
```diff
@@ -0,0 +1,89 @@
...
+| `proxy_outbound_connections_total` | Counter | Total outbound connections by destination |
+| `proxy_tls_handshake_failures_total` | Counter | Total TLS handshake failures (when TLS is enabled) |
+
+All metrics are labeled with `listener` (the listener address). The `proxy_outbound_connections_total` metric is additionally labeled with `destination`.
+
+!!! warning "High Cardinality Metric"
+    The `proxy_outbound_connections_total` metric tracks connections by destination IP address.
+    If your proxy connects to many unique destinations, this can create high cardinality
+    in Prometheus, which may impact memory usage and query performance. This metric uses
+    a pruning counter registry to mitigate this, but monitor cardinality in production environments.
```
wez: We can do something smarter for this in the doc build, but it will require a bit of effort.

https://github.com/KumoCorp/kumomta/commit/d0278947ed7ac04a0fa7483aac8db32de2ea686c made it so that we pull in a sample set of metrics for the shared metrics endpoint.

That sample is currently specific to kumod.  We could make the sample service-specific by adjusting the include statement as part of the processing here:

https://github.com/KumoCorp/kumomta/blob/f53ed1aeae6f270ec6fff50889af85b30536163a/crates/jsonschematodocs/src/main.rs#L140-L143

but have it rewrite this include line:

https://github.com/KumoCorp/kumomta/blob/f53ed1aeae6f270ec6fff50889af85b30536163a/crates/kumo-server-common/src/http_server/mod.rs#L517

by adding in the service name from the current set of docs to the path that it includes.

crates/proxy-server/src/proxy_handler.rs
```diff
@@ -112,17 +256,59 @@ where
...
+                    session_metrics.record_bytes_sent(*to_client as usize);
+                    session_metrics.record_bytes_received(*to_remote as usize);
+                    conn_info.log_disconnect(*to_client, *to_remote, duration);
+                    session_metrics.mark_completed();
+                }
+                Err(_) => {
+                    conn_info.log_disconnect(0, 0, duration);
+                    session_metrics.mark_failed();
+                }
+            }
```
wez: Since this code is (or should be!) identical to the same block above, I feel that we should refactor this into a function that we apply to both cases to reduce the duplicate code and make things more robust to future refactoring.

crates/proxy-server/src/proxy_handler.rs
```diff
@@ -9,6 +13,89 @@ use std::net::{IpAddr, SocketAddr};
...
+    fn log_connect(&self, result: &str) {
+        let timestamp = Utc::now().to_rfc3339();
+        let dest = self
+            .destination_ip
+            .map(|d| d.to_string())
+            .unwrap_or_else(|| "-".to_string());
+        let source = self
+            .source_ip
+            .map(|s| s.to_string())
+            .unwrap_or_else(|| "-".to_string());
```
wez: Do these need to be explicitly converted to strings?  Can the normal Display or Debug impl's be used for these in the logs?

crates/proxy-server/src/metrics.rs
```diff
@@ -0,0 +1,182 @@
...
+        destination,
+    };
+    OUTBOUND_CONNECTIONS_TOTAL.get_or_create(&key as &dyn ConnectionKeyTrait)
+}
+
+/// Metrics bundle for a single proxy session
+#[derive(Clone)]
+pub struct ProxySessionMetrics {
+    #[allow(dead_code)]
+    listener: String,
```
wez: Can we eliminate this? It's a high cardinality heap allocation and if it is not used by anything, it's just wasting memory and increasing heap fragmentation

crates/proxy-server/src/proxy_handler.rs
```diff
@@ -9,6 +13,89 @@ use std::net::{IpAddr, SocketAddr};
...
+        let user = self.username.as_deref().unwrap_or("-");
+
+        tracing::info!(
```
wez: Logging this way potentially forms an implicit contract on the format of the stderr of this service, which feels undesirable as it will limit the changes that we can make in the future if we're not explicit about the stability and machine readability of these logs.

Is your intent to parse these logs in some way?  Should there be an explicit and separate log file for this information?

I feel like the logging part of this PR is a separate (although thematically related) concern and should probably be split into a separate PR

docs/reference/kumo/start_proxy_http_listener/_index.md
wez: An update on this: I made some improvements to how we register metrics and we're now able to extract metadata and docs for all of them, and put them into the docs on a per-service basis.  https://docs.kumomta.com/reference/metrics/kumod/ is where kumod's metrics are documented.

Using the new `declare_metric!` macro for the metrics you're adding in this PR will enable extracting the proxy server metrics in a similar way.

crates/proxy-server/src/proxy_handler.rs
AdityaAudi: This comment related to the structured logging feature which has been removed from this PR. The logging features (including connection_id, ConnectionInfo struct, etc.) will be addressed in a follow-up PR where we can properly discuss log format stability. The uuid dependency has been removed from Cargo.toml.

crates/proxy-server/src/mod_proxy.rs
AdityaAudi: Following your suggestion to keep things smaller and in separate PRs:
- kumo.start_proxy_listener - unchanged in this PR (will move to proxy.start_listener in a separate PR as you suggested)
- kumo.start_proxy_http_listener → proxy.start_http_listener - moved in this PR

This keeps the scope focused on adding the new metrics HTTP endpoint without changing the existing SOCKS5 listener API.

The kumo.start_proxy_listener API remains unchanged in this PR. Only the HTTP metrics listener is new.

crates/proxy-server/src/mod_proxy.rs
AdityaAudi: Updated to use proper utoipa path annotations:

1. Added `#[derive(OpenApi)]` with paths(proxy_status) 
2. Added `#[utoipa::path(...)] `annotation to proxy_status endpoint
3. Made `make_router()` public for OpenAPI spec extraction
4. Added -`-dump-openapi-spec` CLI flag

The OpenAPI spec can now be extracted via:
`proxy-server --dump-openapi-spec > proxy-server.openapi.json`
The handler is registered following the same pattern as tsa-daemon.

crates/proxy-server/src/mod_proxy.rs
AdityaAudi: Good point about discussing the value of a static string response. For a simple health check endpoint (used by load balancers, k8s probes, etc.), the static 200 OK is typically sufficient and has zero runtime overhead. It confirms the HTTP listener is running and the process is responsive. For detailed operational data, operators already have:

- `/metrics` - Full Prometheus metrics with active connections, bytes transferred, etc.
- `/metrics.json` - Same data in JSON format

I'm inclined to keep `/proxy/status` minimal for health checks since `/metrics` provides comprehensive operational visibility. However, if you prefer a more informative response (e.g., JSON with active connection count, uptime), I'm happy to add that. What's your preference?

crates/proxy-server/src/proxy_handler.rs
AdityaAudi: This comment related to the structured logging feature which has been removed from this PR. The authentication code still properly uses and returns AuthInfo internally (from perform_auth()), but it's not logged in this PR. The proper use of AuthInfo for logging will be addressed in the follow-up logging PR

crates/proxy-server/src/proxy_handler.rs
AdityaAudi: This comment related to the structured logging feature which has been removed from this PR. The `chrono` dependency has been removed from `Cargo.toml`.

crates/proxy-server/src/mod_proxy.rs
AdityaAudi: Removed the pre-emptive string clone. The local_address is now stored as SocketAddr and only converted to string in the TLS handshake error path:
```
// Only convert to string in the error case
crate::metrics::tls_handshake_failures_for_listener(&local_address.to_string())
```
This avoids a heap allocation per connection for the success path.

docs/reference/kumo/start_proxy_http_listener/_index.md
AdityaAudi: Removed the manual endpoint table. Documentation now points to:

- `/api-docs/openapi.json` - OpenAPI specification
- `/rapidoc` - Interactive API documentation
- `--dump-openapi-spec` CLI flag for static extraction

This can be integrated into the doc build process similar to tsa-daemon (update-openapi.sh, jsonschematodocs) once the hooks are set up.

docs/reference/kumo/start_proxy_http_listener/_index.md
AdityaAudi: Understood - I'll update to use declare_metric! once the macro is available in the codebase.
Current implementation uses the existing `CounterRegistry::register` and `PruningCounterRegistry::register` patterns which work correctly. Once your recent metric improvements land, I can convert these to use the new macro for automatic documentation extraction.

crates/proxy-server/src/proxy_handler.rs
AdityaAudi: Extracted the duplicate code into a `finalize_session_metrics()`
Now used by both the `splice(2)` and `copy_bidirectional` code paths for consistent metric handling.

crates/proxy-server/src/proxy_handler.rs
AdityaAudi: This comment related to the structured logging feature which has been removed from this PR. The explicit string conversions for IPs were part of the logging code. This will be addressed properly (using ? Debug formatting) in the follow-up logging PR

crates/proxy-server/src/metrics.rs
AdityaAudi: Removed the unused `listener: String` field from `ProxySessionMetrics`. This eliminates the unnecessary heap allocation per session.

crates/proxy-server/src/proxy_handler.rs
AdityaAudi: Agreed, splitting makes sense. I've removed all structured logging from this PR to keep it focused on metrics only.

crates/proxy-server/src/metrics.rs
```diff
@@ -0,0 +1,184 @@
...
+}
+
+/// Total number of incoming connections accepted (counter)
+pub static TOTAL_CONNECTIONS_ACCEPTED: LazyLock<CounterRegistry<ListenerKey>> =
+    LazyLock::new(|| {
+        CounterRegistry::register(
+            "proxy_connections_accepted_total",
+            "total number of incoming connections accepted by the proxy".to_string(),
+        )
+    });
```
wez: You can already make use of the new declare_metric! macro like this:

```suggestion
declare_metric! {
/// total number of incoming connections accepted by the proxy
pub static TOTAL_CONNECTIONS_ACCEPTED: CounterRegistry<ListenerKey>(
            "proxy_connections_accepted_total");
}
```

More examples of migrating from the old to the new style can be found in https://github.com/KumoCorp/kumomta/commit/dd41e999c49a8018268a40cc9a2d490b142d7349

crates/proxy-server/src/proxy_handler.rs
```diff
@@ -24,14 +45,32 @@ pub async fn handle_proxy_client<S>(
...
+    // Track connection acceptance
+    connections_accepted_for_listener(&listener_label).inc();
+
+    // Create session metrics (increments active connections gauge)
+    let session_metrics = ProxySessionMetrics::new(&listener_label);
+
     let mut state = ClientState::None;
 
-    let handshake = timeout(timeout_duration, async {
+    let handshake = match timeout(timeout_duration, async {
```
wez: a lot of the work inside this function is to capture the session status into metrics.  I think it will probably be a lot cleaner, simpler, and a smaller diff, if we shift this logic up to the `handle_client` function that calls this one, so it might look something like:

```rust
    async fn handle_client<S>(
        stream: S,
        peer_address: SocketAddr,
        local_address: SocketAddr,
        params: &ProxyListenerParams,
    ) -> anyhow::Result<()>
    where
        S: AsyncReadAndWrite + Unpin + Send + 'static,
    {
        let result = crate::proxy_handler::handle_proxy_client(
            stream,
            peer_address,
            local_address,
            params.timeout,
            params.use_splice,
            params.require_auth,
        )
        .await;

        if result.is_ok() {
            // update counters for success
        } else {
            // update counters for error
        }

        result
    }
```

You might also consider moving the `ProxySessionMetrics` instantiation up to this level and passing it down to `handle_proxy_client` to centralize that accounting logic to `handle_client`?

Similarly, we could redefine the successful return value from `handle_proxy_client` so that it returns the total bytes read/written from the splice/copy calls so that the totals can be updated at the `handle_client` level.

crates/proxy-server/src/proxy_handler.rs
```diff
@@ -24,14 +45,32 @@ pub async fn handle_proxy_client<S>(
 where
     S: AsyncReadAndWrite + Unpin + Send + 'static,
 {
+    let listener_label = local_address.to_string();
```
wez: this string will be live until the end of this function scope, and it would be nice for it not to live so long.   Perhaps the simplest way to deal with this is to change the type of the parameter in the `connections_accepted_for_listener` function to accept `SocketAddr` and have it format the string on-demand inside its own impl.

Similar sentiment for the `ProxySessionMetrics::new` call.

docs/reference/kumo/start_proxy_http_listener/_index.md
wez: `declare_metrics!` is available already! It should be straightforward to migrate to it.  The only thing to take care about is that where you're using `register_gauge` at the moment, you need to use `PruningGaugeRegistry` as the type of the metric.  I've made a comment inline in the code that should point you in the right direction.

crates/proxy-server/src/proxy_handler.rs
AdityaAudi: I've refactored the code to:
- Instantiate `ProxySessionMetrics` in `handle_client` (centralized location)
- Have `handle_proxy_client` return `anyhow::Result<ProxySessionResult>` with byte counts
- Update success/failure metrics in `handle_client` based on the result

This centralizes all metric accounting logic and simplifies `handle_proxy_client`, making it easier to reason about.

crates/proxy-server/src/metrics.rs
AdityaAudi: Great suggestion! I've refactored all metrics to use the declare_metric! macro. All metrics in metrics.rs now follow the same pattern as kumod/src/metrics_helper.rs.

crates/proxy-server/src/proxy_handler.rs
AdityaAudi: You're absolutely right about the string allocations. I've refactored the helper functions to accept `SocketAddr` directly and convert to string inline where needed. This matches the pattern used in `kumod/src/queue/metrics.rs` with `BorrowedQueueKey`.The string is now created on-demand and immediately borrowed, reducing heap allocations and lifetime complexity.

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -181,14 +225,60 @@ declare_event! {
...
+)]
+struct ApiDoc;
+
+/// Create the router for the proxy HTTP endpoint
+pub fn make_router() -> RouterAndDocs {
+    RouterAndDocs {
+        router: Router::new().route("/proxy/status", get(proxy_status)),
+        docs: ApiDoc::openapi(),
+    }
+}
```
wez: Let's use the newer registration stuff for this:

```suggestion
/// Create the router for the proxy HTTP endpoint
pub fn make_router() -> RouterAndDocs {
    router_with_docs! {
       title = "proxy-server",
       handlers = [
          proxy_status
       ]
    }
}
```

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -147,15 +167,39 @@ impl ProxyListenerParams {
...
+                );
+                session_metrics.mark_completed();
+            }
+            Err(_) => {
+                session_metrics.mark_failed();
+            }
+        }
+
+        // Convert the result, discarding the byte counts (already recorded)
+        result.map(|_| ())
```
wez:
```suggestion
        // Update metrics based on result
        match result {
            Ok(session_result) => {
                session_metrics.record_bytes(
                    session_result.bytes_to_remote,
                    session_result.bytes_to_client,
                );
                session_metrics.mark_completed();
                Ok(())
            }
            Err(err) => {
                session_metrics.mark_failed();
                Err(err)
            }
        }
```

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -11,6 +17,7 @@ use std::sync::Arc;
 use std::time::Duration;
 use tokio::net::TcpListener;
 use tokio_rustls::TlsAcceptor;
+use utoipa::OpenApi;
```
wez, line 19: Can probably remove this when using the newer registration stuff

```suggestion
```

crates/proxy-server/src/metrics.rs
```diff
@@ -0,0 +1,219 @@
...
+    pub fn new(listener: SocketAddr) -> Self {
+        let active = active_connections_for_listener(listener);
+        active.inc();
+
+        Self {
+            active_connections: active,
+            connections_completed: connections_completed_for_listener(listener),
+            connections_failed: connections_failed_for_listener(listener),
+            bytes_received: bytes_received_for_listener(listener),
+            bytes_sent: bytes_sent_for_listener(listener),
```
wez: Can we inline these various `_for_listener` functions (including `active_connections_for_listener` above) here?

The rationale is that we can then just do a single `listener.to_string()` to create a single instance of the `BorrowedListenerKey` in here, and this appears to be the only place where we need to grab these particular counters, so we don't really need to export them from this metrics module.

crates/proxy-server/src/metrics.rs
```diff
@@ -0,0 +1,219 @@
...
+);
+}
+
+declare_metric! {
+/// Total bytes received from clients.
+///
+/// This counter tracks the total number of bytes received from proxy clients
+/// (i.e., data flowing from client to destination).
+pub static BYTES_RECEIVED: CounterRegistry<ListenerKey>(
+    "proxy_bytes_received_total"
```
wez: What do you think about naming this something like `proxy_bytes_client_to_dest` to make it clearer which way the bytes are flowing?  I think words like received/sent/ingress/egress are essentially a binary operator and without indicating both operands in the name, it is ambiguous how to interpret this.  For instance, my natural inclination would be to consider this particular metric as sending bytes rather than receiving, but it's all dependent upon your perspective!

crates/proxy-server/src/metrics.rs
```diff
@@ -0,0 +1,219 @@
...
+);
+}
+
+declare_metric! {
+/// Total bytes sent to clients.
+///
+/// This counter tracks the total number of bytes sent to proxy clients
+/// (i.e., data flowing from destination to client).
+pub static BYTES_SENT: CounterRegistry<ListenerKey>(
+    "proxy_bytes_sent_total"
```
wez: per above, how about `proxy_bytes_dest_to_client` for this one?

crates/proxy-server/src/mod_proxy.rs
AdityaAudi: I think we still need it. This is the same pattern used in `kumod`

crates/proxy-server/src/metrics.rs
AdityaAudi: Thanks for the suggestion! I've inlined all the helper functions into ProxySessionMetrics::new(). Now we:
- Call listener.to_string() only once
- Create a single BorrowedListenerKey and reuse it for all metric lookups
- Removed the helper functions since ProxySessionMetrics::new() is the only call site
- Kept only connections_accepted_for_listener, tls_handshake_failures_for_listener, and outbound_connections_for as public helpers since they're called from mod_proxy.rs

This reduces string allocations from 5 to 1 per session.

crates/proxy-server/src/metrics.rs
AdityaAudi: Great point about ambiguity! I've renamed the metrics to be more explicit about the direction:
- `proxy_bytes_received_total` → `proxy_bytes_client_to_dest_total` (upstream)
- `proxy_bytes_sent_total` → `proxy_bytes_dest_to_client_total` (downstream)

This makes it crystal clear which direction the bytes are flowing, removing any perspective-dependent interpretation. Updated the documentation in both the code comments and `docs/reference/proxy/start_http_listener.md` to reflect the new names.

PR #475, opened 2026-02-05
introduce kumo-counter-series crate

wez:
This crate implements an in memory ring-buffer based counter series.

It is intended to be used for bounded (and ideally small!) counter time series.

The idea is that this should be the backing store employed by the proposed https://github.com/KumoCorp/kumomta/pull/470

### DIFF
--- a/crates/kumo-counter-series/src/lib.rs
+++ b/crates/kumo-counter-series/src/lib.rs
@@ -1,0 +1,361 @@
+use tokio::time::{Duration, Instant};
+
+pub struct CounterSeriesConfig {
+    /// How many buckets should be maintained.
+    pub num_buckets: u8,
+    /// How long a time interval each bucket represents,
+    /// expressed in seconds.
+    pub bucket_size: u64,
+}
+
+/// CounterSeries implements a time series stored in a
+/// sequence of a fixed number of buckets each with a fixed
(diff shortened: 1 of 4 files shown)


crates/kumo-counter-series/src/lib.rs
```diff
@@ -0,0 +1,361 @@
...
+
+        if elapsed_slots > 0 {
+            let num_prune = elapsed_slots.min(num_buckets) as isize;
+            self.curr_bucket = ((elapsed_slots + self.curr_bucket as u64) % num_buckets) as u8;
+            // we updated curr_bucket, so revise the updated time
+            self.updated = Instant::now();
+
+            for prune in 0..num_prune {
+                let mut idx = (self.curr_bucket as isize) - prune;
+                if idx < 0 {
```
wez, line 87: something insightful

wez: I am replying to myself. This is a test.

crates/kumo-counter-series/src/lib.rs
```diff
@@ -0,0 +1,361 @@
+use tokio::time::{Duration, Instant};
+
+pub struct CounterSeriesConfig {
```
wez, line 3: nice struct!

PR #477, opened 2026-02-06
Update community shaping.toml to use providers instead of site names

danielsmithsd:
This PR seeks to standardize the behavior of the traffic shaping system inside a provider instead of calling out site names individually.

The only concern I have is that the provider definitions are separate from the individual community rules.  I'm wondering if those should be in the same file or if it makes sense for them to be separated?

### DIFF
--- a/assets/community/shaping.toml
+++ b/assets/community/shaping.toml
@@ -192,31 +192,26 @@ trigger = {Threshold="5/hr"}
 action = "Suspend"
 duration = "1 hour"
 
-#===== orange.fr
-["orange.fr"]
-max_deliveries_per_connection = 75
-connection_limit = 2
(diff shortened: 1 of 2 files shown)


assets/community/shaping.toml
```diff
@@ -29,78 +29,78 @@
 #### BEGIN COMMUNITY CONTRIBUTED SHAPING RULES
 
 #===== GMAIL
-[["gmail.com".automation]]
+[[provider."gmail".automation]]
```
dschaaff, line 32: Is this supposed to be `[[provider."google".automation]]`?

danielsmithsd: Yes.  Good catch

PR #481, opened 2026-02-15
Add Delayed option to Rejection settings

jack-atlas:
For this : KumoCorp#478

Don't feed delayed data to the daemon; we're only interested in data that flows back to us from after the point of delayed.


wez:
Thanks for this; while looking into this, I don't think the magnitude of this issue would account for what you say in the related issue, as we're still excluding things before we put them into the spool.

I've got an alternative implementation of this PR that I will push in just a moment, so I think we can close this one.

PR #496, opened 2026-03-18
load intermediate cert

kayozaki:
"X509::from_pem" doesn't load the intermediate certificate in the provided file.
This causes only the leaf cert to be presented. Make sure other certificates present in the pem file are also loaded.

updated tls_client_certificate.rs to generate and supply intermediate cert in it's test to confirm the delivery of the cert chain.

### DIFF
--- a/crates/integration-tests/src/test/tls_client_certificate.rs
+++ b/crates/integration-tests/src/test/tls_client_certificate.rs
@@ -1,7 +1,10 @@
 use crate::kumod::{DaemonWithMaildir, DeliverySummary, MailGenParams};
 use k9::assert_equal;
 use kumo_log_types::RecordType;
-use rustls_cert_gen::{Ca, CertificateBuilder, EndEntity};
+use rcgen::{
+    BasicConstraints, CertificateParams, CertifiedIssuer, DistinguishedName, DnType,
+    ExtendedKeyUsagePurpose, IsCa, KeyPair, KeyUsagePurpose,
+};
 use std::collections::BTreeMap;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


crates/lruttl/src/lib.rs
```diff
@@ -15,6 +13,7 @@ use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc, LazyLock, Weak};
 use tokio::sync::Semaphore;
 use tokio::time::{timeout_at, Duration, Instant};
+pub use {linkme, pastey as paste};
```
wez: Let's revert these rustfmt changes; they're not really part of this PR

crates/rfc5321/src/client.rs
```diff
@@ -21,8 +21,7 @@ use tokio_rustls::rustls::pki_types::ServerName;
 use tracing::Level;
 
 pub use kumo_tls_helper::TlsOptions;
-pub use openssl;
-pub use tokio_rustls;
+pub use {openssl, tokio_rustls};
```
wez: Let's revert these rustfmt changes; they're not really part of this PR

crates/kumo-tls-helper/src/lib.rs
```diff
@@ -261,8 +261,14 @@ impl TlsOptions {
         if let (Some(cert_data), Some(key_data)) =
             (&self.certificate_from_pem, &self.private_key_from_pem)
         {
-            let cert = X509::from_pem(cert_data)?;
-            builder.set_certificate(&cert)?;
+            let certs = X509::stack_from_pem(cert_data)?;
+            let leaf = certs[0].clone();
```
wez: This can potentially panic if `certs` is empty.  I don't know if that is possible in practice, but if someone does load a PEM that has zero certs I'd rather not panic.

Something like this should do the trick?

```suggestion
            let Some(leaf) = certs.get(0).map(Clone::clone) else {
                anyhow::bail!("certificate PEM data is empty");
            };
```

wez (commented):
Thanks!

crates/kumo-tls-helper/src/lib.rs
kayozaki: makes sense.

PR #497, opened 2026-03-25
Adding kumo.fs.stat to get file/dir meta data

kayozaki:
- Adding function to check the stat of a file or directory.

I personally wanted to get the last modified time of a file for the function i'm working on.
other attributes aren't necessarily required, but adding them as it may help others.

### DIFF
--- a/crates/mod-filesystem/src/lib.rs
+++ b/crates/mod-filesystem/src/lib.rs
@@ -24,6 +24,15 @@ macro_rules! add_metadata_field {
     };
 }
 
+#[cfg(unix)]
+macro_rules! system_time_to_lua_time {
+    ($epoch_secs:expr, $nsecs:expr) => {{
+        let dt = DateTime::<Utc>::from_timestamp($epoch_secs, $nsecs as u32)
+            .ok_or_else(|| mlua::Error::external("invalid timestamp"))?;
+        Ok(Time::from(dt))
+    }};
+}
+
 impl LuaUserData for MetadataWrapper {
     fn add_fields<F: UserDataFields<Self>>(fields: &mut F) {
         add_metadata_field!(fields, "is_file", is_file);
@@ -42,17 +51,20 @@ impl LuaUserData for MetadataWrapper {
             add_metadata_field!(fields, "gid", gid);
             add_metadata_field!(fields, "rdev", rdev);
             add_metadata_field!(fields, "size", size);
-            fields.add_field_method_get("atime", |lua, this| {
-                let atime_secs = this.0.atime();
-                system_time_to_lua_time(lua, Some(atime_secs))
+            fields.add_field_method_get("atime", |_lua, this| {
+                let secs = this.0.atime();
+                let nsec = this.0.atime_nsec();
+                system_time_to_lua_time!(secs, nsec)
             });
-            fields.add_field_method_get("mtime", |lua, this| {
-                let mtime_secs = this.0.mtime();
-                system_time_to_lua_time(lua, Some(mtime_secs))
+            fields.add_field_method_get("mtime", |_lua, this| {
+                let secs = this.0.mtime();
+                let nsec = this.0.mtime_nsec();
+                system_time_to_lua_time!(secs, nsec)
             });
-            fields.add_field_method_get("ctime", |lua, this| {
-                let ctime_secs = this.0.ctime();
-                system_time_to_lua_time(lua, Some(ctime_secs))
+            fields.add_field_method_get("ctime", |_lua, this| {
+                let secs = this.0.ctime();
+                let nsec = this.0.ctime_nsec();
+                system_time_to_lua_time!(secs, nsec)
             });
             add_metadata_field!(fields, "blksize", blksize);
             add_metadata_field!(fields, "blocks", blocks);
@@ -156,14 +168,6 @@ async fn symlink_metadata_for_path(_lua: Lua, path: String) -> mlua::Result<Meta
     Ok(MetadataWrapper(metadata))
 }
 
-#[cfg(unix)]
-fn system_time_to_lua_time(_lua: &Lua, epoch_secs: Option<i64>) -> mlua::Result<Option<Time>> {
-    Ok(epoch_secs.map(|secs| {
-        let dt = DateTime::<Utc>::from_timestamp(secs, 0).unwrap();
-        Time::from(dt)
-    }))
-}
-
 async fn glob(pattern: String, path: Option<String>) -> mlua::Result<Vec<String>> {
     let entries = tokio::task::spawn_blocking(move || {
         let mut entries = vec![];
@@ -209,6 +213,10 @@ assert(meta.is_file)
 assert(not meta.is_dir)
 assert(not meta.is_symlink)
 assert(meta.len == 11)
+
+if meta.ctime ~= nil then
+    assert(meta.ctime.unix_timestamp > 0)
+end
 
 if meta.size ~= nil then
     assert(meta.size == 11)
--- a/docs/reference/kumo.fs/metadata_for_path.md
+++ b/docs/reference/kumo.fs/metadata_for_path.md
@@ -7,18 +7,16 @@ tags:
 # kumo.fs.metadata_for_path
 
 ```lua
-kumo.fs.metadata_for_path(path)
-kumo.fs.symlink_metadata_for_path(path)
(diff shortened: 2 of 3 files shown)


crates/mod-filesystem/src/lib.rs
```diff
@@ -83,6 +97,53 @@ async fn uncached_glob(
     glob(pattern, path).await
 }
 
+async fn stat(lua: Lua, path: String) -> mlua::Result<mlua::Table> {
```
wez: This is a good idea, but we probably need to go a bit deeper for this, as it is a fairly general function with a lot of fields that folks might care about in different circumstances.

I think we should target https://doc.rust-lang.org/std/os/unix/fs/trait.MetadataExt.html as the source of the data, rather than the relatively bland portable set in the base Metadata struct.

That will give you accessors for essentially everything that you're familiar with from `struct stat`, and we can expose those through to lua.

The file_type concept is more convenient than testing against the modes, but rather than pre-calling the filetype methods, I think we should just expose those methods out to lua, so that they are evaluated lazily so that constructing and working with the result is cheaper.

I'd suggest looking at the `BinaryResult` wrapper in `mod-digest`; it uses `fields.add_field_method_get` to register field getter methods, so you can return the rust struct but expose its fields as though they were table fields (eg: `metdata.ino` rather than `metadata:info()`):

So we'd have something like:

```rust
use std::os::unix::fs::MetadataExt;

struct MetadataWrapper(Metadata);

impl LuaUserData for MetadataWrapper {
    fn add_fields<F: UserDataFields<Self>>(fields: &mut F) {
        fields.add_field_method_get("ino", |_, this| {
           Ok(this.0.ino())
        });

        fields.add_field_method_get("is_dir", |_, this| {
           Ok(this.0.is_dir())
         });
    }
}

async fn metadata_for_path(lua: Lua, path: String) -> mlua::Result<MetadataWrapper> {
    let metdata = tokio::fs::metadata(&path).await.map_err(mlua::Error::external)?;
    Ok(MetadataWrapper(metadata))
}
```

Bonus points if you use a macro to reduce the boilerplate, something like:

```rust
impl LuaUserData for MetadataWrapper {
    fn add_fields<F: UserDataFields<Self>>(fields: &mut F) {

macro_rules accessor {
    ($name:literal, $func:ident) => {
        fields.add_field_method_get($name, |_, this| { Ok(this.0.$func()) });
    }
)

        accessor!("ino", info);
        accessor!("dev", dev);
}
```

crates/mod-filesystem/src/lib.rs
```diff
@@ -30,7 +44,7 @@ pub fn register(lua: &Lua) -> anyhow::Result<()> {
     fs_mod.set("read_dir", lua.create_async_function(read_dir)?)?;
     fs_mod.set("glob", lua.create_async_function(cached_glob)?)?;
     fs_mod.set("uncached_glob", lua.create_async_function(uncached_glob)?)?;
-
+    fs_mod.set("stat", lua.create_async_function(stat)?)?;
```
wez: Since we're not strictly the same as `stat(2)`, I'd prefer to mirror the rust naming of `metadata` for this.  I'd lean a bit more towards `metadata_for_path` to be a bit more descriptive, but just `metadata` is OK if you think brevity is reasonable.

Because I spent a very long time working on [watchman](https://github.com/facebook/watchman), I've become very sensitive to the idea that you should always have a way to do the equivalent of an `lstat(2)`, so I'd love to also see a variant here that uses https://docs.rs/tokio/latest/tokio/fs/fn.symlink_metadata.html to get that information.

crates/mod-filesystem/src/lib.rs
```diff
@@ -104,3 +165,67 @@ async fn glob(pattern: String, path: Option<String>) -> mlua::Result<Vec<String>
...
+        let is_symlink: bool = table.get("is_symlink")?;
+        let len: u64 = table.get("len")?;
+        let ctime: mlua::Value = table.get("ctime")?;
+
+        assert_eq!(got_path, path);
+        assert!(is_file);
+        assert!(!is_dir);
+        assert!(!is_symlink);
+        assert_eq!(len, 11);
+        assert!(!matches!(ctime, mlua::Value::Nil), "ctime is missing");
```
wez: Testing the individual fields from rust feels a bit clunky.  I'd suggest bouncing into lua for this, something like:

```rust
    register(&lua)?;
    let chunk = lua.load(r#"
local meta = kumo.fs.metadata(path)
assert(meta.is_file)
assert(not meta.is_dir)
assert(not meta.is_symlink)
assert(meta.size == 11)
"#).unwrap();
    lua.globals().set("path", path);
    chunk.eval_async().await?;
```

and use this approach for the other cases below.

crates/mod-filesystem/src/lib.rs
```diff
@@ -83,6 +97,53 @@ async fn uncached_glob(
...
+
+fn system_time_to_lua_time(lua: &Lua, time: Option<SystemTime>) -> mlua::Result<Value> {
+    match time {
+        Some(t) => {
+            let dt: DateTime<Utc> = t.into();
+            Time::from(dt).into_lua(lua)
+        }
+        None => Ok(Value::Nil),
+    }
+}
```
wez: I think you can simplify this and have stronger typing:

```suggestion
fn system_time_to_lua_time(lua: &Lua, time: Option<SystemTime>) -> mlua::Result<Option<Time>> {
    time.map(|t| {
            let dt: DateTime<Utc> = t.into();
            Time::from(dt)  // could perhaps get away with dt.into() here
        })
}
```

crates/mod-filesystem/src/lib.rs
kayozaki: @wez 
Thank you for the review.
Just to confirm, you're fine using "std::os::unix::fs::MetadataExt" which appears to be limited to unix platform only ?

wez: Yeah, we have no plans to actively support non-unix platforms at this time. If that changes in the future, we can address this sort of thing then.

wez: (I notice that I've at least twice typo'd `ino` -> `info` in my earlier review.  I meant `ino` in both cases!)

crates/mod-filesystem/src/lib.rs
```diff
@@ -1,13 +1,65 @@
...
+            add_metadata_field!(fields, "ino", ino);
+            add_metadata_field!(fields, "mode", mode);
+            add_metadata_field!(fields, "nlink", nlink);
+            add_metadata_field!(fields, "uid", uid);
+            add_metadata_field!(fields, "gid", gid);
+            add_metadata_field!(fields, "rdev", rdev);
+            add_metadata_field!(fields, "size", size);
+            fields.add_field_method_get("atime", |lua, this| {
+                let atime_secs = this.0.atime();
+                system_time_to_lua_time(lua, Some(atime_secs))
```
wez: A couple of things for these time accessors:

1. We have the nanosecond component available and we can feed that into chrono to get sub-second precision
2. We don't need an Option around either seconds or nanoseconds
3. There's an unwrap that we should avoid in case someone manages to put complete nonsense into time fields (eg: might happen if the fs is corrupt, but might also be possible if someone gets creative)

Something like this:

```suggestion
                let secs = this.0.atime();
                let nsec = this.0.atime_nsec();
                let dt = DateTime::<Utc>::from_timestamp(secs, nsec)
                      .map_err(mlua::Error::external)?;
                Ok(Time::from(dt))
```

which you can also turn into a little macro for the 3 different timestamps or amend the `system_time_to_lua_time` helper function accordingly.

docs/reference/kumo.fs/metadata_for_path.md
```diff
@@ -0,0 +1,56 @@
...
+tags:
+ - utility
+ - filesystem
+---
+
+# kumo.fs.metadata_for_path
+
+```lua
+kumo.fs.metadata_for_path(path)
+kumo.fs.symlink_metadata_for_path(path)
```
wez: I'd appreciate having a separate doc page for `symlink_metadata_for_path` to make the search experience (potentially!) better, but it needn't copy the full text from here; it can have a stub paragraph below its synopsis and other metadata that says something like:

```markdown
This function behaves exactly like [metadata_for_path](metadata_for_path.md),
except that it does not follow symbolic links and instead returns information
about the symbolic link itself.
```

wez:
Thank you!

PR #498, opened 2026-03-25
add convenient function for ends_with, starts_with

kayozaki:
There are couple small techniques to achieve ends_with, starts_with in lua, but seems a lot simpler if we have those readily available to use in kumo.string module.

### DIFF
--- a/crates/mod-string/src/lib.rs
+++ b/crates/mod-string/src/lib.rs
@@ -1,6 +1,5 @@
 use bstr::ByteSlice;
 use config::{from_lua_value, get_or_create_sub_module};
-use bstr::ByteSlice;
 use kumo_template::TemplateDialect;
 use mlua::Lua;
 


crates/mod-string/src/lib.rs
```diff
@@ -39,6 +39,16 @@ pub fn register(lua: &Lua) -> anyhow::Result<()> {
         })?,
     )?;
 
+    string_mod.set(
+        "starts_with",
+        lua.create_function(move |_, (s, pattern): (String, String)| Ok(s.starts_with(&pattern)))?,
```
wez: Let's make these capable of operating on the byte form of strings; add `bstr.workspace = true` to the deps in this crate and `use bstr::ByteSlice;`, and you should be able to use https://docs.rs/bstr/latest/bstr/trait.ByteSlice.html#method.starts_with_str for this.

Something like:

```suggestion
        lua.create_function(move |_, (s, pattern): (mlua::String, mlua::String)| Ok(s.as_bytes().starts_with_str(&pattern.as_bytes())))?,
```

docs/reference/string/ends_with.md
```diff
@@ -0,0 +1,13 @@
+# kumo.string.ends_with
+
+```lua
+kumo.string.ends_with(STRING, PATTERN)
+```
+
+{{since('dev')}}
+
+Returns true if STRING ends with PATTERN, otherwise false.
```
wez: Probably worth clarifying that PATTERN is a literal string.  Perhaps use `SUFFIX` instead of pattern throughout this doc to avoid misconception from folks skimming these docs?

docs/reference/string/starts_with.md
```diff
@@ -0,0 +1,13 @@
+# kumo.string.starts_with
+
+```lua
+kumo.string.starts_with(STRING, PATTERN)
+```
+
+{{since('dev')}}
+
+Returns true if STRING starts with PATTERN, otherwise false.
```
wez: Similar to above, switch `PATTERN` to `PREFIX` throughout, and clarify that it is a literal comparison

PR #504, opened 2026-03-30
mod-http: allow disabling certificate verification on the client

Fallmay:
Add `accept_invalid_certs` option to `kumo.http.build_client`.

Allows connecting to servers with self-signed or otherwise invalid certificates, which is useful in development and testing environments where a proper certificate is not available.


Fallmay:
@wez mind taking a look at this PR? Really missing this in the test env.

wez:
Thank you!

PR #505, opened 2026-04-08
Fix multiple issue relating to DMARC evaluation

kayozaki:
- Handle parsing failure when DMARC record ends with semicolon
- Adjust kumo.dmarc.check_msg so it considers SPF/DKIM verdict and not just alignment
- Made kumo.dmarc.check_msg to pass, if either SPF or DKIM succeeds alignment check
- Fix relax mode SPF/DKIM so the comparing hostnames considers top-level domain alignment
- Made alignment comparison case insensitive
- Added support for SPF to fallback to ehlo domain, when mail from domain is blank
- Enrich props against dkim auth_resutls for better troubleshooting capabilities
- Enrich props against dmarc auth_resutls for better troubleshooting capabilities
- Fixed a bug in mail_auth.lua, where you can't disable specific module. whether you set it to false or nil in config, it always defaults to being enabled.


crates/kumo-dmarc/src/tests.rs
```diff
@@ -327,11 +366,107 @@ async fn dmarc_pct_rate() {
...
+        from_domain: "example.com",
+        mail_from_domain: "example.net",
+        dkim_domains: &[],
+        spf_result: Some("fail"),
+        resolver: &resolver,
+    })
+    .await;
+
+    k9::assert_equal!(result.result, Disposition::Reject);
+    k9::assert_equal!(
```
wez, line 390: There's a lot of boilerplate in this section.  Consider using `k9::snapshot!(result.props);` here to have k9 fill out how the props render in Debug print?

Alternatively, make a little inline helper function to get a prop out by name as a string.

wez (commented):
We landed some changes for dmarc reporting and there's a conflict here now; please rebase!

Note: I cherry picked out the typing.lua fix and pushed that to main already.

Thank you!

PR #507, opened 2026-04-22
audit series using kumo-counter-series

kayozaki:
Continuation of https://github.com/KumoCorp/kumomta/pull/470

kumo.audit_series uses kumo-counter-series internally to maintain an in memory sliding window counters.
The main objective of this module is to be able to track counts of specific events.

For example count the number of failures against redis. Repeated failure can negatively impact the performance of the MTA. In such case, I would like the code to bypass making calls to redis.

The counters are tracked in memory thus restarting the service would reset the values.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3914,6 +3914,7 @@ dependencies = [
  "metrics-tracing-context",
  "mlua",
  "mod-amqp",
+ "mod-audit-series",
  "mod-aws-sigv4",
  "mod-crypto",
  "mod-digest",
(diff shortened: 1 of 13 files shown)


wez:
Thanks for this; I made a pass over this to make the define function more deterministic if you eg: reload the config with adjusted parameters, changed that function to accept a struct so that it has "named parameters", added missing docs and renamed it to match our counter series naming rather than the audit series naming which comes from a focused use case in an alternative product.

Issue #508, opened 2026-04-22
proxy-server passthru phase can keep stalled sessions open indefinitely without an idle timeout


jack-atlas:
We are using `proxy-server` as a SOCKS5 proxy for KumoMTA SMTP delivery.

  From reading the code, the SOCKS5 handshake/auth/request path is covered by timeouts, but after `CONNECT` succeeds the session enters
  passthru mode and the proxy just forwards raw bytes between client and remote peer.

  In that passthru phase, if the remote side silently stalls the TCP session without sending more data and without closing the socket
  (`FIN`/`RST`), the session can remain open indefinitely.

  In our environment this shows up as:

  - SMTP delivery stops making progress
  - proxy-side file descriptors keep increasing over time
  - sessions do not get released promptly
  - restarting the proxy clears the accumulated stuck connections

  This looks like a resource-retention issue in passthru mode rather than a classic memory leak.

  Expected behavior:

  - `proxy-server` should support a configurable passthru idle timeout
  - the timeout should be based on lack of byte progress in either direction
  - when the passthru session is idle for too long, both sides should be closed

  It would also help if the documentation clarified that the current `timeout` setting applies to the SOCKS5 negotiation/request phase, but not necessarily to the long-lived passthru phase.

jack-atlas:
We collected evidence from a long-running HA `proxy-server` node that has not been restarted.

  Process stats:
  - total fds: 92360
  - TCP fds: 24325

  TCP state distribution from `lsof -nP -a -p <pid> -iTCP`:
  - FIN_WAIT2: 21370
  - ESTABLISHED: 1608
  - CLOSE_WAIT: 583
  - FIN_WAIT1: 503
  - SYN_SENT: 223
  - LISTEN: 1

  So only ~6.6% of TCP sockets are in ESTABLISHED state, while ~87.8% are stuck in FIN_WAIT2.

  More importantly, we also see many inbound proxy sockets on `:6000` in CLOSE_WAIT at the same time that outbound sockets to remote MX
  hosts on `:25` are in FIN_WAIT2.

  This strongly suggests that passthru sessions are not being fully torn down in half-close / stalled-close scenarios:
  - the client side has already closed
  - the proxy still keeps the inbound fd open (CLOSE_WAIT)
  - the outbound side remains stuck in FIN_WAIT2
  - these socket pairs accumulate over time and eventually exhaust file descriptors

[...]

PR #509, opened 2026-04-23
proxy-server: enable TCP keepalive for black-holed peers

jack-atlas:
## Problem

Some upstream peers (e.g. QQ Mail's rate-limiter) respond to connection-rate pressure by silently holding a proxied TCP connection open — sending neither data, FIN, nor RST — rather than cleanly refusing with a 4xx/5xx response. Without an application-level watchdog, the two file descriptors for such a session remain open for the lifetime of the process, slowly exhausting the kernel's file-descriptor table and occupying proxy-server worker slots.

## Solution

A new `passthru_idle_timeout` field is added to `ProxyListenerParams`. When set, the passthru phase replaces `copy_bidirectional` / `splice` with a userspace copy loop (`copy_with_idle_timeout`) that wraps every individual `read()` call in a `tokio::time::timeout`. If either the client→remote or remote→client direction goes the full `passthru_idle_timeout` without reading a single byte, the session is torn down and both sockets are closed.

### Default

**5 minutes** (300 s). This is well clear of real slow-start windows while tight enough to prevent stalled sessions from accumulating. Operators can tune it or disable it entirely:

```lua
-- shorten to 2 minutes
proxy.start_proxy_listener {
  listen = '0.0.0.0:1080',
  passthru_idle_timeout = '2 minutes',
}

-- disable entirely (restores previous copy_bidirectional / splice path)
proxy.start_proxy_listener {
  listen = '0.0.0.0:1080',
  passthru_idle_timeout = 'none',
}
```

### Interaction with `use_splice`

`splice(2)` does not expose per-direction byte progress to userspace, so **setting `passthru_idle_timeout` forces the userspace copy path** regardless of `use_splice`. Splice is still used when the idle timeout is disabled (`none`).

## Changes

| File | Description |
|------|-------------|
| `crates/proxy-server/src/mod_proxy.rs` | New `passthru_idle_timeout` field with 300 s default; wired through `handle_client` |
| `crates/proxy-server/src/proxy_handler.rs` | `passthru_with_idle_timeout` + `copy_with_idle_timeout` helpers; 2 unit tests |
| `docs/reference/proxy/start_proxy_listener/passthru_idle_timeout.md` | New reference page |

## Tests

Two `#[tokio::test]` unit tests cover the new helper:

- **`copy_eof_returns_byte_count`** — verifies correct byte accounting on clean EOF.
- **`copy_idle_timeout_fires`** — uses `tokio::time::pause()` + `advance()` to confirm a stalled read returns `io::ErrorKind::TimedOut` without real wall-clock delay.

```
test proxy_handler::tests::copy_eof_returns_byte_count ... ok
test proxy_handler::tests::copy_idle_timeout_fires ... ok
```


jack-atlas:
For fix this issue  https://github.com/KumoCorp/kumomta/issues/508

crates/proxy-server/src/mod_proxy.rs
```diff
@@ -34,7 +34,26 @@ pub struct ProxyListenerParams {
...
     pub timeout: Duration,
 
-    /// Whether to use splice(2) on Linux for proxied connections
+    /// If set, tear the passthru session down when no data has flowed in
+    /// either direction for this long. Defaults to 5 minutes so that a
+    /// silently stalled upstream peer (for example, an ISP black-holing
+    /// a connection without sending FIN or RST) cannot pin the two file
+    /// descriptors for the lifetime of the process.
+    ///
+    /// Setting this to `none` (or omitting it in a build that explicitly
```
wez: I think this is effectively impossible to achieve in lua, because a table field with a nil value disappears and is indistinguishable from not being set, and the logic in here will treat that the same as setting it to 5 minutes

wez (commented):
Thanks for digging into this!
It feels complex because we've now got three different ways to proxy data through.
What do you think about the suggestion made over here:
https://github.com/tokio-rs/tokio/issues/4556#issuecomment-2440021582

Can we just tell the kernel to handle this for us, and then it should work with splice?
Could you try this out?

jack-atlas (commented):
I reworked this to use the TCP keepalive approach suggested in tokio-rs/tokio#4556.

The branch now enables TCP keepalive on both the accepted client socket and the outbound proxy TCP socket. On Linux it configures a 300s keepalive time, 30s probeinterval, and 3 retries.

This keeps the existing splice(2) fast path intact, avoids adding a third passthrough copy path, and removes the new `passthru_idle_timeout` Lua parameter, which also
  avoids the nil-vs-unset ambiguity you pointed out.

One caveat: this is not a fully equivalent application-level idle timeout. If a peer still ACKs TCP keepalive probes but simply stops sending application data, TCP
  keepalive will not terminate the connection. For this PR, though, I think the kernel-level keepalive approach is the cleaner tradeoff.

  Verified with:
  - cargo test -p proxy-server
  - cargo fmt --check
  - cargo clippy -p proxy-server --no-deps -- -D warnings

wez:
Thanks; I made another pass over it to expose keepalive options to lua and pushed that version

PR #512, opened 2026-05-01
make_message fails for from address which starts with postmaster

kayozaki:
When running make_message with from address which is prefixed with postmaster, the EnvelopeAddress::parse function is failing.

```
Error: Shutdown due to error: Error at line 1, in Eof:
postmaster@domain.com
          ^____________________


stack traceback:
        [C]: in function 'kumo.make_message'
        [string "/tmp/test.lua"]:154: in upvalue 'scan'
        [string "/tmp/test.lua"]:201: in function <[string "/tmp/test.lua"]:188>
```

Not directly related to the fix, but crates/rfc5321/src/client.rs is failing to run the test with error below.
This is also addressed

```
running 1 test
error[E0308]: mismatched types
 --> crates/rfc5321/src/client.rs:1189:1
  |
2 |   fn main() { #[allow(non_snake_case)] fn _doctest_main_crates_rfc5321_src_client_rs_1188_0() {
  |                                                                                              - help: try adding a return type: `-> [&str; 8]`
3 | / ["C=US", "ST=CA", "L=SanFrancisco", "O=Fort-Funston", "OU=MyOrganizationalUnit",
4 | | "CN=do.havedane.net", "name=EasyRSA", "emailAddress=me@myhost.mydomain"]
  | |________________________________________________________________________^ expected `()`, found `[&str; 8]`

error: aborting due to 1 previous error

For more information about this error, try `rustc --explain E0308`.
Couldn't compile the test.test crates/rfc5321/src/client.rs - client::subject_name (line 1188) - compile ... FAILED
```

### DIFF
--- a/crates/rfc5321/src/client.rs
+++ b/crates/rfc5321/src/client.rs
@@ -1187,7 +1187,7 @@ fn apply_dot_stuffing(data: &[u8]) -> Option<Vec<u8>> {
 /// eg:
 /// ```no_run
 /// ["C=US", "ST=CA", "L=SanFrancisco", "O=Fort-Funston", "OU=MyOrganizationalUnit",
-/// "CN=do.havedane.net", "name=EasyRSA", "emailAddress=me@myhost.mydomain"]
+/// "CN=do.havedane.net", "name=EasyRSA", "emailAddress=me@myhost.mydomain"];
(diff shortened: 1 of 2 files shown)


wez:
Thank you!

PR #513, opened 2026-05-01
Fix multiple issue relating to DMARC evaluation

kayozaki:
This is a redo of https://github.com/KumoCorp/kumomta/pull/505
The amount of conflict looked safer to start the PR over.

- Handle parsing failure when DMARC record ends with semicolon
- Adjust kumo.dmarc.check_msg so it considers SPF/DKIM verdict and not just alignment
- Made kumo.dmarc.check_msg to pass, if either SPF or DKIM succeeds alignment check
- Fix relax mode SPF/DKIM so the comparing hostnames considers top-level domain alignment
- Made alignment comparison case insensitive
- Added support for SPF to fallback to ehlo domain, when mail from domain is blank
- Enrich props against dkim auth_resutls for better troubleshooting capabilities
- Enrich props against dmarc auth_resutls for better troubleshooting capabilities

### DIFF
--- a/crates/dkim/src/lib.rs
+++ b/crates/dkim/src/lib.rs
@@ -258,6 +258,33 @@ pub async fn verify_email_with_resolver<'a>(
     email: &'a ParsedEmail<'a>,
     resolver: &dyn Resolver,
 ) -> Result<Vec<AuthenticationResult>, DKIMError> {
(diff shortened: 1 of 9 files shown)


wez:
Thanks; I re-split this into more focused commits, and expanded the psl::domain_str commit to apply more generally because we had 3 different crates that called into it.

PR #515, opened 2026-05-04
option to choose the first named header when calling import_x_headers

kayozaki:
When there are multiple of the same header on an message, import_x_headers would prefer the bottom most value. This behavior tends to not match what mail agent would render, as mail agent would typically display the top most header. 

Add an 2nd argument when calling import_x_headers, instructing import_x_headers to prefer the top value instead of bottom.

### DIFF
--- a/crates/message/src/message.rs
+++ b/crates/message/src/message.rs
@@ -1167,21 +1167,42 @@ impl Message {
         .await
     }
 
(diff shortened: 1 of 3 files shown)


wez:
Thanks for this! I actually went a different route and added a more powerful alternative interface as `msg:import_headers`

PR #516, opened 2026-05-05
Added support of providing static per-recipient metadata in HTTP Injection API

Harshjha3006:
Adds a metadata field to the Recipient object in the HTTP injection API. Key-value pairs supplied in this field are stored in the message's rcpt_meta metadata entry, making them accessible throughout the message lifecycle via `msg:get_meta('rcpt_meta')` in Lua policy hooks.    
This enables us to include this metadata in ways like 
- Supplemental Trace Headers 
   ```
   kumo.apply_supplemental_trace_header(msg, {
     supplemental_header = true,
     header_name = 'X-KumoRef',
     include_meta_names = {'rcpt_meta'},
   })
   ```
   Or directly in the request payload using its `trace_headers` field. 
   ```
   {
   "trace_headers": {
     "received_header": false,
     "supplemental_header": true,
     "header_name": "X-KumoRef",
     "include_meta_names": {'rcpt_meta'},
    },
  }
   ```
   
- Webhook Events and Logs
  ```
  kumo.configure_local_logs {
     log_dir = '/var/log/kumomta',
     meta = { 'rcpt_meta' },
  }
  ```

and in various other forms of custom logic without using substitutions and thus improving performance for the case where there are a large number of metadata variables per recipient. 

I have added an integration test for one of the possible use-cases.

### DIFF
--- a/crates/integration-tests/source-rcpt-meta.lua
+++ b/crates/integration-tests/source-rcpt-meta.lua
@@ -26,15 +26,18 @@ kumo.on('init', function()
   }
 end)
 
-kumo.on('get_queue_config', function(domain, _tenant, _campaign, _routing_domain)
-  return kumo.make_queue_config {
-    protocol = {
-      smtp = {
-        mx_list = { 'localhost:' .. SINK_PORT },
+kumo.on(
+  'get_queue_config',
+  function(domain, _tenant, _campaign, _routing_domain)
+    return kumo.make_queue_config {
+      protocol = {
+        smtp = {
+          mx_list = { 'localhost:' .. SINK_PORT },
+        },
       },
-    },
-  }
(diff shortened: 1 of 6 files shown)


wez:
Thank you!

sinfloodmusic:
Thank you so much!  I struggled for a bit on how to get the data coming though to the webhook but got it working  with this config using the 'dev' (most recent) docker image.  Posting this in case this helps someone:
```
--  Send webhooks for delivery receipt
log_hooks:new_json {
  name = 'webhook',
  url = 'http://myapi/kumo_callback',
  log_parameters = {
    headers = { 'Subject', 'X-Customer-ID' },
    meta = { 'extra' }
  },
}
```

PR #517, opened 2026-05-06
query DANE against target MX host instead of routing domain

dirkcjelli:
DNS resolution for DANE is incorrectly using the routing domain (`mx.domain_name`) instead of the specific MX host (`address.name`). 
This causes validation failures when the MX record differs from the routing domain.

As per [RFC7672 (3.2.2)](https://www.rfc-editor.org/rfc/rfc7672#section-3.2.2) "the MX hostname is the only reference identifier."

To be honest, I'm not sure if tests need to be adjusted


wez:
Thanks for raising this; I folded the equivalent change into the stack of DANE related changes that I pushed yesterday

PR #523, opened 2026-05-17
bugfix: invalid control characters in Authentication-Results

raphting:
Microsoft 365 / Outlook tends to send malformed ARC headers. They lead to an error report from `mail_auth.check`. The function writes this report as a mail header:

```
...
arc=fail reason="An ARC-Authentication-Results header could not be parsed: invalid header: Error at line 2, in Eof: smtp.mailfrom=example.com; dmarc=pass action=none header.from=example.com; dkim=pass ^______________________________________________________________________"
```

Before this bugfix, the `reason` string contained control characters including `\n` from the error formatting. This led to a malformed header. Software like Stalwart bounces the mail.

The already existing function `is_ctl` is RFC conform and filters invalid characters before adding them to the header.

### DIFF
--- a/crates/dkim/src/arc.rs
+++ b/crates/dkim/src/arc.rs
@@ -583,4 +583,63 @@ mod test {
             email_content = sealed;
         }
     }
+
+    /// Microsoft 365 / Outlook sends ARC-Authentication-Results headers with
+    /// non-standard syntax (e.g. bare `action=none` without a `ptype.`
+    /// prefix).  The nom parser consumes what it understands and then
+    /// `all_consuming` produces an `Eof` error for the leftover input.
+    /// `explain_nom` formats that error as a multi-line diagnostic string
+    /// containing `\n`.  That string ends up in `ARCIssue::reason` and, from
+    /// there, in the `reason=` field of the emitted `Authentication-Results`
+    /// header — producing an invalid folded header value.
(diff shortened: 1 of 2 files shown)


wez:
Sorry, I meant to comment on this earlier.  I looked into this and, while the bug is valid, fixing it here this way is invalid because this particular function is used in a number of different contexts and this change would break some of them.

In the long run I think that it may be best to split this function up so that each location is using the appropriate encoding rules.

In the short term it's probably simplest to ensure that correct input at the appropriate location, but it's not super clear to me exactly where that is because this PR is very light on those details and there is no test coverage.

What I'd love to see here is a unit test that runs an example of the offending input through whatever code path you're seeing things trigger in.  We can then backtrack from there to find the appropriate place to resolve this, without impacting other code paths.

raphting:
I reverted the initial fix and added a test case instead.

It reveals the issue in the call chain:

`explain_nom (nom-utils) -> MailParsingError::HeaderParse(String) -> ARC::verify (dkim/src/arc.rs) -> AuthenticationResult::reason -> emit_value_token (mailparsing)`

I believe `ARC::verify` should sanitize the `reason` before writing it to `ARCIssue`. Curious if you consider this the appropriate place or if you would fix the code somewhere else?

PR #524, opened 2026-05-17
add custom resolver to lookup_mx

kayozaki:
lookup_mx does not accept custom resolver like other dns functions. 
This change adds capability for lookup_mx call to honor custom resolver as an secondary argument.
Mainly used while creating unit tests, where lookup_mx can be made less dependent on the environment which it's runnig in.


wez:
I folded the equivalent functionality into the stack over in https://github.com/KumoCorp/kumomta/pull/557 (not yet committed).  Closing this in favor of that!

PR #531, opened 2026-05-29
Updates to the community shaping config

Solmea:
It was time for an update to the community rules and moved more configurations over to provider level automations.
Short list oof changes:
- Added automations for provider.apple with a remark about icloud.com as that is the primary used domain
- Added automations for yahoo.co.jp as it uses its own mail servers
- Moved web.de domain automations to gmx_net_web_de provider automations
- Added provider definitions for: barracudanetworks.com, netvigator.com, kpnmail.nl
- Added  one automation for netvigator.com, mimecast provider
- Moved qq.com and 163.com automations into provider automations

I also tried to remove matching rules with the default.automations as those will apply instead of later configued automations. As for example:  
421 4.7.0 [TS01] Messages from x.x.x.x temporarily deferred due to user complaints
will be caught in the regex match of the "default".automation:
'''Messages from \d+\.\d+\.\d+\.\d+ temporarily deferred'''

### DIFF
--- a/assets/community/shaping.toml
+++ b/assets/community/shaping.toml
@@ -28,7 +28,8 @@
 
 #### BEGIN COMMUNITY CONTRIBUTED SHAPING RULES
 
-#===== GMAIL
+#===== GMAIL 
+# default provider configs in policy-extra/shaping.toml
 [[provider."gmail".automation]]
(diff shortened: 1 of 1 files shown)


MHillyer:
Thanks!

PR #534, opened 2026-05-31
add initial support for dmarc reporting email

sophiajt:
This starts laying out the groundwork for emailing out the generated DMARC report.

A few steps remain:
- Test coverage
- Documenting the new functions
- How do we want to handle if we can't generate the DMARC report?
- How do we want to handle if the DMARC email send fails? (as this is handled on the lua side, that may be up to the user?)
- There's one spot where I tokio spawn because of the limitations of async fn's. There may be a cleaner way to do this.


crates/kumo-dmarc/src/lib.rs
```diff
@@ -31,6 +35,219 @@ mod tests;
 
 const DMARC_REPORT_LOG_FILEPATH: &'static str = "/var/log/kumomta/dmarc.log";
 
+static ERROR_STREAM: LazyLock<Mutex<Option<mpsc::Sender<ErrorRecord>>>> =
+    LazyLock::new(|| Mutex::new(None));
+
+pub(crate) async fn send_error_message(record: ErrorRecord) -> Result<(), SendError<ErrorRecord>> {
```
wez: Naming: I'd lean towards something like `capture_error_record` for this because "send" in the context of kumomta is more closely associated with transmitting out of process

crates/kumo-dmarc/src/lib.rs
```diff
@@ -31,6 +35,219 @@ mod tests;
 
 const DMARC_REPORT_LOG_FILEPATH: &'static str = "/var/log/kumomta/dmarc.log";
 
+static ERROR_STREAM: LazyLock<Mutex<Option<mpsc::Sender<ErrorRecord>>>> =
```
wez: Naming: I'd lean towards something like `RECORD_STREAM` for this, because "error" has more of a runtime error connotation than a dmarc report record

crates/kumo-dmarc/src/lib.rs
```diff
@@ -31,6 +35,219 @@ mod tests;
...
+    LazyLock::new(|| Mutex::new(None));
+
+pub(crate) async fn send_error_message(record: ErrorRecord) -> Result<(), SendError<ErrorRecord>> {
+    // Take a copy of the sending stream for ourselves so we can safely send on it
+    let stream = ERROR_STREAM.lock().clone();
```
wez: I'm always a bit leery of `.lock().clone()`, followed by blocking actions on the returned value. I've been bitten several times by the lock guard lifetime living longer than it appears, leading to a deadlock.  As a defensive measure I typically will make a helper function to guarantee that the guard is released:

```rust
fn get_record_stream() -> Option<Sender<ErrorRecord>> {
   RECORD_STREAM.lock().clone()
}
```

and then here:

```rust
if let Some(stream) = get_record_stream() {
     stream.send(record).await
}
```

crates/kumo-dmarc/src/lib.rs
```diff
@@ -31,6 +35,219 @@ mod tests;
...
+    if let Some(stream) = stream {
+        stream.send(record).await
+    } else {
+        Ok(())
+    }
+}
+
+pub fn startup_error_reporter(reporting_callback: fn(String, String)) {
```
wez: I think I'd structure this a bit differently.  In broad strokes, I think we might want some configuration for the reporter:

```rust
const DEFAULT_REPORT_WINDOW_SIZE: Duration = Duration::new(60*60);
static REPORTER_CONFIG: LazyLock<Mutex<ReporterConfig>> = LazyLock::new(||Default::default());

#[derive(Default)]
struct ReporterConfig {
   /// How long between sending reports.  You can pick a better name!
   window_size: Duration, 
   /// Where to put the records on disk.  Default `/var/log/kumomta/dmarc-reports`
   log_dir: PathBuf,
}
```

In `start_dmarc_reporter`, we'd have it accept an optional ReporterConfig as its parameter and then it would do:

1. `*REPORTER_CONFIG.lock() = config.unwrap_or_default();`
2. Use a once-init to spawn the error_reporter_loop

That config struct replaces the parameter passing through start_error_reporter and error_reporter_loop.

With this structure, when the config is reloaded, the changes are picked up through the globals and we aren't required to directly manipulate the task.

crates/kumo-dmarc/src/lib.rs
```diff
@@ -31,6 +35,219 @@ mod tests;
...
+    tokio::spawn(async move {
+        error_reporter_loop(receiver, reporting_callback).await;
+    });
+
+    *ERROR_STREAM.lock() = Some(sender);
+}
+
+async fn error_reporter_loop<'a>(
+    mut receiver: Receiver<ErrorRecord>,
+    reporting_callback: fn(String, String),
```
wez: per the above comment, I'd remove the callback parameter in favor of using global configuration

```suggestion
  
```

In the body of this function I think we have a couple of goals:

1. Figure out the current wall time and compare with the reporting window parameters.  We need to dispatch processing for all the report data that occurred within complete segments.  It is possible to use a simple modulo division on the unix timestamp to "round off" to the nearest segment.  eg: if you want hourly-on-the-hour, you can `ts % 3600` to get the starting timestamp for that hour segment, then look for that `+ 3600 -1` to find the ending records.   I think that we may want to consider spawning that record processing into a separate task so that we can then continue with...
2. the main waiting loop, which is basically what you've sketched out below, but modified so that the timer ticks align on (or rather, just slightly after) the window size.  When the tick wakes us up, I'd suggest then spawning a task to process the records in the prior window.  (we need to ensure that we only spawn one task for a given window.  the task could be spawned with multiple windows, but only one should own a given window).
3. Accumulate records send via the channel.  I'd suggest that we record them into files that are named with the starting window timestamp embedded, under the configured log directory.  eg: `/var/log/kumomta/dmarc-reports/2026-06-09T01:00:00Z` (rounded back to an hour per the above).  Having the logs segment per-file makes the catchup logic simpler and easier to isolate
4. I think we could then reasonably keep the current segment file open here in a local to avoid re-opening it; when a record comes in we'd sanity check the timestamp and close/flush if the current open one is no longer within the window, then open a new file for the current window.  When the timer ticks we'd also flush and close the current open file, if any, before we spawn the dispatcher task.

crates/kumo-dmarc/src/lib.rs
```diff
@@ -31,6 +35,219 @@ mod tests;
...
+                let email_reports = aggregate_errors().await;
+
+                if let Ok(email_reports) = email_reports {
+                    for (email, report) in email_reports {
+                        if let Ok(result) = instant_xml::to_string(&report) {
+                            reporting_callback(result, email);
+                        }
+                    }
+                } else {
+                    // todo: we need a way to warn the user the report could not be processed
```
wez: I think this bit could be lifted into its own async function and the structure adjusted so that we can do something like:

1. On startup, discover any segments that we haven't dispatched yet.  Logically this means reading the log directory from the config and batch the filename(s) into their appropriate windows, then pass the list of filenames to `aggregate_errors` for each batch/window. 
2. in "normal operation", when we recognize that we've closed a window, we'd then process just the file name(s) for that window.

(it could potentially be multiple windows if the user reloads the config with a changed window size!)

crates/kumo-dmarc/src/lib.rs
```diff
@@ -31,6 +35,219 @@ mod tests;
...
+                    }
+                }
+            }
+            _ = timer.tick() => {
+                let email_reports = aggregate_errors().await;
+
+                if let Ok(email_reports) = email_reports {
+                    for (email, report) in email_reports {
+                        if let Ok(result) = instant_xml::to_string(&report) {
+                            reporting_callback(result, email);
```
wez: I was going to suggest that we could simply and directly declare_event! here in this crate and it would eliminate wrangling the passing of this callback.  I could go either way on that TBH.  I don't mind if kumo-dmarc depends on the `config` crate to facilitate that.  If you'd prefer not to add that dep, then I suppose that passing the callback via that config struct is *a* way to go, but that has it's own rough edges.

crates/kumod/src/dmarc.rs
```diff
@@ -15,9 +18,34 @@ struct CheckHostOutput {
     result: AuthenticationResult,
 }
 
+declare_event! {
+    static DMARC_REPORT_GENERATED: Multiple("dmarc_report_generated",
```
wez: Multiple allows multiple instances of the report hook to be registered, and without a good cooperative protocol for coordinating between those instances of hooks, it seems that someone could easily end up sending duplicates if they eg: have copy-pasta and end up with two instances of the same one.

Let's make this a Single registration for the time being and see if anyone shows up with a good use case for Multiple.

```suggestion
    static DMARC_REPORT_GENERATED: Single("dmarc_report_generated",
```

crates/kumo-dmarc/src/lib.rs
sophiajt: Can you say a bit more about this? Specifically, I'm a bit confused around:

`*REPORTER_CONFIG.lock() = config.unwrap_or_default();`

If we're passing in the report config, do we expect this is changing regularly? Or would it change on restarts of the server?

I'm a little nervous about this needing to lock every time a message comes in to read the config, as that lock may negatively impact performance under high message loads.

crates/kumo-dmarc/src/lib.rs
sophiajt: I think this might be one of those "how did you come to Rust" things. I came via functional programming, so this callback is the cleaner way rather than tighter dependency coupling. That said, happy to go either way.

crates/integration-tests/src/test/queue_dmarc.rs
```diff
@@ -0,0 +1,61 @@
...
+        .start()
+        .await
+        .context("DaemonWithMaildir::start")?;
+
+    eprintln!("sending message");
+    let mut client = daemon.smtp_client().await.context("make smtp_client")?;
+
+    let response = MailGenParams {
+        recip: Some("permfail@sub.example.com"),
+        body: Some("woot"),
```
wez: nit: unless you need this specific body, you could just allow the default for this

crates/integration-tests/src/test/queue_dmarc.rs
```diff
@@ -0,0 +1,61 @@
...
+        .to_header_string()
+        .contains("for <dmarc-feedback@example.com>"));
+    assert!(parsed.headers().get_first("X-KumoRef").is_some());
+
+    assert_equal!(parsed.headers().subject().unwrap().unwrap(), "DMARC Report");
+
+    let body = parsed.body().unwrap();
+    let body = body.to_string_lossy();
+
+    assert!(body.contains("<email>dmarc-feedback@example.com</email>"));
```
wez: let's have a comment here that says something along the lines of:

```suggestion
    // Spot check that the content looks sufficiently like a dmarc report.
    // We don't do a full validation here because testing the report shape
    // is covered by tests in kumo-dmarc itself
    assert!(body.contains("<email>dmarc-feedback@example.com</email>"));
```

crates/integration-tests/dmarc.lua
```diff
@@ -0,0 +1,152 @@
...
+]],
+  }
+
+  kumo.dmarc.set_report_window_in_seconds(4)
+  kumo.dmarc.start_dmarc_reporter()
+
+  kumo.configure_local_logs {
+    log_dir = TEST_DIR .. '/logs',
+    max_segment_duration = '1s',
+    headers = { 'X-*', 'Y-*', 'Subject' },
```
wez: let's take out the header logging from here: we're planning on marking this function as deprecated and you don't need it for this test

```suggestion
```

crates/integration-tests/dmarc.lua
```diff
@@ -0,0 +1,152 @@
...
+
+  local ok, err = pcall(kumo.inject_message, msg)
+  if not ok then
+    kumo.log_error('failed to inject DMARC: ', err)
+  end
+end)
+
+kumo.on('smtp_server_message_received', function(msg)
+  local dkim_auth_results = msg:dkim_verify()
+  if #dkim_auth_results == 0 then
```
wez, line 113: is this a load-bearing part of the test setup? eg: we're injecting a message that isn't signed, and we rely on a non-empty auth results?

Could do with a comment here to explain what's happening and how it relates to the test outcome

crates/integration-tests/dmarc.lua
```diff
@@ -0,0 +1,152 @@
...
+    dkim_auth_results = {
+      {
+        method = 'dkim',
+        result = 'none',
+        reason = 'message was not signed',
+      },
+    }
+  end
+
+  msg:set_meta('received_from', '127.0.0.1:42')
```
wez, line 125: similarly, what's happening here?  Does the test rely on this specific value? It should already be set to `127.0.0.1:XXX` with a randomish port.  If you need a fixed port number it would be cleaner, albeit with a bit more ceremony, to use the XCLIENT SMTP extension to fake the the value as part of the send.

crates/integration-tests/dmarc.lua
```diff
@@ -0,0 +1,152 @@
...
+  local spf_auth_result = spf_disp.result
+
+  kumo.log_info 'DMARC checking message'
+  kumo.dmarc.check_msg(msg, dkim_auth_results, nil, spf_auth_result, true, {
+    org_name = 'testorg',
+    email = 'dmarc-feedback@example.com',
+  })
+
+  local result = msg:import_scheduling_header 'X-Schedule'
+  kumo.log_info('schedule result', kumo.serde.json_encode(result))
```
wez: I think you can remove these lines as it doesn't look like you're passing in an X-Schedule header in the message that you're generating in the test

```suggestion
```

crates/kumo-dmarc/src/lib.rs
```diff
@@ -13,23 +13,392 @@ use crate::types::results::{AuthResults, DmarcResult, PolicyEvaluated, Results,
...
 
 #[cfg(test)]
 mod tests;
 
-const DMARC_REPORT_LOG_FILEPATH: &'static str = "/var/log/kumomta/dmarc.log";
+const DMARC_REPORT_LOG_FILEPATH: &'static str = "/tmp/kumomta-dmarc.log";
+const MAX_ERROR_BATCH_SIZE: usize = 10000;
+
+declare_event! {
+    static DMARC_REPORT_GENERATED: Single("dmarc_report_generated",
```
wez, line 43: Doc comments here are used to populate the user-facing docs; check out a couple of examples at https://docs.kumomta.com/reference/events/get_egress_pool/ and https://docs.kumomta.com/reference/events/get_egress_path_config/

It would be good to briefly explain what the two parameters are (the generated report and the email address to which the report should be sent).

I think we're just passing the raw XML through, and I'm not sure if the current test is actually encoding that message correctly.

In the doc comments here we should demonstrate the right way to do it; I'll go comment on that part of test itself to give it better locality of context!

```suggestion
    /// Triggered when a DMARC report has been generated.  The `report_content` parameter
    /// holds the DMARC report XML content and the `report_email` parameter specifies where the report
    /// should be sent.
    ///
    /// The expectation is that your implementation of this event handler will make a `Message`
    /// from the report and inject it into the queue subsystem for delivery.
    static DMARC_REPORT_GENERATED: Single("dmarc_report_generated",
```

crates/kumo-dmarc/src/lib.rs
```diff
@@ -13,23 +13,392 @@ use crate::types::results::{AuthResults, DmarcResult, PolicyEvaluated, Results,
...
+use tokio::sync::mpsc::{self, Receiver, Sender};
 use uuid::Uuid;
 
 mod types;
 
 #[cfg(test)]
 mod tests;
 
-const DMARC_REPORT_LOG_FILEPATH: &'static str = "/var/log/kumomta/dmarc.log";
+const DMARC_REPORT_LOG_FILEPATH: &'static str = "/tmp/kumomta-dmarc.log";
```
wez: I'd love to eliminate this hard coded path both in general (ideally it's just a required user configuration option and not a static global), and particularly because writing to `/tmp` is a bit of a minefield for a service that starts as root then drops privs.

crates/integration-tests/dmarc.lua
```diff
@@ -0,0 +1,152 @@
...
+kumo.on('get_listener_domain', function(domain, listener, conn_meta)
+  return kumo.make_listener_domain {
+    relay_to = true,
+    log_oob = true,
+    log_arf = true,
+  }
+end)
+
+kumo.on('dmarc_report_generated', function(body, email)
+  local msg = kumo.make_message(
```
wez, line 95: I think `body` is really better labelled as `report_xml`, and if it is just the raw xml, we need to wrap this up in a mime container, something like this:

```lua
local builder = kumo.mimepart.builder()
builder:text_plain "This is an aggregate DMARC report from MYSITE"
builder:attach("text/xml", report_xml, {
  -- I think we need to pass most of this stuff to the event callback in order to create
  -- a well formed message
  file_name = "receiver!policy-domain!begin-timestamp!end-timestamp!unique-id.xml"
})
local part = builder:build()
part.headers:set_subject "DMARC report"
part.headers:set_to(email)
part.headers:set_from('sender@example.com')

local msg = kumo.make_message('sender@example.com', email, tostring(part))
```

Looks like https://datatracker.ietf.org/doc/html/rfc7489#section-7.2 has some rules around how to bundle up the email content, and I think we need to exposure more of the information about the report chunk to the event callback to be able to handle that here.

The alternative is that we do the mime construction in kumo-dmarc ourselves and this event receives a `MimePart` that the consumer can then finish up as they wish.  (or we turn it into a `Message` in kumo-dmarc and pass that).

wez (commented):
just a partial review for the moment: I think we have plenty to chat about in a sec!

crates/kumo-dmarc/src/lib.rs
```diff
@@ -13,23 +13,392 @@ use crate::types::results::{AuthResults, DmarcResult, PolicyEvaluated, Results,
...
+pub fn set_report_window_in_seconds(secs: u64) {
+    REPORT_WINDOW_IN_SECONDS.store(secs, Ordering::Relaxed)
+}
+
+pub fn startup_dmarc_reporter() {
+    tokio::spawn(async move {
+        // Do the initial flush of all pre-existing records that need to be reported from the previous instance
+        match send_aggregated_reports().await {
+            Err(_err) => {
+                // todo: do something with this error
```
wez: I think we should consider making `send_aggregated_reports()` infallible and have it internally log what it needs to provide context to the system operator

crates/kumo-dmarc/src/lib.rs
```diff
@@ -13,23 +13,392 @@ use crate::types::results::{AuthResults, DmarcResult, PolicyEvaluated, Results,
...
+            if let Ok(result) = instant_xml::to_string(&report) {
+                send_generated_report(result, email).await?;
+            }
+        }
+    } else {
+        // todo: we need a way to warn the user the report could not be processed
+    }
+
+    Ok(())
+}
```
wez: I'd lean towards something like this:

```suggestion
pub(crate) async fn send_aggregated_reports() {
    let email_reports = match aggregate_errors().await {
       Ok(reports) => reports,
       Err(err) => {
          tracing::error!("failed to aggregate reports: {err:#}");
          return;
       }
    };

        for (email, report) in email_reports {
            if let Ok(result) = instant_xml::to_string(&report) {
                if let Err(err) = send_generated_report(result, email).await {
                   tracing::error!("dmarc_report_generated event handler failed: {err:#}");
                }
            }
        }
}
```

crates/kumo-dmarc/src/lib.rs
```diff
@@ -13,23 +13,392 @@ use crate::types::results::{AuthResults, DmarcResult, PolicyEvaluated, Results,
...
+
+    let mut remaining_time = match secs_since_epoch {
+        Ok(duration) => {
+            let secs = duration.as_secs();
+
+            if secs % window_in_seconds > 0 {
+                Duration::from_secs(secs % window_in_seconds)
+            } else {
+                Duration::from_secs(window_in_seconds)
+            }
```
wez: I think this just collapses to this in practice?

```suggestion
                Duration::from_secs(secs % window_in_seconds)
```

crates/kumo-dmarc/src/lib.rs
```diff
@@ -13,23 +13,392 @@ use crate::types::results::{AuthResults, DmarcResult, PolicyEvaluated, Results,
...
+        }
+        _ => Duration::from_secs(window_in_seconds),
+    };
+
+    loop {
+        match tokio::time::timeout(remaining_time, receiver.recv()).await {
+            Ok(r) => {
+                match r {
+                    Some(error_record) => {
```
wez: could we collapse up like this to save on nesting level?

```suggestion
            Ok(Some(error_record)) => {
```

crates/kumo-dmarc/src/lib.rs
```diff
@@ -13,23 +13,392 @@ use crate::types::results::{AuthResults, DmarcResult, PolicyEvaluated, Results,
...
+                                if let Ok(result) = serde_json::to_string(&error_record) {
+                                    let _ = writeln!(f, "{result}");
+                                }
+                            }
+                        } else {
+                            tracing::error!("Failed to make spool for DMARC reporting");
+                        }
+                    }
+                    None => {
+                        // break?
```
wez: yep, breaking out here seems righteous

```suggestion
```

crates/kumo-dmarc/src/lib.rs
```diff
@@ -13,23 +13,392 @@ use crate::types::results::{AuthResults, DmarcResult, PolicyEvaluated, Results,
...
+    loop {
+        match tokio::time::timeout(remaining_time, receiver.recv()).await {
+            Ok(r) => {
+                match r {
+                    Some(error_record) => {
+                        let mut incoming_batch = vec![error_record];
+
+                        let mut num_records = 0;
+                        while !receiver.is_empty() && num_records < MAX_ERROR_BATCH_SIZE {
+                            match receiver.recv().await {
```
wez: Consider using something like:

```rust
   while incoming_batch.len() < MAX_ERROR_BATCH_SIZE {
       match receiver.try_recv() {
          Ok(next_record) => { ... }
          Err(TryRecvError::Disconnected) => {
             // dmarc reporting channel closed, must be shutting down?
             // log what we have before we terminate the task
             break; 
          }
          Err(TryRecvError::Empty) => { break }
       }
    }
```

I think we can also eliminate `num_records` here if we just use `incoming_batch.len()` ?

crates/kumo-dmarc/src/lib.rs
```diff
@@ -13,23 +13,392 @@ use crate::types::results::{AuthResults, DmarcResult, PolicyEvaluated, Results,
...
+                            File::options().append(true).create(true).open(current_log)
+                        {
+                            for error_record in incoming_batch {
+                                // TODO how do we want this to report out if we can't serialise for the report?
+                                if let Ok(result) = serde_json::to_string(&error_record) {
+                                    let _ = writeln!(f, "{result}");
+                                }
+                            }
+                        } else {
+                            tracing::error!("Failed to make spool for DMARC reporting");
```
wez: Always a good idea to capture the error message and log it, as it is guaranteed to be the one thing missing in a 2am support session! :)

```suggestion
                            tracing::error!("Failed to write to {current_log}: {err:#}");
```

crates/kumo-dmarc/src/lib.rs
```diff
@@ -13,23 +13,392 @@ use crate::types::results::{AuthResults, DmarcResult, PolicyEvaluated, Results,
...
+
+                remaining_time = match secs_since_epoch {
+                    Ok(duration) => {
+                        let secs = duration.as_secs();
+
+                        if secs % window_in_seconds > 0 {
+                            Duration::from_secs(secs % window_in_seconds)
+                        } else {
+                            Duration::from_secs(window_in_seconds)
+                        }
```
wez: I think this collapses as well?

```suggestion
                            Duration::from_secs(secs % window_in_seconds)

```

worth breaking this match-since-epoch -> Duration pattern out into a little helper function?

crates/kumo-dmarc/src/lib.rs
```diff
@@ -13,23 +13,392 @@ use crate::types::results::{AuthResults, DmarcResult, PolicyEvaluated, Results,
...
+                            ))
+                        })
+                        .into();
+
+                input_records.push(result?);
+            }
+        }
+
+        // What should we do if this fails to remove?
+        let _ = std::fs::remove_file(matching_historical_log);
```
wez: Let's log it at least.  This could be something for which we should add a prometheus metric.  I suspect that if we can't remove the file, the world is effectively ending for this server in any case, so I'm not sure that we could do anything more useful or proactive than logging.

```suggestion
         if let Err(err) = std::fs::remove_file(matching_historical_log) {
            tracing::error!("Failed to remove dmarc log {matching_historical_log}: {err:#}. Duplicate reports may be sent on the next run in XXX secs");
         }
```

crates/kumo-dmarc/src/lib.rs
sophiajt: That would not work for the case where we're at 0 but we want to wait the full amount

PR #537, opened 2026-06-16
http injection: reject newlines in generated headers

### LINKED ISSUE #296: detect and either error out or fixup trying to use newlines in headers in the http injection api
We had a user report issues when trying to do this:

```json
{
  "envelope_sender": "ok@example",
  "content": {
    "subject": "hello\n\nHello there",
    "html_body": "<html>This is the HTML content of the message</html>",
    "text_body": "This is the Text content of the message",
    "from": {
      "email": "ok@example.com",
      "name": "Test"
    },
    "headers": {
      "X-Tenant": "tenant"
    }
  },
  "recipients": [
    {
      "email": "ubuntu@example.com"
    }
  ]
}
```

the newlines in the subject caused the message body to start in the wrong place and caused issues with the tenant header seemingly not being present.

We should either raise an error for this case, or automatically fixup the newlines in the input so that the header value cannot escape and break the message.

Whichever we choose, we should apply it both to the subject header and any other arbitrary headers that are being passed into the builder.

gbretas:
Fixes #296.

A newline in the subject or a custom header passed to the HTTP injection builder must not be able to split the header block and inject an extra header.

It turns out `mailparsing::Header::new_unstructured` already neutralizes this: `wrap_bytes` folds ASCII values (collapsing the whitespace run) and `qp_encode` escapes non-ASCII ones, so the value cannot escape. This PR relies on that existing behavior instead of a separate check in the inject builder:

- Drop the redundant validation in `inject_v1`.
- Lock the guarantee with a regression test at the `new_unstructured` layer, covering ASCII, non-ASCII and CRLF inputs (so Lua/Rust callers are covered too).
- Keep an inject-path regression test for #296.

Note: the fold collapses a multi-line value onto a single logical line. Preserving the author's line breaks as proper CFWS folds would be new behavior and can be a follow-up.

Test:

```
cargo test -p mailparsing test_unstructured_newline_cannot_inject_header
cargo test -p kumod test_generate_builder_subject_newline_cannot_inject_header
```

### DIFF
--- a/crates/kumod/src/http_server/inject_v1.rs
+++ b/crates/kumod/src/http_server/inject_v1.rs
@@ -476,6 +476,14 @@ struct Compiled<'a> {
     attached: Vec<MimePart<'a>>,
 }
 
+fn validate_header_value(header_name: &str, value: &str) -> anyhow::Result<()> {
+    anyhow::ensure!(
+        !value.as_bytes().contains(&b'\r') && !value.as_bytes().contains(&b'\n'),
+        "content.headers['{header_name}'] must not contain newline characters"
+    );
+    Ok(())
+}
+
 impl<'a> Compiled<'a> {
     pub fn expand_for_recip(
         &self,
@@ -557,6 +565,7 @@ impl<'a> Compiled<'a> {
                         need_to = false;
                     }
                     let expanded = self.env_and_templates.borrow_dependent()[id].render(&subst)?;
+                    validate_header_value(name, &expanded)?;
                     id += 1;
                     builder.push(mailparsing::Header::new_unstructured(
                         name.to_string(),
@@ -1621,6 +1630,50 @@ Some(
         );
 
         k9::assert_equal!(structure.attachments.len(), 1);
+    }
+
+    #[tokio::test]
+    async fn test_generate_builder_rejects_newline_in_header_value() {
+        let mut request = InjectV1Request {
+            envelope_sender: "noreply@example.com".to_string(),
+            recipients: vec![Recipient {
+                email: "user@example.com".to_string(),
+                name: Some("James Smythe".to_string()),
+                substitutions: HashMap::new(),
+                metadata: HashMap::new(),
+            }],
+            substitutions: HashMap::new(),
+            content: Content::Builder {
+                text_body: Some("I am the plain text".to_string()),
+                amp_html_body: None,
+                html_body: None,
+                attachments: vec![],
+                subject: Some("hello\n\nHello there".to_string()),
+                from: None,
+                reply_to: None,
+                headers: Default::default(),
+            },
+            deferred_spool: true,
+            deferred_generation: false,
+            trace_headers: Default::default(),
+            template_dialect: Default::default(),
+        };
+
+        request.normalize().unwrap();
+        let compiled = request.compile().unwrap();
+        let err = compiled
+            .expand_for_recip(
+                &request.recipients[0],
+                &request.substitutions,
+                &request.content,
+            )
+            .unwrap_err();
+
+        assert!(
+            err.to_string()
+                .contains("content.headers['Subject'] must not contain newline characters"),
+            "unexpected error: {err:#}"
+        );
     }
 
     #[tokio::test]
--- a/crates/kumod/src/http_server/inject_v1.rs
+++ b/crates/kumod/src/http_server/inject_v1.rs
@@ -476,14 +476,6 @@ struct Compiled<'a> {
     attached: Vec<MimePart<'a>>,
 }
 
-fn validate_header_value(header_name: &str, value: &str) -> anyhow::Result<()> {
-    anyhow::ensure!(
-        !value.as_bytes().contains(&b'\r') && !value.as_bytes().contains(&b'\n'),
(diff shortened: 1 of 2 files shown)


gbretas:
Tested this with Docker using the repo Rust toolchain.

```
cargo test -p kumod test_generate_builder_rejects_newline_in_header_value
```

Result:

```
running 1 test
test http_server::inject_v1::test::test_generate_builder_rejects_newline_in_header_value ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 32 filtered out; finished in 0.07s
```

Also ran:

```
cargo fmt --check --package kumod -- crates/kumod/src/http_server/inject_v1.rs
git diff --check
```

crates/kumod/src/http_server/inject_v1.rs
```diff
@@ -476,6 +476,14 @@ struct Compiled<'a> {
     attached: Vec<MimePart<'a>>,
 }
 
+fn validate_header_value(header_name: &str, value: &str) -> anyhow::Result<()> {
+    anyhow::ensure!(
+        !value.as_bytes().contains(&b'\r') && !value.as_bytes().contains(&b'\n'),
```
wez: From a performance perspective, this is non-ideal as in the common case neither condition will be true but we will have scanned the string twice.  It's also bad because it does a byte-wise traversal of a multi-byte utf-8 string: it may false-positive match parts of a multi-byte sequence.

This could be rephrased as something like `!value.chars().any(|c| c == '\r' || c == '\n')`

crates/kumod/src/http_server/inject_v1.rs
```diff
@@ -476,6 +476,14 @@ struct Compiled<'a> {
     attached: Vec<MimePart<'a>>,
 }
 
+fn validate_header_value(header_name: &str, value: &str) -> anyhow::Result<()> {
```
wez: TBH, I think this is the wrong layer for this.  `Header::new_unstructured` should probably reject this, so that we can catch it when eg: constructing headers from other rust or lua code.

However, since `Header::new_unstructured` has logic to (re)wrap the input, I think it may be more friendly to have it rewrite a newline not followed by a space into a CFWS sequence (newline followed by space).  We don't have a function that does literally that today.

Take a look and let me know what you find/think?

wez (commented):
Thanks for diving in!

crates/kumod/src/http_server/inject_v1.rs
gbretas: Good catch on the double scan. I folded this into the bigger change below: the check moved out of `inject_v1` entirely, so this line is gone now.

For the record, the multi-byte worry doesn't apply here. `value` is a `&str` (so valid UTF-8) and CR/LF are `0x0D`/`0x0A`, so a byte search for them can't match a continuation byte, which are always `>= 0x80`. But it's moot with the new approach.

crates/kumod/src/http_server/inject_v1.rs
gbretas: You were right that this is the wrong layer, and it turns out `new_unstructured` already handles it. I ran the #296 payload through it directly (mailparsing, rust 1.92) and all three arms neutralize the newline:

```
"hello\n\nHello there"   => Subject: hello Hello there       (wrap splits on ascii whitespace)
"héllo\n\nInjected: yes" => =?UTF-8?q?...=0A=0A...?=         (qp_encode escapes it)
"X\r\nInjected: yes"     => Subject: X Injected: yes
```

So the value can't escape the way #296 describes; `wrap_bytes` already collapses the whitespace run. That made the `inject_v1` check redundant, so I dropped it and instead locked the guarantee with a regression test at the `new_unstructured` layer (which also covers the Lua/Rust callers you were worried about), plus an inject-path test for #296.

The only residual is that the collapse is silent: a multi-line subject quietly becomes a single line. If you'd rather preserve the author's line breaks as proper CFWS folds instead of collapsing them to a single space, that's genuinely new behavior, since `wrap` reflows purely by width and discards the original breaks. Happy to do that as a follow-up if you prefer it. Let me know which way you want to go.

gbretas:
@wez thanks for the review! I pushed a new approach based on what I found (details in the inline replies).

tl;dr: `new_unstructured` already neutralizes newlines (folds ASCII, qp-encodes non-ASCII), so the value can't escape and the separate check in the inject builder was redundant. I dropped it and locked the guarantee with a regression test at the `new_unstructured` layer (covers Lua/Rust callers too) plus an inject-path test for #296.

One open question for you: the fold collapses a multi-line value onto a single line. Want to keep that, or should a bare newline be rewritten into a CFWS fold to preserve the author's line breaks? That'd be new behavior, so I'm happy to do it as a follow-up if you'd prefer it.

wez:
TBH, since this is already working, then I don't really want to spend more time on this; it's work for the sake of doing work.  This PR sounds like I'm talking to claude in disguise and it's not worth the time.

gbretas:
Fair. Once I saw new_unstructured already folds it, this turned into a test for something that already works, not worth your time. Closing it. Sorry for the noise, and thanks for pointing me at the right layer.

Issue #539, opened 2026-06-18
Messages stuck indefinitely in a ready queue (never delivered, bounced, or expired) — max_age not enforced for ready-queue residents


alexanderwburch:
## Summary

A message can remain in a **ready queue** indefinitely — never delivered, never bounced, and **never expired** — if that ready queue stops making delivery progress while still holding a backlog (e.g. a dispatcher/connection wedges without producing an error). `max_age` is only enforced on the **scheduled-queue** requeue path, so a message that is resident in a ready queue is never re-evaluated against `max_age`. The only mechanism that flushes a ready queue back to the scheduled queue (reaping) requires `ready_count() == 0`, which a wedged-with-backlog queue never reaches.

We hit this in production: ~11,140 messages across 14 nodes were stuck this way against ~8 different providers, some for **3+ days** despite a 6h `max_age`. They showed up to customers as "accepted but no further events" (not delivered, not bounced, not expired).

## Environment

- Observed on `2026.03.04.074707_bb93ecb1` (amzn2023), but the relevant code is unchanged on `main` (`1632d4a92295` at time of writing).

## Root cause

**`max_age` is only checked when a message is (re)inserted into the scheduled queue.** In `crates/kumod/src/queue/queue.rs` the expiry checks live exclusively in:

- `increment_attempts_and_update_delay` — regular queue-based expiry at ~L854 (`get_max_age()` … "exceeds max_age").
- `requeue_message_internal` — expiry at ~L932.

Both are part of the **scheduled-queue requeue** flow (transient failure → delay → reinsert). They only run when a delivery is actually *attempted* and the message is being requeued.

**The ready queue never checks message age.** `crates/kumod/src/ready_queue.rs` contains **no `max_age` reference at all**. Its only time concept is `AGE_OUT_INTERVAL` (10m, L83), which governs the maintainer wake interval and idle-queue reaping — not message age. The maintainer (`maintainer_task`, ~L530) calls `maintain()` to drive dispatchers, but never evaluates resident messages against `max_age`.

**The one ready→scheduled escape hatch can't fire for a wedged queue.** `reapable()` (L1164) requires:

```rust
&& ((last_change.elapsed() >= AGE_OUT_INTERVAL) ...)
&& self.ready_count() == 0   // L1175
```

[...]

PR #542, opened 2026-06-22
rfc5321: bound the outbound TLS handshake with a timeout (fixes ready-queue wedge #539)

alexanderwburch:
## Summary

Fixes the connection-wedge root cause behind #539.

Every network operation in the SMTP client is bounded by a timeout **except the TLS handshake**. `starttls_timeout` only covers the `STARTTLS` command/response exchange (`parser.rs`: `Command::StartTls => timeouts.starttls_timeout`); the handshake negotiation that follows is awaited bare:

- rustls: `connector.connect(server_name, socket).into_fallible().await`
- openssl: `Pin::new(&mut ssl_stream).connect().await`

So a peer that accepts `STARTTLS` (returns `220`) and then **stalls the handshake** — sending nothing further — parks the dispatcher on that `.await` indefinitely. The connection stays `ESTABLISHED`, produces neither a delivery nor a transient error, and holds its slot in the ready queue. The messages behind it are never attempted, and because `max_age` is only enforced on the scheduled-queue path, they are never aged out either — they sit forever (the "stuck in Processed" symptom in #539). We observed this in production correlated with destinations that throttle/block a sending IP (the edge accepts TCP + STARTTLS but won't complete TLS).

## Changes

- Add a new `tls_handshake_timeout` egress-path option (default `60s`) to `SmtpClientTimeouts`.
- Wrap both the rustls and openssl handshake awaits in `tokio::time::timeout`. On timeout, return a new `ClientError::TimeOutHandshake`, which tears down the connection and requeues the message for a later attempt (potentially via a different egress source) — preventing the wedge and allowing existing `max_age` enforcement to apply.
- Regression test: a fake SMTP server that accepts `STARTTLS` then stalls; asserts `starttls()` fails promptly with `TimeOutHandshake`.
- Docs page for `tls_handshake_timeout` + changelog entry; updated the four `k9` shaping snapshots.

This is the connection-layer fix. The complementary `max_age`-in-ready-queue safety net discussed in #539 can follow as a separate change.

## Test plan

- [ ] `cargo test -p rfc5321 starttls_handshake_timeout`
- [ ] `cargo test -p kumo-api-types` (shaping snapshots)
- [ ] CI build/clippy


wez:
Thanks for this! I independently came up with https://github.com/KumoCorp/kumomta/commit/1b726e4db22d9925114de75819b13f983a9ce7e9 before you submitted this.  I wasn't convinced that it matched all of the symptoms that you described, in particular because some of what you mentioned made it sound like hangs could happen mid-session.  Also, while the commit I pushed doesn't include tests like your mocked up hanging server, I did build out a couple of variations on that during some investigation and while the stall did reproduce, I wasn't confident that the handshake-alone replicated exactly the conditions you described.I also added a timeout on flushing which was absent, and for belt-and-suspenders, added a more general built-in watchdog facility that you can control via https://docs.kumomta.com/reference/kumo/make_egress_path/dispatcher_progress_watchdog_timeout/

There's also a new `kcli inspect-ready-q` command that can be used to introspect the state of a readyq (but not the messages).  Output looks something like:

```
queue: my-source->gmail.com@smtp_client

egress pool:        default
egress source:      my-source
protocol:           smtp_client
watchdog threshold: 10m

mx (site: gmail-smtp-in.l.google.com, dnssec verified):
    10  gmail-smtp-in.l.google.com
    20  alt1.gmail-smtp-in.l.google.com

ready: 1,142

ceilings:
  concurrent dispatchers: 32
    source: connection_limit

connections: 4
  SESSION                              DELIV TRANS FAIL    AGE   RATE         
  9e4b6a01-2345-6789-abcd-ef0123456789 1,391     0    0 47m 8s  0.5/s ⚠ STUCK 
  DeliveringMessage for 42m 12s
  send msg with 1 recip(s)

  7f3a8d4e-9c12-4abc-8def-1234567890ab   142     4    0 8m 12s 18.0/s         
  DeliveringMessage for 1s 200ms
  send msg with 1 recip(s)

  b1c2d3e4-1111-2222-3333-444455556666    45     1    0     8m  0.1/s ⚠ STALL?
  DeliveringMessage for 8m
  send msg with 1 recip(s)

  c2d35e10-aaaa-bbbb-cccc-dddddddddddd     0     0    0    31s  0.0/s         
  Idle for 31s
```

Issue #543, opened 2026-06-24
DANE implementation does not appear to be correct


vdukhovni:
The DANE implementation in this MTA does not appear to have the required downgrade resistance, largely defeating the purpose of implementing DANE in the first place.

* The DNS lookups for DANE TLSA records raise an error ONLY if the TLSA records are **bogus**, and not when the lookup SERVFAILs, times out or encounters various other errors.  The only non-error states should be:
    - Lookup success returning zero (NODATA) or more records, possibly DNSSEC validated or otherwise validated insecure)
    - Lookup success returning NXDOMAIN (possibly DNSSEC validated, or otherwise validated insecure)
* I see no code that verifies whether the MX host's A or AAAA RRset is signed, without which, there's no way to know whether it is safe to query for the TLSA records. For example:

```
; NOERROR qr rd ra ad
nist.gov. IN MX 0 nist-gov.mail.protection.outlook.com.

; SERVFAIL qr rd ra
_25._tcp.nist-gov.mail.protection.outlook.com. IN TLSA ?
```
That `SERVFAIL` from my local resolver is a result of longstanding broken nameservers at Microsoft:
```
; NOERROR qr rd ra
mail.protection.outlook.com. IN NS ns1-proddns.glbdns.o365filtering.com.
mail.protection.outlook.com. IN NS ns2-proddns.glbdns.o365filtering.com.

; <<>> DiG 9.18.48 <<>> +noall +norecur +noedns +cmd +comment +question -t TLSA _25._tcp.nist-gov.mail.protection.outlook.com @ns1-proddns.glbdns.o365filtering.com.
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOTIMP, id: 46768
;; flags: qr; QUERY: 1, ANSWER: 0, AUTHORITY: 0, ADDITIONAL: 0

;; QUESTION SECTION:
;_25._tcp.nist-gov.mail.protection.outlook.com. IN TLSA
```

This is explained in detail in section 2 of RFC7672, but none of the required logic is seen here.

Also, the implementation passes PKIX-TA(0) and PKIX-EE(1) TLSA records to OpenSSL, which it should not be doing for SMTP with DANE.  Only DANE-TA(2) and DANE-EE(3) are in scope for DANE SMTP (see RFC7672). Passing of the private code points is particularly  surprising here.

* If libunbound is used as an in-application validating resolver, how are the root zone trust-anchor keys managed?
[...]

PR #545, opened 2026-06-25
fix: DANE downgrade resistance

wez:
We weren't checking the DNSSEC status for A/AAAA when considering DANE, and were overly permissive in a couple of related cases.

This diff ended up fairly large because the bulk of it is to facilitate testing.

refs: https://github.com/KumoCorp/kumomta/issues/543
refs: https://github.com/KumoCorp/kumomta/pull/517


wez:
@vdukhovni could I trouble you for a review? Thank you!

crates/dns-resolver/src/lib.rs
```diff
@@ -167,47 +169,148 @@ pub fn get_resolver() -> Arc<Box<dyn Resolver>> {
...
+
+/// Maps the result of the TLSA lookup onto a [`DaneStatus`], applying the
+/// RFC 7672 rules. Split out from [`resolve_dane`] so that it can be unit
+/// tested without a live (DNSSEC validating) resolver.
+fn classify_tlsa_answer(mx_host: &str, port: u16, answer: Result<Answer, DnsError>) -> DaneStatus {
+    let answer = match answer {
+        Ok(answer) => answer,
+        Err(err) => {
+            // A SERVFAIL/timeout from a validating resolver leaves the TLSA
+            // status unknown; we must not downgrade.
```
vdukhovni: FWIW, the comment is perhaps misleading, this is case is about there not being an answer at all, so typically a timeout or some sort of resource or communication issue. A `SERVFAIL` is just of the failure `RCODE` values handled below, that aren't either validated signed or validated unsigned (insecure) NOERROR/NXDOMAIN codes that are not errors as such.

crates/dns-resolver/src/lib.rs
```diff
@@ -167,47 +169,148 @@ pub fn get_resolver() -> Arc<Box<dyn Resolver>> {
...
+/// `mx_host` must be the MX hostname (RFC 7672 section 3.2.2: the MX hostname is
+/// the only reference identifier), not the envelope/routing domain.
+///
+/// The caller is responsible for only invoking this when the chain to the MX
+/// host was securely (DNSSEC) resolved: both the MX RRset (see
+/// [`MailExchanger::is_secure`]) and the MX host's address records (see
+/// [`ResolvedAddress::is_secure`](kumo_log_types::ResolvedAddress)) must be
+/// secure. RFC 7672 section 2.2 requires the address records to be secure
+/// before it is safe to rely on TLSA records; we satisfy that by gating on the
+/// secure status of the very address we are about to connect to.
```
vdukhovni, line 235: Note that in less common (roughly 1%) of cases where the MX hostname is a CNAME into an unsigned zone, but the MX hostname's zone is signed, it is still possible for the TLSA records to be signed, so a check for that can be made by validating the security status of a second explicit CNAME query. There are very few domains for which this is applicable, but some do exist (and even mostly have certificates that match the TLSA records):

```
braindeath.be. IN MX 10 mail.braindeath.be. ; NoError AD=1
mail.anoebis.be. IN A 144.21.32.133 ; NoError AD=0
mail.anoebis.be. IN AAAA ? ; NODATA AD=0
mail.braindeath.be. IN CNAME mail.anoebis.be. ; NOERROR AD=1
_25._tcp.mail.braindeath.be. IN TLSA 3 1 1 2f92...9ed3 ; NoError AD=1
```

If you feel this is not worth supporting, I'd understand, but if the CNAME fallback is not too painful, that would be nice to have.

crates/dns-resolver/src/lib.rs
```diff
@@ -167,47 +169,148 @@ pub fn get_resolver() -> Arc<Box<dyn Resolver>> {
...
+
+    if usable.is_empty() {
+        // TLSA records exist but none are usable for DANE SMTP.
+        return DaneStatus::Unusable;
+    }
+
+    // DNS results are unordered. For the sake of tests, sort these records.
+    // Unfortunately, the TLSA type is not Ord so we convert to string and order
+    // by that, which is a bit wasteful but the cardinality of TLSA records is
+    // generally low.
```
vdukhovni: I'd be tempted to implement `Ord` for TLSA records, based on the (usage, selector, mtype, data) values.
FWIW, my Haskell `dnsbase` stub resolver requires and implements `Ord` for all DNS RData types.

crates/dns-resolver/src/lib.rs
vdukhovni: This is quite wrong. The implicit MX record for a domain with no MX records is "secure", if and only if the NODATA response from the resolver is "secure". Many signed domains have no MX records, just secure A/AAAA records, particularly common for `.br` domains for various reasons.

```
hotanuncios.com.br. IN MX ? ; NODATA AD=1
hotanuncios.com.br. IN A 85.31.63.148 ; NoError AD=1
hotanuncios.com.br. IN AAAA 2a02:4780:14:ad45::1 ; NoError AD=1
_25._tcp.hotanuncios.com.br. IN TLSA 3 0 1 e38b...1d0a ; NoError AD=1
```

docs/reference/kumo/make_egress_path/enable_dane.md
```diff
@@ -6,10 +6,35 @@ When set to `true` (the default is `false`), then `TLSA` records will be
 resolved securely to determine the destination site policy for TLS according
 to [DANE](https://datatracker.ietf.org/doc/html/rfc7672).
 
-If TLSA records are available, then the effective value of `enable_tls` will
-be treated as though it were set to `"Required"` and the OpenSSL DANE implementation
-will be used to verify the server certificate against the TLSA records found
-in DNS.
+The table below applies only when `enable_dane = true` *and* the destination is
+resolved via DNS `MX` records (it does not apply when an explicit `mx_list` is
+in use). A *secure chain* means that both the `MX` RRset and the MX host's
```
vdukhovni: If a user configures override explicit "MX" records for a destination domain, then those should also be treated as "secure" because they are not obtained from potentially unvalidated DNS lookups. Perhaps I don't understand what an "explicit mx_list" is, but if it is local configuration, then these SHOULD be considered a priori secure.

docs/reference/kumo/make_egress_path/enable_dane.md
```diff
@@ -6,10 +6,35 @@ When set to `true` (the default is `false`), then `TLSA` records will be
...
+| Not secure | *(not queried)* | your configured value | DANE does not apply; MTA-STS may still apply |
+| Secure | usable DANE-TA(2)/DANE-EE(3) record(s) | `Required`, with the server certificate checked against the `TLSA` records | deliver only if the certificate matches a record; otherwise defer and try the next host |
+| Secure | record(s) present but none usable | `RequiredInsecure` | STARTTLS required, but the server certificate is **not** checked |
+| Secure | securely absent (NODATA / NXDOMAIN) | your configured value | DANE does not apply; MTA-STS may still apply |
+| Secure | lookup failed (`SERVFAIL`, timeout, or bogus) | — | delivery is deferred (downgrade resistance); nothing is sent in the clear |
+
+Notes:
+
+* When DANE applies (the `Required` / `RequiredInsecure` rows), MTA-STS is
+  **not** consulted; the DANE result wins.
```
vdukhovni: When DANE TLSA records are present, but none are usable, and so only STARTTLS is required, without any means to do authentication, it may be reasonable to then take MTA-STS into account. The domain effectively has no DANE policy, the inferred "at least encrypt" is not strong enough to rule out best-effort MTA-STS as a fallback.

vdukhovni (changes requested):

crates/dns-resolver/src/lib.rs
wez: I agree, but that type comes from the upstream hickory-proto crate so we cannot tackle it in the scope of this PR

docs/reference/kumo/make_egress_path/enable_dane.md
wez: Since we're scriptable and the user may have done some other dns lookups to populate mx_list, I opted to add an explicit `treat_mx_list_as_secure` flag around this, so mx_list is not by default treated as secure, but you can opt into it explicitly if desired/required.

crates/dns-resolver/src/lib.rs
wez: This should be addressed in the latest revision of this PR

docs/reference/kumo/make_egress_path/enable_dane.md
wez: this should be resolved now too

crates/dns-resolver/src/lib.rs
wez: should be better now

crates/dns-resolver/src/lib.rs
wez: I'm still considering this bit; it might end up in a separate PR

wez:
One of the commits in this stack makes the trust-anchor situation more explicit.  Previously, the embedded unbound resolver implicitly enabled its built in trust anchors and didn't discuss or directly reference this concern.  The state here in this commit now also enables DNSSEC in hickory (it was previously blocked by limitations in hickory's implementation) and discusses how to explicitly configure external trust anchors.  At the present time, hickory doesn't support RFC 5011, but unbound does.

crates/dns-resolver/src/lib.rs
wez: I've updated this to add that second CNAME query.
I'd appreciate a second look over the PR when you have a moment, as I believe that I've resolved the issues you raised.
Thank you!

crates/dns-resolver/src/lib.rs
```diff
@@ -320,6 +320,92 @@ fn classify_tlsa_answer(mx_host: &str, port: u16, answer: Result<Answer, DnsErro
...
+///
+/// This is a narrow fallback in the DANE path: when the MX host's address
+/// (A/AAAA) records did not resolve securely we may still be looking at a secure
+/// CNAME whose target merely lives in an unsigned zone. Querying the `CNAME`
+/// type explicitly isolates the alias's own DNSSEC status (a CNAME-type query is
+/// answered by the alias RRset and is not chased into the insecure target), and
+/// works uniformly across resolver backends because it relies only on the
+/// per-answer secure bit rather than per-record validation proofs.
+pub async fn resolve_secure_cname(mx_host: &str) -> anyhow::Result<SecureCnameStatus> {
+    let name = fully_qualify(mx_host)?;
```
vdukhovni, line 356: What does "fully_qualify" do? I did not expect to see that here. The result of an MX lookup is by definition already "fully-qualified".

vdukhovni (commented):
Sadly, I don't quite have enough cycles for a detailed review.
The right sort of code seems to be there.  The CNAME lookup is only needed when the insecure A/AAAA records arrive packaged up as an indirect response:
```
; NOERROR qr rd ra
mx1.example.net. IN CNAME mx1.provider.net.
mx1.provider.net. IN A 192.0.2.1
```
It is in this scenario that a fallback query `mx1.example.net. IN CNAME ?` might reveal the "example.net" zone to be signed, even though the A records at the provider are not.

You seem to have a reasonable grasp on the requirements now. Good luck.

crates/dns-resolver/src/lib.rs
wez: we append a `.` to the name to force the resolver not to add local search domains

wez:
Thanks for your help!

PR #549, opened 2026-06-29
improve cgroup memory usage calculation to avoid counting purgeable cache

dschaaff:
When running under a cgroup limit, get_cgroup() reported stat.usage_in_bytes (memory.current on v2, memory.usage_in_bytes on v1) as the process memory usage. That counter includes all page cache, including cold reclaimable cache (inactive_file). For a workload that writes logs and spool to disk, cold cache can dominate: on a real pod memory.current was ~51GB while inactive_file was ~46GB and the true non-cache footprint was ~1.8GB. kumod saw usage near its limit, hit get_headroom() == 0, and ran
shrink_ready_queue_due_to_low_mem
and other reductions against pressure that was almost entirely reclaimable cache the kernel would drop before any OOM.

Report the working set instead:
```
working_set = max(memory.current - inactive_file, anon)
```
This matches container_memory_working_set_bytes (kubelet/cAdvisor). It still counts everything that can drive an OOM: anonymous memory, dirty page cache, active_file, tmpfs/shm, and slab.

Memory accounting outside of a cgroup is untouched. 

Disclosure: I used Claude to aid in the development of this.


docs/reference/memory.md
```diff
@@ -29,33 +29,63 @@ as `memory_limit`, and is reported in bytes.
...
+- Otherwise, the Resident Set Size (RSS) as reported in `/proc/self/statm`.
+- The value configured via
   [kumo.set_memory_soft_limit](kumo/set_memory_soft_limit.md), if any,
   will always take precedence over the above.
 
 The current usage is published via the prometheus metrics endpoint
 as `memory_usage` and is reported in bytes.
 
+### Working Set under a cgroup
+
```
wez, line 44: It's worth calling out that the behavior is different in different versions.  For net-new stuff we typically stick a `{{since('dev')}}` in the section and call it done.  When we're describe new/old behavior with a bit more nuance, there are a few ways to go; one way is to identify it in prose (you can add `inline=True` to that since macro to make parenthetical), or you could make a little table that contrasts how the math works to briefly call it out and then summarize better below.

docs/reference/memory.md
```diff
@@ -29,33 +29,63 @@ as `memory_limit`, and is reported in bytes.
...
+directly or from tooling such as cAdvisor/kubelet, which expose
+`container_memory_working_set_bytes` and the underlying cache figures.
+
+If the cgroup `memory.stat` cannot be read, KumoMTA falls back to reporting the
+raw `memory.current` as `memory_usage` and logs a single warning.
+
 ### Headroom and Load Shedding
 
-The *memory headroom* of the system is defined as the current value of
+The _memory headroom_ of the system is defined as the current value of
```
wez: There's a lot of "noise" in this file where the syntax has changed the bytes in the file, even though I think these changes are rendering-neutral.  Please revert those changes from this PR to keep it focused on the primary thesis!

wez (changes requested):
Thanks for this! It looks reasonable and understandable.  My comments are all in the docs!

docs/reference/memory.md
dschaaff: @wez Thanks for the feedback! I corrected the auto formatting that clobbered the diff and I believe I have the macro set correctly for describing the behavior change between versions in the docs. Let me know if you have any other requested changes.

wez:
Thank you!

Issue #484, opened 2026-02-28
Aliasing issue with MTA-STS and DANE


wez:
KumoMTA's *site_name* concept enables a degree of automatic configuration "rollup" and sharing of shaping/egress configuration via an egress path.  Multiple routing domains can map to (alias to) the same ready queue because they produce the same *site_name*.  In this situation, the ready queue object is nominally associated with whichever of those routing domains were resolved first to instantiate the ready queue.

We currently have two options that are scoped to an egress path configuration in the code, but whose value should still be interpreted based on the true routing domain:

* enable_mta_sts
* enable_dane

These options provide a way for the routing domain to publish additional policy information that can influence the effective configuration of an egress path.

In most cases this is fine, but it is possible for a misconfigured domain policy to impact other domains that have the same site name.

For example, if a "random" domain is configured with the same set of MX records as a major MBP, and that random domain publishes an MTA-STS policy that restricts the set of valid MX hosts to something that cannot ever match, not only will we not allow sending to that domain, but also to any other domain that matches that site name _**if that random domain was the one that caused that ready queue to instantiate**_.

Example of the misconfigured domain:

`random-domain-example.com` shares the same set of MX records as `icloud.com`:

```console
$ dig +short mx random-domain-example.com
10 mx02.mail.icloud.com.
10 mx01.mail.icloud.com.
```

but publishes an MTA-STS policy that says that they must match something at cloudflare:

```console
$ curl https://mta-sts.random-domain-example.com/.well-known/mta-sts.txt
version: STSv1
mode: enforce
mx: *.mx.cloudflare.net
max_age: 86400
```

If we were to instantiate `random-domain-example.com` before a "native" `icloud.com` address, then we'll end up with a ready queue that cannot deliver because of the conflicting MTA-STS policy, and that ready queue will apply to all other domains that share that same set of MX records.  In this situation you'll see logs like:

```
400 KumoMTA internal: failed to connect to any candidate hosts: 
MTA-STS policy for random-domain-example.com. is set to enforce
but the current MX candidate mx01.mail.icloud.com. does not match 
the list of allowed hosts. MtaStsPolicy { mode: Enforce, 
mx: ["*.mx.cloudflare.net"], max_age: 86400, fields: {} }
```

These are transient errors; no mail will be immediately bounced if your team recognize and remediate it (see below) before the delay becomes too problematic.

[...]

PR #557, opened 2026-07-06
Evaluate MTA-STS during MX resolution to fix aliasing issue

wez:
refs: https://github.com/KumoCorp/kumomta/issues/484

Briefly, the issue is that if some random domain that shares MX records
    with another (eg: someone is using google apps or icloud for their
    vanity domain) publishes a broken MTA-STS policy that requires eg:
    cloudflare MX hosts then because we roll up by site name, that broken
    MTA-STS policy bleeds into all the other domains that share those MX
    records.

The resolution is simple, but is technically a breaking change.

Moving the policy resolution to happen during site_name resolution
    allows us to resolve both per-domain things at the same and have the
    MTA-STS policy amend the effective set of MX hosts.  The output of that
    is then used for site_name aggregation/rollup.

The consequence of this is quite nice: an MTA-STS policy that is more
    restrictive than the full set of MX hosts now prevents delivering to
    any of the excluded hosts, and a totally broken policy that prevents all
    of its MX hosts is now completely undeliverable and will produce
    transient failures.

The downside is that for users that had previously disabled mta-sts in
    their default shaping block, they will need to change a different config
    option to continue to prevent MTA-STS from being consulted.  One example
    of this that I recall is that one user's network posture prevented
    MTA-STS from making HTTPS requests to fetch the policy.  Another user
    just wanted to cut out the additional DNS traffic.  Those use cases
    require altering the new kumo.dns.set_mta_sts_enabled enabled to false
    during `init`.


crates/dns-resolver/src/lib.rs
```diff
@@ -1945,38 +1139,12 @@ MailExchanger {
...
-}
-"#
-        );
-    }
-
-    #[cfg(feature = "live-dns-tests")]
     #[tokio::test]
     async fn txt_lookup_gmail() {
+        let resolver =
+            TestResolver::default().with_txt("_mta-sts.gmail.com", "v=STSv1; id=20190429T010101;");
```
wez, line 1145: testing a comment on the RHS

crates/integration-tests/mta-sts.lua
```diff
@@ -0,0 +1,86 @@
+-- Source policy for the mta_sts_enforce_impossible integration test.
+local kumo = require 'kumo'
+
+local TEST_DIR = os.getenv 'KUMOD_TEST_DIR'
+local SINK_PORT = tonumber(os.getenv 'KUMOD_SMTP_SINK_PORT')
```
wez, line 5: another RHS comment

crates/dns-resolver/src/lib.rs
```diff
@@ -36,10 +29,6 @@ pub use resolver::{
 static RESOLVER: LazyLock<ArcSwap<Box<dyn Resolver>>> =
     LazyLock::new(|| ArcSwap::from_pointee(Box::new(default_resolver())));
 
-declare_cache! {
-/// Caches domain name to computed set of MailExchanger records
-static MX_CACHE: LruCacheWithTtl<(Name, Option<u16>), Result<Arc<MailExchanger>, String>>::new("dns_resolver_mx", 64 * 1024);
-}
 declare_cache! {
 /// Caches domain name to ipv4 records and their DNSSEC secure status
```
wez, line 33: spanning comment

crates/dns-resolver/src/lib.rs
```diff
@@ -54,104 +43,13 @@ declare_cache! {
 static IP_CACHE: LruCacheWithTtl<(Name, IpLookupStrategy), Arc<IpAddresses>>::new("dns_resolver_ip", 1024);
 }
 
-/// Maximum number of concurrent mx resolves permitted
-static MX_MAX_CONCURRENCY: AtomicUsize = AtomicUsize::new(128);
-static MX_CONCURRENCY_SEMA: LazyLock<Semaphore> =
-    LazyLock::new(|| Semaphore::new(MX_MAX_CONCURRENCY.load(Ordering::SeqCst)));
-
-/// 5 seconds in ms
-static MX_TIMEOUT_MS: AtomicUsize = AtomicUsize::new(5000);
```
wez, line 63: lhs only comment

crates/kcli/src/queue_summary.rs
```diff
@@ -1,8 +1,8 @@
 use clap::Parser;
-use dns_resolver::MailExchanger;
 use kumo_api_client::KumoApiClient;
 use kumo_api_types::ReadyQueueStateRequest;
 use lexicmp::natural_lexical_cmp;
+use mailexchanger::MailExchanger;
```
wez, line 5: rhs comment

crates/kumod/src/smtp_dispatcher.rs
```diff
@@ -779,56 +778,33 @@ impl SmtpDispatcher {
...
-                    Err(err) => {
-                        self.tracer.diagnostic(Level::INFO, || {
-                            format!("MTA-STS resolve error for {}: {err:#}", mx.domain_name)
-                        });
+            match dispatcher.mx.as_ref().map(|mx| mx.mta_sts) {
+                Some(PolicyMode::Enforce) => {
+                    enable_tls = Tls::Required;
+                    self.tracer.diagnostic(Level::INFO, || {
+                        "MTA-STS enforce policy in effect; requiring TLS".to_string()
+                    });
```
wez, line 794: push this comment for RHS.

*edited*

crates/dns-resolver/src/lib.rs
wez: replying to lhs-only comment #4

docs/reference/kumo/make_egress_path/enable_mta_sts.md
```diff
@@ -2,14 +2,24 @@
...
-You can set it to `false` to disable querying of the MTA-STS policy.
-
-When set to `true`, we'll issue additional DNS and HTTP requests appropriate
-for the destination domain.  These will be cached so that we won't perform them
-on every request.  For example, for `gmail.com` we'll issue a TXT lookup for
+You can set it to `false` to prevent a policy from raising the TLS posture for
+this egress path.
+
+{{since('dev', indent=True)}}
+    This option influences only whether the TLS portion of the MTA-STS policy
```
wez, line 12: a comment over here in the md file

**boop**

wez:
making a review-level comment
(edited on GH)

(edited again locally)

wez:
manually pushed to `main`

PR #558, opened 2026-07-06
add check_trailing_bits: false

kayozaki:
Adding check_trailing_bits: false into base64 decoder.

While testing message parsing code, there's rare chance sender isn't fully compliant with base64.
This leads to failure such as below

```base64 decode: non-zero trailing bits at 20211 b='i' in HRtbD4NCi==```

As long as it's not destructive, its more convenient to be able to support non-RFC messages so we can extract message artifacts for it's decision making.


wez:
Thank you!

PR #559, opened 2026-07-06
rfc5321: tolerate stray space after MAIL FROM:/RCPT TO: colon

alexanderwburch:
## Summary

Some legacy SMTP clients emit a stray space between the colon and the address
in `MAIL FROM:` / `RCPT TO:` commands (e.g. `MAIL FROM: <addr>`). RFC 5321 does
not allow this space, so `kumod` rejects these commands with `501 5.1.7 Bad
sender's mailbox address syntax`.

Per review feedback, this relaxes the `rfc5321` grammar directly rather than
adding an opt-in listener option: the parser now accepts and discards an
optional run of spaces/tabs after the colon in both `MAIL FROM:` and
`RCPT TO:`. This is applied to both the success arm and the
"valid-address + trailing junk" arm so the resulting `Syntax` /
`InvalidSenderAddress` classification stays correct.

The previous approach (`allow_space_before_path` listener option + helper +
reference doc) has been dropped; `smtp_server.rs` is unchanged from `main`.

## Changes

- `crates/rfc5321/src/parser.rs`: allow optional `wsp` after the `FROM:`/`TO:`
  colon in `parse_mail_from` and `parse_rcpt_to`.
- Added `test_mail_from_space_before_path` and `test_rcpt_to_space_before_path`
  covering single/tab/multi-space, null sender, `<Postmaster>`, and ESMTP
  parameter cases.
- Changelog entry under "Other Changes and Enhancements".

## Test plan

- [x] `cargo test -p rfc5321 --lib parser::` — 165 passed, 0 failed (includes
  the two new tests, no regressions).

### DIFF
--- a/crates/rfc5321/src/parser.rs
+++ b/crates/rfc5321/src/parser.rs
@@ -1820,10 +1820,14 @@ fn parse_mail_from(input: Span) -> IResult<Span, MaybePartialCommand> {
                     tag_no_case("MAIL"),
                     wsp,
                     tag_no_case("FROM:"),
+                    // RFC 5321 has no space between the colon and the reverse-path,
+                    // but many legacy clients emit one (`MAIL FROM: <addr>`). Tolerate
+                    // and discard it rather than rejecting with 501 5.1.7.
+                    opt(wsp),
                     reverse_path,
                     opt(map((wsp, mail_parameters), |(_, p)| p)),
                 )),
-                |(_, _, _, address, parameters)| {
+                |(_, _, _, _, address, parameters)| {
                     MaybePartialCommand::Full(Command::MailFrom {
                         address,
                         parameters: parameters.unwrap_or_default(),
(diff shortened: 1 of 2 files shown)


wez:
Thanks for this!  TBH, I think it would be much simpler overall to just allow an optional space in the grammar; it seems harmless enough and we can avoid the complexity of adding another option and the docs and so on.  WDYT?

alexanderwburch:
@wez Revised per feedback: dropped the opt-in `allow_space_before_path` listener option and instead relaxed the rfc5321 grammar directly to accept and discard an optional space/tab after the `FROM:`/`TO:` colon. `smtp_server.rs` is now unchanged from main. Added parser tests for both verbs (single/tab/multi-space, null sender, `<Postmaster>`, and ESMTP params). PTAL, thanks!

